PR #10, opened 2019-06-14
Merging in stm32h7x3-hal

hargoniX:
This PR merges in the stm32h7x3-hal project, bringing:
* An I2C implementation
* A Watchdog implementation
* Examples for both of them

Furthermore it adds myself as an author and replaces all the `...` inclusive range syntax with `..=` syntax instead as the first one has been deprecated.

Me and a friend of mine are currently also working on an ADC implementation, still for our version of the HAL so maybe it would be better to wait until we are done with it, ported it over and continue our work over at this version of the HAL.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -6,13 +6,12 @@ use crate::gpio::gpioc::PC9;
 use crate::gpio::gpiod::{PD12, PD13};
 use crate::gpio::gpiof::{PF0, PF1, PF14, PF15};
 use crate::gpio::gpioh::{PH11, PH12, PH4, PH5, PH7, PH8};
-use crate::gpio::{Alternate, AF4};
+use crate::gpio::{Alternate, AF4, AF6};
+use crate::hal::blocking::i2c::{Read, Write, WriteRead};
 use crate::rcc::Ccdr;
-use crate::time::Hertz;
-use crate::hal::blocking::i2c::{Write, WriteRead, Read};
 use crate::stm32::{I2C1, I2C2, I2C3, I2C4};
-use cast::{u8, u16};
-
+use crate::time::Hertz;
+use cast::{u16, u8};
 
 /// I2C error
 #[derive(Debug)]
@@ -81,9 +80,9 @@ macro_rules! i2c {
 
                     assert!(freq <= 1_000_000);
 
-                    let i2cclk = ccdr.clocks.$pclkX().0;    
+                    let i2cclk = ccdr.clocks.$pclkX().0;
 
-                    // Refer to figure 539 for this:
+                    // Refer to RM0433 Rev 6 - Figure 539 for this:
                     // Clear PE bit in I2C_CR1
                     unsafe { &(*I2C1::ptr()).cr1.modify(|_, w| w.pe().clear_bit())};
 
@@ -146,7 +145,7 @@ macro_rules! i2c {
                     let scll = u8(scll).unwrap();
 
                     // Configure for "fast mode" (400 KHz)
-                    i2c.timingr.write(|w| 
+                    i2c.timingr.write(|w|
                         w.presc()
                             .bits(presc)
                             .scll()
@@ -165,7 +164,7 @@ macro_rules! i2c {
                     I2c { i2c, pins}
 
                 }
-                
+
                 /// Releases the I2C peripheral and associated pins
                 pub fn free(self) -> ($I2CX, (SCL, SDA)) {
                     (self.i2c, self.pins)
@@ -245,7 +244,7 @@ macro_rules! i2c {
                         // put byte on the wire
                         busy_wait!(self.i2c, txis);
                         self.i2c.txdr.write(|w| w.txdata().bits(*byte));
-                        
+
                     }
 
                     // Wait until the last transmission is finished
@@ -337,7 +336,7 @@ macro_rules! pins {
 }
 
 pins! {
-    I2C1: 
+    I2C1:
         SCL: [
             PB6<Alternate<AF4>>,
             PB8<Alternate<AF4>>
@@ -347,7 +346,7 @@ pins! {
             PB7<Alternate<AF4>>,
             PB9<Alternate<AF4>>
         ]
-    
+
     I2C2:
         SCL: [
             PB10<Alternate<AF4>>,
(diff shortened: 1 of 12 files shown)


richardeoin:
Have a go at using [`rustfmt`](https://github.com/rust-lang/rustfmt) on this. It'll automatically tidy up [trailing whitespace](https://blog.codinghorror.com/whitespace-the-silent-killer/) and other potential gremlins.

src/i2c.rs
```diff
@@ -0,0 +1,398 @@
...
+        SCL: [
+            PD12<Alternate<AF4>>,
+            PF14<Alternate<AF4>>,
+            PH11<Alternate<AF4>>,
+            PB6<Alternate<AF4>>,
+            PB8<Alternate<AF4>>
+        ]
+
+        SDA: [
+            PB7<Alternate<AF4>>,
```
richardeoin: I2C4 on PB6/7/8/9 is on `AF6`

src/i2c.rs
```diff
@@ -0,0 +1,398 @@
...
+            PB9<Alternate<AF4>>,
+            PD13<Alternate<AF4>>,
+            PF15<Alternate<AF4>>,
+            PH12<Alternate<AF4>>
+        ]
+}
+
+i2c!(
```
richardeoin: I think add the default feature gate here, and on the `pins!` definition

```
#[cfg(any(
    feature = "stm32h742",
    feature = "stm32h743",
    feature = "stm32h753",
    feature = "stm32h750"
))]

File: src/i2c.rs [hunk 3]
```diff
@@ -0,0 +1,398 @@
...
+                    ccdr.rb.$apbXrstr.modify(|_, w| w.$i2cXrst().clear_bit());
+                    ccdr.rb.$apbXrstr.modify(|_, w| w.$i2cXrst().set_bit());
+
+                    let freq = freq.into().0;
+
+                    assert!(freq <= 1_000_000);
+
+                    let i2cclk = ccdr.clocks.$pclkX().0;    
+
+                    // Refer to figure 539 for this:
```
richardeoin: Add document reference

src/watchdog.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+impl WatchdogEnable for SystemWindowWatchdog {
+    type Time = u32;
+    /// Starts the watchdog with a given timeout period, if this period is out of bounds the function
+    /// is going to panic
+    fn start<T>(&mut self, period: T)
+    where
+        T: Into<Self::Time>,
+    {
```
richardeoin: The units of `period` should be clear and documented.

A clear option is to use a separate type for milliseconds, defined in `time.rs`
```
/// Time unit
#[derive(PartialEq, PartialOrd, Clone, Copy)]
pub struct MilliSeconds(pub u32);
```

src/watchdog.rs
```diff
@@ -0,0 +1,82 @@
...
+        let period = period.into();
+        let maximum =
+            (4096 * 2u32.pow(7) * 64) / (self.pclk3_frequency.0 / 1000);
+        assert!(period <= maximum);
+
+        // cant approximate this at compile time as the apb clock frequency is not known at compile time
+        // TODO: find a better way for this
+        let mut best_config: (u32, u32) = (0, 0);
+        let mut closest: u32 = 33334;
+        for wdgtb in 0..8 {
```
richardeoin: This should be improved. Assume a constraint on T to the range [32,64), then use a match statement to find the WDGTB divisor value.

src/i2c.rs
hargoniX: Woops, assumed too much consistency on STMs side there, I'll double check this in the future then, thanks for the catch,

src/i2c.rs
hargoniX: How exactly do I add a document reference? (Never done that before)

src/watchdog.rs
hargoniX: Sorry but I'm not actually getting what you want me to do here, if I constraint T to the range of 32,64 I still gotta approximate two values for the WDGTB dvisior, one for 32 and one for 64 or not?

hargoniX:
So @Jan561 and me spend some time on trying to get a bug in the ADC implementation fixed and we believe that this is related to the fact that the ADC is per default clocked from pll2 which we don't enable. If we wanted to change that clock source we would have to have access to the ADCSEL bit which is in the version of stm32h7 we are using (and in the latest version of docs.rs) not available (or at least not findable through a search: https://docs.rs/stm32h7/0.7.1/stm32h7/?search=adcsel) So we are not exactly sure how to go from there, any ideas?

src/i2c.rs
richardeoin: Ah, I just mean reference the ST documentation number and revision in the comment.
Something like `// RM0433 Rev 6 - Figure 539`

src/watchdog.rs
richardeoin: To use the rust notation, constrain T[5:0] to the range `32..64` T[5:0] can't be 64. 

Something like this should do the trick. It's a common pattern in `rcc.rs`, and uses the round towards zero behaviour of integer division in rust (LLVM's `sdiv`). 
```
let ratio = period_ms * (self.pclk3_frequency.0 / 1000) / 4096;

// Prescaler
let (tb_div, tb_bits) = match ratio / 64 {
    0 => (1, 0),
    1 => (2, 1),
    2..=3 => (4, 2),
    4..=7 => (8, 3),
    // More branches here
    _ => (128, 7),
};

// Counter load value
let t = ratio / tb_div;
assert!(t < 64);

self.down_counter = u8(t).unwrap() | (1 << 6);
```

For extra clarity, use the field variants from `stm32-rs`! eg. [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/rcc.rs#L516).

richardeoin:
> If we wanted to change that clock source we would have to have access to the ADCSEL bit which is in the version of stm32h7 we are using (and in the latest version of docs.rs) not available (or at least not findable through a search: https://docs.rs/stm32h7/0.7.1/stm32h7/?search=adcsel) So we are not exactly sure how to go from there, any ideas?

`stm32h7` uses field names from the SVD issued by ST, until someone goes through and corrects them. I [already did this for the RCC](https://github.com/stm32-rs/stm32-rs/pull/206), so the next version of `stm32h7` will have the correct name `ADCSEL`.

src/watchdog.rs
```diff
@@ -0,0 +1,82 @@
...
+                }
+            }
+        }
+
+        let wdgtb = u8(best_config.0).unwrap();
+        self.down_counter = u8(best_config.1).unwrap() | (1 << 6);
+
+        // write the config values, matching the set timeout the most
+        self.wwdg.cfr.modify(|_, w| w.wdgtb().bits(wdgtb));
+        self.wwdg.cfr.modify(|_, w| w.w().bits(self.down_counter));
```
richardeoin: The `wdgtb` field definition currently used in `stm32h7` is [only 2 bits](https://github.com/stm32-rs/stm32-rs/blob/master/peripherals/wwdg/wwdg.yaml#L16) . This needs to be fixed there first - call it `wwdg_v2.yaml` perhaps.

hargoniX:
So @Jan561 figured out that the 8-bit resolution is not implementable yet as the method for setting raw bits to the res field is: https://docs.rs/stm32h7/0.7.1/src/stm32h7/stm32h7x3/adc3/cfgr/mod.rs.html#662-6682 which will turn the flag required for 8 bit resolution (0b100) into 0b000, this is however fixed in stm32-rs/stm32-rs#231. So adding the ADC with a working implementation of 8 bit resolution would depend on this PR being merged ->  I'm gonna open a seperate WIP ADC PR once this one is done

richardeoin:
Looks good with the review changes. This can probably go in now - any thoughts @astraw ?

Great, please do open a separate PR for the ADC.

astraw:
I quickly looked through the code but didn't do a thorough review, and unfortunately I won't have time to do so for some weeks, so I merged it now. Thanks a lot for your contribution!

Issue #9, opened 2019-06-13
Tracking Issue for Revision V silicon


richardeoin:
ST has released a new silicon revision for the `H742/742/750/753` with significant changes.

* [eevblog](https://www.eevblog.com/forum/microcontrollers/stm32h7-series-revision-beware-of-the-changes!/)
* [ST community](https://community.st.com/s/question/0D50X0000Aixasy/stm32h7-cpu-speed-increase)

Please report or tag issues with Revision V here! Also see [upstream](https://github.com/stm32-rs/stm32-rs).

hargoniX:
In the eevblog link above I see two notes regarding the ADC (which as mentioned in #10 would've been my next addition)
```
-ADC Now has a extra 'divide by 2' on its clock inside it, making existing clock configurations too slow
-Changes to the ADC registers (For example a value in a register that used to set the ADC to 8bit mode is now setting it to 'reserved' so likely doesn't work)
```
Now I assume this could be easily resolved using feature gates, however a feature gate named "V" or "Revision-V" might be hard for the average user to understand as they are probably not aware of Revision V.

richardeoin:
Yes, these are significant changes! For now I say implement for Rev Y, then add `-rev_v` feature gates when someone has V silicon to test on.

jordens:
FYI: The NUCLEO-H743ZI**2** I just got from Digikey has Silicon rev V (and non-detachable ST-Link-v3)

jordens:
https://github.com/stm32-rs/stm32-rs/pull/247 for the PAC split.

jordens:
@richardeoin Can we close this? The crates/features are nicely split to address this, right?

PR #12, opened 2019-06-21
Adding an ADC implementation

hargoniX:
Adding the ADC implementation I finally got working yesterday together with an example for it. Furthermore I added @Jan561 as an author as he did most of the work on the ADC except solving a few issues with them. (I'm doing this PR instead of him as he's currently out on vacation)

### DIFF
--- a/examples/adc.rs
+++ b/examples/adc.rs
@@ -19,7 +19,6 @@ use cortex_m_log::println;
 use cortex_m_log::{
     destination::Itm, printer::itm::InterruptSync as InterruptSyncItm,
 };
-use cortex_m_semihosting::hprintln;
 
 #[entry]
 fn main() -> ! {
@@ -59,6 +58,5 @@ fn main() -> ! {
         let data: u32 = adc1.read(&mut channel).unwrap();
         // voltage = reading * (vref/resolution)
         println!(log, "ADC reading: {}, voltage for nucleo: {}", data, data as f32 * (3.3/65535.0));
-        hprintln!( "ADC reading: {}, voltage for nucleo: {}", data, data as f32 * (3.3/65535.0)).unwrap();
     }
 }
--- a/examples/adc.rs
+++ b/examples/adc.rs
@@ -19,6 +19,8 @@ use cortex_m_log::println;
 use cortex_m_log::{
     destination::Itm, printer::itm::InterruptSync as InterruptSyncItm,
 };
+use cortex_m_semihosting::hprintln;
+
 
 #[entry]
 fn main() -> ! {
@@ -34,8 +36,10 @@ fn main() -> ! {
     // Constrain and Freeze clock
     println!(log, "Setup RCC...                  ");
     let rcc = dp.RCC.constrain();
-    // setting this per_ck to 4 mhz here as the last bit doesnt get sampled
-    // if we wouldn't do that
+
+    // setting this per_ck to 4 Mhz here (which is gonna choose the CSI that runs at exactly 4 Mhz) as the adc requires per_ck as its
+    // own kernel clock and wouldn't work at all if per_ck wouldnt be enabled or loose a few bits if it was too fast 
+    // (the maximum for this is 36 Mhz)
     let mut ccdr = rcc.sys_ck(100.mhz()).per_ck(4.mhz()).freeze(vos, &dp.SYSCFG);
 
     println!(log, "");
@@ -45,18 +49,18 @@ fn main() -> ! {
     let mut delay = Delay::new(cp.SYST, ccdr.clocks);
 
     // Setup ADC
-    let mut adc1 = adc::Adc::adc3(dp.ADC3, &mut delay, &mut ccdr);
-    adc1.set_resolution(adc::Resolution::SIXTEENBIT);
+    let mut adc3 = adc::Adc::adc3(dp.ADC3, &mut delay, &mut ccdr);
+    adc3.set_resolution(adc::Resolution::SIXTEENBIT);
 
-    // Setup GPIOB
+    // Setup GPIOC
     let gpioc = dp.GPIOC.split(&mut ccdr.ahb4);
 
-    // Configure pb0, pb1 as an analog input
+    // Configure pc0 as an analog input
     let mut channel = gpioc.pc0.into_analog();
 
     loop {
-        let data: u32 = adc1.read(&mut channel).unwrap();
+        let data: u32 = adc3.read(&mut channel).unwrap();
         // voltage = reading * (vref/resolution)
-        println!(log, "ADC reading: {}, voltage for nucleo: {}", data, data as f32 * (3.3/65535.0));
+        println!(log, "ADC reading: {}, voltage for nucleo: {}", data, data as f32 * (3.3/adc3.max_sample() as f32));
     }
 }
--- a/examples/adc.rs
+++ b/examples/adc.rs
@@ -1,0 +1,64 @@
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+
+extern crate panic_itm;
+
+use cortex_m;
+use cortex_m_rt::entry;
+
+use stm32h7xx_hal::{
+    adc,
+    pac,
+    delay::Delay,
+    prelude::*,
+};
+
+use cortex_m_log::println;
+use cortex_m_log::{
+    destination::Itm, printer::itm::InterruptSync as InterruptSyncItm,
+};
+use cortex_m_semihosting::hprintln;
+
+#[entry]
+fn main() -> ! {
+    let cp = cortex_m::Peripherals::take().unwrap();
+    let dp = pac::Peripherals::take().unwrap();
+    let mut log = InterruptSyncItm::new(Itm::new(cp.ITM));
+
+    // Constrain and Freeze power
+    println!(log, "Setup PWR...                  ");
+    let pwr = dp.PWR.constrain();
+    let vos = pwr.freeze();
+
+    // Constrain and Freeze clock
+    println!(log, "Setup RCC...                  ");
+    let rcc = dp.RCC.constrain();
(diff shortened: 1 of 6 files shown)


astraw:
Again, I'm unfortunately too busy to test this right now but it looks good to me. @richardeoin @jordens any comments?

richardeoin:
Awesome, I can have a look at this over the next few days. I have a suggestion to use the enum definitions from `stm32h7` rather than repeating them here - will comment this on the code.

src/rcc.rs
```diff
@@ -204,6 +227,17 @@ impl APB4 {
...
 }
 
+/// RCC Domain 3 Kernel Clock Configuration Register
+pub struct D3CCIPR {
+    _0: (),
+}
+
+impl D3CCIPR {
+    pub(crate) fn constrain(&mut self) -> &rcc::D3CCIPR {
+        unsafe {&(*RCC::ptr()).d3ccipr}
```
richardeoin: The logic looks good, but I think the name "constrain" is misleading here. This word is used to mean a function that consumes the structure it is called on, and returns some other type that is a constrained representation. Eg. after calling `constrain` on `PWR` the user's code can't access `PWR` again. This function is nothing like that - it's an accessor to say "it's okay to modify D3CCIPR within this crate at any time".

Something like "kernel_clock_configuration_input" or "kernel_ccip" would be better

src/adc.rs
```diff
@@ -0,0 +1,680 @@
...
+/// ADC sampling resolution
+///
+/// Options for sampling resolution
+//
+// Refer to RM0433 Rev 6 - Chapter 24.2
+#[derive(Clone, Copy, Debug, PartialEq)]
+#[allow(non_camel_case_types)]
+pub enum AdcSampleResolution {
+    /// 16 bit resulution
+    B_16,
```
richardeoin: Here would be a good place to use the stm32-rs enum definition. You can do this like:

```rust
/// Resolution to sample at
pub type Resolution = crate::stm32::adc3::cfgr::RESR;
```

The definition is then the automatically generated one from [here.](https://github.com/stm32-rs/stm32-rs/blob/5142af62d53343c8110eaac1ff55d167c32b07c0/peripherals/adc/adc_h7_revision_y.yaml#L5)

This will save the enum definition and the From for u8 implementation. The default will then be `Resolution::SIXTEENBIT`.

jordens:
I'm swamped with proposal writing. I won't be able to help until next week.

src/adc.rs
hargoniX: I'd usually agree with you on this, however this specific piece of code is revision dependent as the config for the 8 bit resolution is dependent on the revision (as can be seen below), however this code is generated for revision y and would thus not be compatible with rev v

hargoniX:
(As my commit message says I fixed an error I made when translating my i2c implementation to this HAL, i messed up the order of a line which causes the I2C peripheral to end in reset state permanently)

src/adc.rs
richardeoin: The intention is that when stm32-rs does add support for Rev V, this crate will pass a `rev_v` feature gate back to stm32-rs. In that case, stm32-rs will have the correct definition.

This means that the difference between revisions only needs to be written and maintained once instead of twice.

Tagging tracking issue #9

hargoniX: Okay so I tried to implement this approach, however in the example codes this causes some problems:
```
error: enum variants on type aliases are experimental
  --> examples/adc.rs:49:25
   |
49 |     adc1.set_resolution(adc::Resolution::SIXTEENBIT);
   |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: add `#![feature(type_alias_enum_variants)]` to the crate attributes to enable

error: aborting due to previous error
```

jordens: Add `#![feature(type_alias_enum_variants)]` to each example as well. The lib has it already.

examples/adc.rs
```diff
@@ -0,0 +1,62 @@
...
+    // Constrain and Freeze power
+    println!(log, "Setup PWR...                  ");
+    let pwr = dp.PWR.constrain();
+    let vos = pwr.freeze();
+
+    // Constrain and Freeze clock
+    println!(log, "Setup RCC...                  ");
+    let rcc = dp.RCC.constrain();
+    // setting this per_ck to 4 mhz here as the last bit doesnt get sampled
+    // if we wouldn't do that
```
richardeoin: The example should explain that the driver selects `adc_ker_ck` = `per_ck` [here](https://github.com/stm32-rs/stm32h7xx-hal/pull/12/files#diff-a4bc73d1077eb9a314216894b4643500R462). Therefore `per_ck` must be configured by the user. Setting `per_ck` to 4MHz selects `CSI` on the `per_ck` input multiplex.

examples/adc.rs
```diff
@@ -0,0 +1,62 @@
...
+    let mut ccdr = rcc.sys_ck(100.mhz()).per_ck(4.mhz()).freeze(vos, &dp.SYSCFG);
+
+    println!(log, "");
+    println!(log, "stm32h7xx-hal example - ADC");
+    println!(log, "");
+
+    let mut delay = Delay::new(cp.SYST, ccdr.clocks);
+
+    // Setup ADC
+    let mut adc1 = adc::Adc::adc3(dp.ADC3, &mut delay, &mut ccdr);
```
richardeoin: Variable name should be `adc3` ?

examples/adc.rs
```diff
@@ -0,0 +1,62 @@
...
+    adc1.set_resolution(adc::Resolution::SIXTEENBIT);
+
+    // Setup GPIOB
+    let gpioc = dp.GPIOC.split(&mut ccdr.ahb4);
+
+    // Configure pb0, pb1 as an analog input
+    let mut channel = gpioc.pc0.into_analog();
```
richardeoin: Comments inconsistent with code

src/adc.rs
```diff
@@ -0,0 +1,636 @@
...
+                        panic!("Cannot read linear calibration value when the voltage regulator is disabled");
+                    }
+                    if self.rb.cr.read().aden().bit_is_clear() {
+                        panic!("Cannot read linear calibration value when the ADC is disabled");
+                    }
+                }
+
+                fn reset(&mut self, ahb: &mut $AHB) {
+                    ahb.rstr().modify(|_, w| w.$adcxrst().set_bit());
+                    ahb.rstr().modify(|_, w| w.$adcxrst().clear_bit());
```
richardeoin: The reset line `adc12en` is shared between ADC1 and ADC2. Therefore currently ADC1 and ADC2 can't be safely instantiated at the same time.
One solution would be to add a check in `init`, so that reset is only called if the AHB clock was previously disabled.

src/adc.rs
```diff
@@ -0,0 +1,636 @@
...
+                }
+
+                fn enable_clock(&mut self, ahb: &mut $AHB, d3ccipr: &mut D3CCIPR) {
+                    // Set per_ck as adc clock, TODO: we might want to change this so we can also
+                    // use other clocks as input for this
+                    d3ccipr.kernel_ccip().modify(|_, w| unsafe { w.adcsel().bits(0b10) });
+                    ahb.enr().modify(|_, w| w.$adcxen().set_bit());
```
richardeoin: If the adc_ker_ck is out of specification (DS12110 Rev 7 Table 87.), then the ADC will return bad results (with the LSBs always zero, I think you've seen this). Therefore it would be good to add an `assert!` here. For example

```rust
#[cfg(any(
    feature = "stm32h742",
    feature = "stm32h743",
    feature = "stm32h753",
    feature = "stm32h750"
))]
let adc_ker_ck_max = 36_000_000;

assert!(ccdr.clocks.per_ck().unwrap().0 <= adc_ker_ck_max,
        "per_ck is not running or too fast");
```

richardeoin:
I've added some more review comments, around helping users understand how to use this implementation.
The most complicated issue is the provision of a suitable `adc_ker_ck`. Annoyingly the chip doesn't do this by default. It should be explained in the example and checked with `assert!`

src/adc.rs
```diff
@@ -0,0 +1,636 @@
...
+                ///
+                /// LSHIFT[3:0] must be in range of 0..=15
+                pub fn set_lshift(&mut self, lshift: AdcLshift) {
+                    self.lshift = lshift;
+                }
+
+                /// Returns the largest possible sample value for the current settings
+                pub fn max_sample(&self) -> u32 {
+                    ((1 << self.get_resolution().bits() as u32) - 1) << self.get_lshift().value() as u32
+                }
```
richardeoin: This is broken (currently returns 0 for 16-bit). Try implementing a `number_of_bits` method to use here.

```rust
trait NumberOfBits {
    fn number_of_bits(&self) -> u32;
}
impl NumberOfBits for Resolution {
    fn number_of_bits(&self) -> u32 {
        match *self {
            Resolution::EIGHTBIT => 8,
            Resolution::TENBIT => 10,
            Resolution::TWELVEBIT => 12,
            Resolution::FOURTEENBIT => 14,
            _ => 16,
        }
    }
}
```

examples/adc.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    loop {
+        let data: u32 = adc1.read(&mut channel).unwrap();
+        // voltage = reading * (vref/resolution)
+        println!(log, "ADC reading: {}, voltage for nucleo: {}", data, data as f32 * (3.3/65535.0));
+    }
```
richardeoin: Use `max_sample()` instead of hard-coded 65535

richardeoin:
Great, nice work on getting these sorted so quickly @hargoniX 

The checking `ADC1`/`ADC2` enable before resetting isn't quite right. The check is after `enable_clock`, so the if statement will always evaluate to false. Maybe store the previous state in a variable?

Other than that small point, ready to merge.

hargoniX:
That issue is also fixed now I think, thanks for the feedback from your side!

PR #13, opened 2019-07-03
pwm: implementation and example

richardeoin:
* Add `GetClk` trait and implementations in `timer.rs` to return a
  timer's clock

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -4,14 +4,103 @@
 // TIM1 is 16 bit.
 
 use crate::hal::timer::{CountDown, Periodic};
-use crate::stm32::{TIM1, TIM2};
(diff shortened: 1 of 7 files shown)


src/timer.rs
```diff
@@ -4,18 +4,107 @@
...
 use void::Void;
 
-use crate::rcc::{CoreClocks, APB1, APB2};
+use crate::rcc::Ccdr;
+use crate::rcc::{APB1, APB2};
+use crate::stm32::rcc::{d2ccip2r, d3ccipr};
 use crate::time::Hertz;
 
+/// Associate clocks with timers
+pub trait GetClk {
```
jordens: This sounds like it would benefit from being and is generic enough to be moved into `embedded-hal`. Together with the `Ccdr` construction.

PR #14, opened 2019-07-03
rng: add implementation and example

richardeoin:
* From @dotcypress's `stm32g0xx-hal` with thanks
* Enable HSI48

### DIFF
--- a/src/rcc.rs
+++ b/src/rcc.rs
@@ -71,6 +71,8 @@ pub struct Ccdr {
     pub clocks: CoreClocks,
     /// AMBA High-performance Bus (AHB1) registers
     pub ahb1: AHB1,
+    /// AMBA High-performance Bus (AHB2) registers
+    pub ahb2: AHB2,
     /// AMBA High-performance Bus (AHB3) registers
     pub ahb3: AHB3,
     /// AMBA High-performance Bus (AHB4) registers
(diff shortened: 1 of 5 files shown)


src/rng.rs
```diff
@@ -0,0 +1,175 @@
...
+                    let val = self.next()?;
+                    Ok(val as $type)
+                }
+
+                /// Returns a single element with random value in range, or error
+                fn gen_range(&mut self, low: $type, high: $type) -> Result<$type, ErrorKind> {
+                    assert!(high > low);
+                    let range = high - low;
+                    let val: $type = self.gen()?;
+                    Ok(low + val % range)
```
jordens: Are you sure this is correct i.e. unbiased/"random"? This `gen_range(0u8, 196u8)` will return a value from 0-63 more often than 64-195.
Random numbers are not easy. I would gravitate towards passing any functionality that looks remotely complicated to a special random crate that knows how to do that.

dotcypress: @jordens could you explain why `gen_range(0u8, 196u8)` will return a value from 0-63 more often than 64-195?

jordens: From the possible values of a `u8`, both ranges `0:64` and `196:256` will be mapped to `0:64`, while only `64:196` will be mapped to `64:196`. Each number in `0:64` will have `1/128` probability of occurring, while each number in `64:196` has only `1/256` probability of occurring.

richardeoin: It'd certainly be preferrable for these functions to be implemented on a trait in `embedded-hal`, or in a specialised crate. In the meantime they could go here and in `g0xx-hal` as a demonstration?

Worth fixing the modulo bias either way.

jordens: I would rather drop that higher level functionality. It's currently wrong. It also seems unwise to commit to providing functionality in the HAL that has not been abstracted or may never be abstracted. It defeats the purpose of the HAL and chains this crate to providing that functionality while `embedded-hal` may move on. You can't really remove it without breaking people's code and it's rather likely that the API will be changed when `embedded-hal`-ified. I would avoid collecting unproven functionality in the HAL that's outside `embedded-hal`.

src/rcc.rs
```diff
@@ -111,8 +113,27 @@ impl AHB1 {
         // NOTE(unsafe) this proxy grants exclusive access to this register
         unsafe { &(*RCC::ptr()).ahb1rstr }
     }
+}
+
+/// AMBA High-performance Bus (AHB) peripheral registers
+pub struct AHB2 {
```
hargoniX: Considering we're now implementing all AHB and APB busses one by one by basically writing out the same code again and again with changing numbers after the AHB/APB it might be worth turning these definitions and impls into a macro

src/rng.rs
richardeoin: Removed in another commit

src/rcc.rs
richardeoin: Agreed. However this doesn't belong in this PR, I will create an issue instead.

src/rng.rs
jordens: Thanks!

Issue #6, opened 2019-06-13
Doesn't build on stable Rust


richardeoin:
Currently nightly is required for `type_alias_enum_variants`.

## `type_alias_enum_variants`

[Tracking issue](https://github.com/rust-lang/rust/issues/49683)
[Stabilisation PR](https://github.com/rust-lang/rust/pull/61682)

jordens:
Stabilized in https://github.com/rust-lang/rust/commit/5748825cc8c74cccef0059cdd4043e6e9b4aa188. Will appear in Rust 1.37.0

jordens:
Now that there is a 1.37 beta we can use that in CI.

Issue #5, opened 2019-06-13
Local copy of latest stm32-rs required to build


richardeoin:
Waiting on an [upstream](https://github.com/stm32-rs/stm32-rs) release.

jordens:
@richardeoin @astraw @hargoniX do you have pending breaking changes for stm32-rs that you would like to land before 0.8 or can we try pushing for that release to get this out of the way? I don't have anything pending for 0.8.

hargoniX:
I don't have done anything related to stm32-rs yet so that's a clear no from me

astraw:
Also nothing pending from me.

richardeoin:
@jordens I've got two PRs open

https://github.com/stm32-rs/stm32-rs/pull/241
https://github.com/stm32-rs/stm32-rs/pull/246

Would be good to land them for 0.8

PR #15, opened 2019-07-03
add travis-ci and bors config

jordens:
Once a new stm32-rs 0.8 with the current pending breaking stm32h7 changes is released (#5) the hack to locally build and use that crate can be removed.
Stable/beta builds will only succeed once #6 is resolved.

I would still suggest merging this as it already exposes, e.g. https://travis-ci.com/stm32-rs/stm32h7xx-hal/jobs/213079743#L354-L408


richardeoin:
Great, looking forward to having this working

PR #16, opened 2019-07-03
Add ADC3 internal inputs and example for internal temperature

richardeoin:


src/signature.rs
```diff
@@ -0,0 +1,93 @@
+//! Device electronic signature
+//!
+//! (stored in system flash memory)
+
+/// This is the test voltage, in millivolts of the calibration done at
+/// the factory
+pub const VDDA_CALIB: u32 = 3300;
+
+macro_rules! define_ptr_type {
```
jordens: Aren't these locations something that should go into the PAC crate? The interface is "register-like" and other crates like `stm32ral` would benefit just as well.

richardeoin: There's certainly an argument for that, it's just that ST don't ever put them in SVDs (since they're not a 'peripheral'). So my slight preference is to leave them out of the PAC crate and implement like this.

This solution was inspired by the [f4](https://github.com/stm32-rs/stm32f4xx-hal/pull/89) crate, but there's more direct ways of doing it (like [f0](https://github.com/stm32-rs/stm32f0xx-hal/blob/master/src/adc.rs#L40)).

jordens: Ok. If there is precedent. Don't know whether the f0 style being more compact or the f4 style of being more functional is better.

richardeoin: Also relevant https://github.com/stm32-rs/stm32-rs/issues/245

src/adc.rs
```diff
@@ -149,6 +151,42 @@ macro_rules! adc_pins {
...
+macro_rules! adc_internal {
+    ($($input:ty => ($chan:expr, $en:ident)),+ $(,)*) => {
+        $(
+            impl $input {
+                pub fn new() -> Self {
+                    Self {}
+                }
+
+                /// Enables the internal voltage/sensor
+                /// ADC must be disabled.
```
hargoniX: Is it really a must have requirement for the ADC to be disabled for this to work? If so why not try and follow a design pattern based on type states like for the GPIO in order to enforce this at compile time.

src/adc.rs
```diff
@@ -149,6 +151,42 @@ macro_rules! adc_pins {
...
+                pub fn enable(&mut self) {
+                    let common = unsafe { &*ADC3_COMMON::ptr() };
+
+                    common.ccr.modify(|_, w| w.$en().enabled());
+                }
+                /// Disables the internal voltage/sdissor
+                /// ADC must be disabled.
```
hargoniX: Same here

src/adc.rs
richardeoin: Great point, the answer is yes the ADC should be disabled. RM0433 page 1028 says:
>Note: The software is allowed to write this bit only when the ADCs are disabled (ADCAL=0, JADSTART=0, ADSTART=0, ADSTP=0, ADDIS=0 and ADEN=0).

richardeoin:
Implemented Enabled/Disabled type states!

bors try

richardeoin:
I will merge shortly if no further feedback

PR #19, opened 2019-07-12
timers, bugfix: fix periodic CountDown timer

richardeoin:
Previously the `CountDown` timer failed to start a new countdown if the periodic timer was running and no `wait()` call was made in the previous period. This was as the UIF bit was already set.

See embedded-hal docs for the contract:
https://docs.rs/embedded-hal/0.2.3/embedded_hal/timer/trait.CountDown.html#tymethod.start

Add  example to demonstrate correct behaviour.

### DIFF
--- a/examples/blinky_timer.rs
+++ b/examples/blinky_timer.rs
@@ -1,0 +1,66 @@
+#![deny(warnings)]
+#![deny(unsafe_code)]
+#![no_main]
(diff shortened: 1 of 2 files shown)


richardeoin:
CI fails due to #20

richardeoin:
The timer example could be the template for #24

PR #23, opened 2019-07-17
 rcc: write accessors for AHB and APB busses in a macro

### LINKED ISSUE #21: rcc: write accessors for AHB and APB busses in a macro

hargoniX:
Closes #21

### DIFF
--- a/src/pwm.rs
+++ b/src/pwm.rs
@@ -463,7 +463,7 @@ macro_rules! pwm_ext_hal {
 
 // Implement PWM configuration for timer
 macro_rules! tim_hal {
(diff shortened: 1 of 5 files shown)


richardeoin:
LGTM :thumbsup:

It is actually possible to have different documentation for each iteration of a macro. With a macro variable like `$doc:expr` (which is a string):
```
#[doc=$doc]
pub struct $AXBn {
```

hargoniX:
I'm actually not sure what's going on with the CI right now, or rather why this worked before if its an error now.

jordens:
#25 needs to land first (or you'd have to target an old `stm32-rs`/`stm32h7`).

richardeoin:
Okay, #25 has gone in.

bors try

jordens:
Bors try again

PR #25, opened 2019-07-21
revision_v: Update feature gates from revision v split in PAC

richardeoin:
Upstream change in stm32-rs
https://github.com/stm32-rs/stm32-rs/pull/247

Remove feature gates where they are not required.

Add new v feature targets to travis.

ADC is currently the only area where this crate differs between v and
non-v.

Revision v tracking issue
https://github.com/stm32-rs/stm32h7xx-hal/issues/9

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -17,13 +17,11 @@ use crate::gpio::Analog;
 use crate::rcc::Ccdr;
 use crate::rcc::D3CCIPR;
 
-#[cfg(any(
-    feature = "stm32h742",
-    feature = "stm32h743",
-    feature = "stm32h753",
-    feature = "stm32h750"
-))]
+#[cfg(not(feature = "revision_v"))]
 const ADC_KER_CK_MAX: u32 = 36_000_000;
+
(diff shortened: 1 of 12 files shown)


src/adc.rs
```diff
@@ -17,14 +17,12 @@ use crate::gpio::Analog;
 use crate::rcc::Ccdr;
 use crate::rcc::D3CCIPR;
 
-#[cfg(any(
-    feature = "stm32h742",
-    feature = "stm32h743",
-    feature = "stm32h753",
-    feature = "stm32h750"
-))]
+#[cfg(not(feature = "revision_y"))]
```
jordens: not `revision_v` i guess

src/adc.rs
```diff
@@ -17,14 +17,12 @@ use crate::gpio::Analog;
...
 const ADC_KER_CK_MAX: u32 = 36_000_000;
 
+#[cfg(feature = "revision_y")]
```
jordens: same

richardeoin:
Sorted. I also added details about Rev V in the README.

bors r+

PR #26, opened 2019-07-21
rcc: add options to configure more PLL outputs

richardeoin:
* All 3 outputs of PLL1 can be configured.
* Moving closer to a generic interface for PLL2/PLL3; just needs macro
  definitions and call to setup.
* An `assert` will occour if pll1_p_ck is set independently when it is
  requried for sys_ck.
* When `pll1_r_ck` is `None` in the configuration struct, it is set if
  required to keep `traceclk` running. Afaict there's no documentation
  on acceptable frequencies for `traceclk`, set p_ck/2 as a sensible
  choice (can be overridden by setting some `pll1_r_ck`).

### DIFF
--- a/examples/spi.rs
+++ b/examples/spi.rs
@@ -30,7 +30,10 @@ fn main() -> ! {
     // Constrain and Freeze clock
     println!(log, "Setup RCC...                  ");
     let rcc = dp.RCC.constrain();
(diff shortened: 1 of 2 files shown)


richardeoin:
bors try

richardeoin:
Anyhow, it's okay since #25 was merged.

Ready for review.

PR #27, opened 2019-07-23
Init traits prelude

### LINKED ISSUE #20: Timers not included in prelude

richardeoin:
adc: Add initialisation trait

Simplifies usage:

```rust
let _ = adc::Adc::adc3(dp.ADC3, &mut delay, &mut ccdr);
```
to
```rust
let _ = dp.ADC3.adc(&mut delay, &mut ccdr);
```

Previous method still works.

Re-work arguments for timer, doesn't break anything as there aren't any examples yet.

Closes #20

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -270,6 +270,10 @@ adc_internal!(
           Vrefint => (19, vrefen)
 );
 
+pub trait AdcExt<ADC>: Sized {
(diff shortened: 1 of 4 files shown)


hargoniX:
If we're already changing the Timer API this way, shouldn't we extend this to all peripherals in order to keep some sort of consistency?

richardeoin:
Absolutely, this should be consistent for all peripherals. From the peripherals with multiple instances, I think it's just `i2c` remaining once `adc` and `timer` are updated here.

PR #28, opened 2019-07-27
Porting i2c to the extension trait pattern

hargoniX:
See #27

### DIFF
--- a/examples/i2c.rs
+++ b/examples/i2c.rs
@@ -3,7 +3,7 @@
 
 extern crate panic_itm;
 
-use stm32h7xx_hal::{i2c::I2c, pac, prelude::*};
+use stm32h7xx_hal::{pac, prelude::*};
 
 use cortex_m_rt::entry;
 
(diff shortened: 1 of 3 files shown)


richardeoin:
LGTM

I can't reproduce the CI failure though. Let's try re-running it.

richardeoin:
The CI fail looks like a breakage from svd2rust 0.15.0 ([changelog](https://github.com/rust-embedded/svd2rust/blob/master/CHANGELOG.md#v0150---2019-07-25))

> * [breaking-change] for access to alternate registers functions now used instead of untagged_unions (no more nightly features)

hargoniX:
Also as this is apparently 0.2.0 now I wanted to add another thought, the busy_wait macro in this implementation, responsible for waiting for an ACK on the bus would send the program to an infinite loop if there was no ACK present, it would probably be better to implement it so it actually throws an error, however I'm not exactly sure how to do that as this is the only way proposed in the reference manual for ACK detection I could find

richardeoin:
@hargoniX : The meaning of the milestone is something that should resolved before 0.2.0 is released. These changes look good and are simple, so merging now.

The ACK issue is interesting, there's definitely an argument for passing a `delay`/`timer` so the HAL can timeout. It should be the subject of another Issue/PR though.

bors r+

PR #41, opened 2019-10-22
Fix clear_interrupt_pending_bit()

Windfisch:
According to RM0433, sec. 19.6.24, a "bit is cleared by writing a 1 into the bit or by changing the sensitivity of the edge detector."

I have not tested nor compiled this on stm32h7xx. However, I found this issue while porting the ExtiPin trait to stm32f1xx. I have tested this change on my stm32f1xx-hal working copy.

See also stm32-rs/stm32f4xx-hal#114

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -97,7 +97,7 @@ pub trait ExtiPin {
     fn trigger_on_edge(&mut self, exti: &mut EXTI, level: Edge);
     fn enable_interrupt(&mut self, exti: &mut EXTI);
     fn disable_interrupt(&mut self, exti: &mut EXTI);
-    fn clear_interrupt_pending_bit(&mut self, exti: &mut EXTI);
+    fn clear_interrupt_pending_bit(&mut self);
 }
 
 macro_rules! gpio {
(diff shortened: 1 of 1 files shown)


Windfisch:
Seems like the Travis CI is broken; it fails with `error: failed to download stm32h7 v0.8.0`, which is ultimately caused by `failed to unpack /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/stm32h7-0.8.0/src/stm32h743/dfsdm/dfsdm3_cr1.rs`

Maybe an out-of-memory?

The very same command works on my machine locally.

richardeoin:
Thank you for the PR @Windfisch , and for picking up @therealprof 's review on the other thread.

I'm unclear why the `&mut` reference to EXTI argument has been removed. This removes the requirement on the caller to own the EXTI peripheral. Is this because `clear_interrupt_pending_bit()` is now completely idempotent?

The build failure looks to be transient, I'll trigger a re-run.

bors try

therealprof:
> I'm unclear why the &mut reference to EXTI argument has been removed. This removes the requirement on the caller to own the EXTI peripheral. Is this because clear_interrupt_pending_bit() is now completely idempotent?

Yes, idempotent and atomic so it's safe to call at any time.

PR #43, opened 2019-12-08
i2c: flush the TXDR register when a NACK condition occurs

richardeoin:
Without this, the state of the TXIS flag for the next transactions is
undefined.


richardeoin:
I think this is pretty uncontroversial, as the only behaviour change is what happens after a NACK. And it works well on hardware

bors r+

PR #44, opened 2019-12-08
Iterative PLL strategy

richardeoin:
Add alternative PLL configuration strategy, which can be selected instead of the current default.

This strategy uses the high frequency VCO, and iterates the M divider value to give an accurate frequency on the P output.

**Normal**: VCOL, highest PFD frequency, highest VCO frequency
**Iterative**: VCOH, choose PFD frequency for accuracy, highest VCO frequency


richardeoin:
Merging as the default behaviour is unchanged, and it's useful downstream

bors r+

PR #45, opened 2019-12-09
gpio: initialise gpio in the correct typestate

richardeoin:
See reset state of GPIOx_MODER, RM0433 Section 10.4.1

Different reset values apply for JTMS, JTCK, JTDI, JTDO, !JTRST


hargoniX:
LGTM! Just did a cross check again in case you misread something but it looks perfectly fine to me. Probably saved some poor soul in the future from some weird error with their GPIO config so a great spot!

richardeoin:
Ok, let's merge!

bors r+

PR #46, opened 2019-12-09
timer: add method to clear interrupt flag

richardeoin:


richardeoin:
This change is trivial, but longer term it would be good to have a standard listen/unlisten/clear functions for all peripherals from a usability perspective

PR #47, opened 2019-12-14
Add HAL for Extended interrupt and event controller (EXTI)

richardeoin:


richardeoin:
Needs fix in stm32-rs

richardeoin:
bors try

PR #48, opened 2019-12-20
Adds DMA implementation

Jan561:
This PR adds a DMA implementation with the following features:

-  [x] DMA1 and DMA2 support
-  [x] DMAMUX1 support
-  [x] 16 Streams (8 for DMA1, 8 for DMA2)
-  [x] 16 Mux-Channels (8 for DMA1, 8 for DMA2)
-  [x] 8 Request Generators
-  [x] Transfer module for memory safe DMA transfers (Not completed yet)
-  [ ] Implementations of Transfer for all peripherals

TODO
-  [x] Extension trait
-  [ ] Documentation
-  [ ] Support for other MCUs (currently only STM32H743 is supported)
-  [ ] Enable clippy warnings for the whole crate again

Discussion
- As the DMA is not an official embedded-hal item, we should consider making this an opt-in feature.

**Note**: Write access on my fork is enabled for collaborators. Collaboration is appreciated.


Jan561:
There are edited files completely unrelated to the DMA. These are edits of `cargo fmt`. I know this isn't great, but opening a PR just for formatting some files correctly isn't ideal either.
Should we keep these changes in this PR or revert them at the end and merge them separately?

richardeoin:
Looks good, at some point I will have a look and try this out.

There's some documentation of the register fields happening in stm32-rs which may be useful https://github.com/stm32-rs/stm32-rs/pull/322

Is there an existing DMA HAL that this is based on, maybe in one of the other stm32 crates?

Jan561:
This is currently a standalone implementation, not supported by any HAL I'm aware of, nor it is based on implementations of other HALs of the stm32-rs group.

My aim is to fulfill [these requirements] to make it as compatible as possible with embedded-hal, when the DMA api gets included (or not ...).

@richardeoin

[these requirements]: https://github.com/rust-embedded/embedded-hal/issues/37

Jan561:
There is a problem with my current SafeTransfer-Implementation: It requires buffers with static lifetimes. But such buffers (when created safely) get allocated in .bss, which itself gets allocated in DTCM-RAM. The problem is that the DMA can't access DTCM.

(Actually everything gets allocated in DTCM-RAM currently, maybe we should allocate the stack in SRAM1)

A possible fix would be to allow non-static lifetimes for the buffers and implement Drop implementations for SafeTransfer, which disables the stream, and for the buffers, which panics if the stream is still running so the buffer won't get deallocated.

Do you have other suggestions? Let me hear.

Merry Christmas btw :)

richardeoin:
How is this progressing? There's a huge amount of functionality here, and it's interesting how it compares to other implementations (eg. [F7](https://github.com/stm32-rs/stm32f7xx-hal/blob/master/src/dma.rs))

More examples and more documentation are definitely needed. A paragraph at the top of each file (`//!`) would be very helpful.

I see the change to the memory map, and you're right that the 'memory' side of a DMA transfer will be the AXISRAM (or SRAM1,2,3). But moving the stack out of DTCM will hurt performance for all applications. The better option is to add an attribute the to static mut buffers as required, like this:
```
#[link_section = ".axisram.buffers"]
```

Jan561:
I was quite busy learning for exams for the last month so sadly i didnt make any progress since then.
But good news is that i have now semester vacation until end of april so i have now time to make progress here again.

The lack of documentation is for sure the first thing i'll address to enable collaboration and better reviewing.

I've never heard of this macro before, but what are u suggesting looks awesome. I'll look into it soon.

Jan561:
So quick question, as you probably noticed I have several `effective_CONFIG` methods defined for `dma::Stream` additionally to CONFIG. The reason for this is because the **hardware** forces these configs to a specific value when the stream gets enabled because otherwise the config would be invalid (see `RM 0433 Rev 7 Chapter 15.3.18, 15.5.5` for details).

I personally dislike this behaviour, although its well documented in the reference manual.
A workaround would be to catch these situations and panic, forcing the **software** to manually fixing the config. Or we could add a `Config` builder, which guarantees valid configs at type level, but could end up bit messy and not easy to use. Or we leave it like this.

richardeoin:
I see, the hardware sets the register value so it is always a valid configuration. For returning the current state, then try to read it from the hardware where possible.
For setting a valid configuration, I'd try prototyping a builder for a `Config` type. I don't think it will be that complicated, and one of the benefits of the builder type is that it allows you to write the logic to return None (or panic!) based on the values already set.

Jan561:
The safe_transfer module is currently not really usable, and not intended to be used directly. There will be implementations of `SafeTransfer` for all peripherals, which take part of the peripheral configuration of the channel. I recently pushed an example into `serial::dma` for demonstration purposes. Note that this module is configured as a `test`-module to exclude it from compilation, as the implementation is based on an old version of `SafeTransfer` and has very likely many bugs.

Jan561:
Closing as this PR is getting really cluttered with commits. When most of the work is done, I'll reopen / open an new PR.

PR #49, opened 2020-01-07
i2c: wait for end of previous address phase before setting START

richardeoin:
It appears to take some additional cycles of PCLK to end the previous
address sequence, after we may have returned with an ACK or NACK. Wait
for this. Once the START bit is clear, we can set it even whilst the
bus is busy, as explicitly stated in a note by the START bit in the RM.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -232,7 +232,14 @@ macro_rules! i2c {
                     // TODO support transfers of more than 255 bytes
                     assert!(bytes.len() < 256 && bytes.len() > 0);
 
-                    // START and prepare to send `bytes`
+                    // Wait for any previous address sequence to end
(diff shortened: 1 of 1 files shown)


Jan561:
looks good, but perhaps a more flexible approach would be to use `nb` instead, returning `WouldBlock` if it cannot be executed right now (instead of while loops). This would be a breaking change though, requiring the caller to e.g. `block!` the call before matching its result.

richardeoin:
I had a look at this in more detail. The extra delay added is up to 50% of a I2C bus cycle, ie. if the bus is running at 100kHz the delay would be up to 5µs. I've updated the comments and commit message to note this.

Using `nb` for this one delay doesn't make sense, but the comment should give enough context if/when `nb` is added in future.

PR #50, opened 2020-02-02
pwr: configure core supply on parts with SMPS (dual core)

richardeoin:
Changes the default behaviour for dual core parts from enabling the
SMPS to keeping the default configuration. This prevents the core
voltage being cut if the board uses a different power supply scheme -
eg. LDO. Typically the default configuration will result in an
infinite loop waiting for `actvosrdy`, but this allows the part to be
re-programmed.

Unsafe builder methods are the added to set the correct core supply
scheme for the board.


richardeoin:
Merging now, this is an improvement for dual core and hopefully no change for single core

bors r+

PR #54, opened 2020-03-13
Fixes for i2c:

tstellanova:
- Use TXE to detect transmission finished in `write_read`
- Allow `busy_wait` to wait on variants of isr flags
- `busy_wait` returns immediately on success
- run rustfmt 

Tested with the following embedded hal i2c drivers:
- bmp280 (barometer)
- ssd1306 (oled display)
- bno080 (ahrs wip)

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -68,30 +68,29 @@ macro_rules! flush_txdr {
         }
 
         // If TXDR is not flagged as empty, write 1 to flush it
-        if $i2c.isr.read().txe().bit_is_clear() {
+        if $i2c.isr.read().txe().is_not_empty() {
             $i2c.isr.write(|w| w.txe().set_bit());
         }
     };
 }
 
-// Wait until specified flag is raised, or an error condition occours
 macro_rules! busy_wait {
-    ($i2c:expr, $flag:ident) => {
+    ($i2c:expr, $flag:ident, $variant:ident) => {
         loop {
             let isr = $i2c.isr.read();
 
-            if isr.berr().bit_is_set() {
+            if isr.$flag().$variant() {
+                break;
+            } else if isr.berr().is_error() {
                 $i2c.icr.write(|w| w.berrcf().set_bit());
                 return Err(Error::Bus);
-            } else if isr.arlo().bit_is_set() {
+            } else if isr.arlo().is_lost() {
                 $i2c.icr.write(|w| w.arlocf().set_bit());
                 return Err(Error::Arbitration);
             } else if isr.nackf().bit_is_set() {
                 $i2c.icr.write(|w| w.stopcf().set_bit().nackcf().set_bit());
                 flush_txdr!($i2c);
                 return Err(Error::NotAcknowledge);
-            } else if isr.$flag().bit_is_set() {
-                break;
             } else {
                 // try again
             }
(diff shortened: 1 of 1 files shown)


richardeoin:
Thanks for the PR!

* The readability improvements are great.
* Re-ordering the `busy_wait` loop potentially means missing ARLO and BERR flags. We don't (yet) have hardware unit tests for this behaviour. Could you add a comment to explain the change? 
* From my reading of RM0433, detecting the end of transmission on TXE is equally valid as TC. Is there a reason for this change? Could you reference a H7 RM instead (RM0433 / RM0399) or explain why another RM is applicable with 100% certainty? Also TXE is set *high* on stop, perhaps just mention the `is_empty` state in the comment.

I'm willing to merge this with more explanation. Does master not work with one of those devices?

tstellanova:
Thank you for the feedback.  Yes, master `write_read` was specifically failing with the bno080 sensor, on both stm32h7 and stm32f3, which lead to my investigation. When I'm back at my desk I'll attach an analyzer trace, as well as update comments with more info. 

I'm not too attached to the ordering of the `busy_wait` checks. In my mind the purpose of that macro is to first succeed if the flag/variant is satisfied and secondarily check, if the criterion is not satisfied, if there are peripheral errors to handle. I made the change to crystallize those priorities. Let me know if you'd like me to revert that.

tstellanova:
An example capture of a failed `write_read` failed_write_read_bno080

richardeoin:
Ok great, I understand the reason for changing the `busy_wait` check. I don't think it matters much anyhow because the loop is likely polling very fast. Thanks for fixing the RM reference

Looking at the bno080 datasheet (ref 1000-3927) in more detail, it says:
> The BNO080 uses Hillcrest’s SHTP (Sensor Hub Transport Protocol) protocol to communicate. The BNO080 application does not support the repeated start method for typical I2C register based interfaces. More details are available in [2].

The `write_read` method is explicitly designed to issue a repeated start (SR). It's an implementation of the `embedded_hal::blocking::i2c::WriteRead` trait, and the [contract](https://docs.rs/embedded-hal/0.2.3/embedded_hal/blocking/i2c/trait.WriteRead.html#i2c-events-contract) for that describes a repeated start. This is the least surprising behaviour imo, at least for experienced I2C people.

If the change makes `write_read` work with the bno080, then that's actually a problem! Our goal here is to implement the embedded_hal trait as closely as possible.

richardeoin:
This PR does actually remove the repeated start (SR). Here's my capture with a ft6206 touchscreen controller. Both cases work fine for the application.

---
master

Screenshot 2020-03-14 at 23 06 32

---
this PR at a5c608a3b2c8d883102e85aa2a722514278c0b73

Screenshot 2020-03-14 at 22 09 40

tstellanova:
Good catch, and I appreciate the analyzer recapture.  I’ll update the driver and retest to confirm and then possibly resubmit this PR with just the readable variants cleanup.

tstellanova:
Updated the driver to use `write` followed by `read` rather than `write_read` and this seems to work well.  Updated the PR to just include the macro changes.  Thanks again for the sanity check.

richardeoin:
Thanks!

bors r+

PR #56, opened 2020-03-19
Improve README.md, add minimum supported rust version

richardeoin:
Update README to be clearer and more useful to those new to embedded rust. Prefer linking to more general embedded rust resources instead of giving build instructions ourselves. [Rendered here](https://github.com/richardeoin/stm32h7xx-hal/blob/readme1/README.md)

Add minimum supported rust version (MSRV).

Stop running run CI against beta rust.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: rust
 
 rust:
-  - nightly
-  - beta
+  - 1.39.0
   - stable
+  - nightly
 
 cache: cargo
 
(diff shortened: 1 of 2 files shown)


astraw:
Looks great.

On the topic of easy onboarding: I think targeting easy-to-acquire hardware for examples would make sense, also to document in the readme. I know of the nucleo board with h743, but nothing else.

richardeoin:
Agreed, I've added a hardware list!

bors r+

PR #57, opened 2020-03-21
Fix invalid docs.rs `targets` syntax

jyn514:
The `target` field does nothing. There are two options you could have been trying to set:

- `default-target`, the target built for the main page (docs.rs/stm32h7xx-hal)
- `targets`, which controls all targets built.

This changes it to the second since I assume `stm32h7xx-hal` is the only target you want to build, but I can change it to the first if you want to build all tier one platforms instead.


richardeoin:
Thanks! You're correct, this crate is targeting `thumbv7em-none-eabihf` only.

bors r+

PR #58, opened 2020-03-21
Add some tests for RCC! These run on the host.

richardeoin:
I'm not entirely sure cargo + travis can do this, but this is a first
attempt. The numerical behaviour of the host platform might not quite
correspond to the target.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ matrix:
 
 script:
   - RUSTFLAGS="-D warnings" cargo build --verbose --release --examples --features rt,$MCU
+  - cargo test --lib --features rt,$MCU --target x86_64-unknown-linux-gnu
 
 notifications:
   email: false
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,4 +1,4 @@
-#![no_std]
+#![cfg_attr(not(test), no_std)]
 #![allow(non_camel_case_types)]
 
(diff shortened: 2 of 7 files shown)


hargoniX:
With over 1k lines the RCC file is already pretty fat imo. Should we really make use of the rust feature to have tests in the same file as the productive code in this case?

richardeoin:
Yes, rcc.rs could do with being broken up. However writing inline was an easy way to test macros in this case, as it avoids using `#[macro_use]` to get access to the macros elsewhere.

Perhaps we could break up rcc.rs by functionality, as has been done [here](https://github.com/stm32-rs/stm32g4xx-hal/tree/master/src/rcc). Particularly the PLL functionality could go to a separate file.

richardeoin:
Do the test pass on travis now?

bors try

richardeoin:
I've broken up rcc.rs into multiple files, with the largest less than 1k lines

hargoniX:
LGTM! it would be suuuuuper cool if we got tests bases on utest running on real world hardware at some point. (Just in case you wanna look further into testing)

richardeoin:
Agreed, unit test running on real hardware would be ideal. Is there any hardware platform that has good peripheral coverage?

richardeoin:
That doesn't impact this PR, so merging now

bors r+

PR #59, opened 2020-03-25
Clear usart errors after reading, implement default blocking write 

tstellanova:
- Clear serial errors in ICR  after reading in serial::Read
- Implement marker trait to take advantage of blocking write_n implementation from embedded hal


richardeoin:
Looks good to me! I see no reason not to implement the marker trait.

bors r+

PR #60, opened 2020-03-27
bugfix i2c: Fix reference to I2C1 in macro generic for all I2Cs

richardeoin:
Also remove unsafe keyword on traits that don't need it.

After this the i2c hal is completely free of `unsafe`.


richardeoin:
Useful bugfix

bors r+

PR #61, opened 2020-03-27
spi: Documentation, support 16-bit words also

richardeoin:
Now requires type annotation / hinting somewhere


richardeoin:
Note: more changes to SPI expected, see #68

bors r+

Issue #66, opened 2020-04-18
Expose RCC registers for external usage


ryan-summers:
When the HAL is used in an application, there are (many) peripherals that are not yet supported by the HAL (such as ethernet, quadspi, DMA, etc.). Often, these peripherals may need a custom implementation for performance or specific use-case considerations. 

The `Rcc::Ccdr` should expose the underlying RCC register block publicly so that other drivers may modify the necessary RCC registers to enable clocks and reset peripherals as needed. This is currently done only for the HAL crate, but it may be appropriate to allow this block to be exposed for all usage. 

As currently implemented, the HAL takes exclusive ownership of the RCC registers, so it's impossible to implement application-specific drivers outside of the HAL alongside drivers that utilize the HAL implementation.

richardeoin:
Indeed, many peripherals don't have embedded-hal traits and hence don't really belong here. The DMA is probably one exception because it tightly integrates with other peripherals, and there is a WIP PR for that in https://github.com/stm32-rs/stm32h7xx-hal/pull/48

Others work well as external crates, like [stm32h7-ethernet](https://crates.io/crates/stm32h7-ethernet) (Although @jordens et al's implementation in stabilizer looks really good too!).

External crates can already modify the RCC registers as needed with `unsafe`, stm32h7-ethernet does that [here](https://docs.rs/stm32h7-ethernet/0.1.1/src/stm32h7_ethernet/ethernet.rs.html#419).

ryan-summers:
Although we can already access the registers using `unsafe` as mentioned, is that how we would like external drivers to access the register blocks? It would bypass the ownership concept of the register block, but it seems fine for the use case of the RCC. The main point takeaway I have is that external drivers may want the `Ccdr` structure anyways to access the configured clock frequencies and the register block is already contained in there, so it seems a bit odd to have to unsafely re-instantiate a pointer to the registers just to modify them.

richardeoin:
We really don't want external crates to access the RCC register block, as they could do anything! (change bus clocks, reset peripherals, even reset the core). Unfortunately with the current PAC we can't stop external crates doing these things with unsafe. (There are some [ideas](https://github.com/japaric/embedded2020/tree/master/host/regen#singletons-are-always-singletons) about changing this, but not likely any time soon).

[...]

PR #67, opened 2020-04-18
A safe interface for the user and other crates to access the RCC

richardeoin:
Safe access to enable, disable and reset peripherals that are not used in this hal.

Uses a macro with https://github.com/dtolnay/paste for brevity.

From discussion on #66

### DIFF
--- a/examples/safe_enable_fdcan.rs
+++ b/examples/safe_enable_fdcan.rs
@@ -1,0 +1,36 @@
+#![deny(warnings)]
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+extern crate panic_itm;
+
+use cortex_m_rt::entry;
+use stm32h7xx_hal::rcc::{rec, ResetEnable};
+use stm32h7xx_hal::{pac, prelude::*};
+
+fn enable_fdcan(rec: rec::Fdcan) {
+    rec.enable();
+
+    // rec is dropped here, and can never be changed again
+}
+
+#[entry]
+fn main() -> ! {
+    let dp = pac::Peripherals::take().unwrap();
+
+    // Constrain and Freeze power
+    let pwr = dp.PWR.constrain();
+    let vos = pwr.freeze();
+
+    // Constrain and Freeze clock
+    let rcc = dp.RCC.constrain();
+    let prec = rcc.take_peripherals().unwrap();
+    let _ccdr = rcc.sys_ck(100.mhz()).freeze(vos, &dp.SYSCFG);
+
+    enable_fdcan(prec.FDCAN);
+
+    //prec.FDCAN.disable(); // Compile error
+
+    loop {}
+}
--- a/examples/safe_enable_fdcan.rs
+++ b/examples/safe_enable_fdcan.rs
@@ -25,12 +25,11 @@ fn main() -> ! {
 
     // Constrain and Freeze clock
     let rcc = dp.RCC.constrain();
-    let prec = rcc.take_peripherals().unwrap();
-    let _ccdr = rcc.sys_ck(100.mhz()).freeze(vos, &dp.SYSCFG);
+    let ccdr = rcc.sys_ck(100.mhz()).freeze(vos, &dp.SYSCFG);
 
-    enable_fdcan(prec.FDCAN);
+    enable_fdcan(ccdr.peripheral.FDCAN);
 
-    //prec.FDCAN.disable(); // Compile error
+    //ccdr.peripheral.FDCAN.disable(); // Compile error
 
     loop {}
 }
--- a/examples/safe_enable_fdcan.rs
+++ b/examples/safe_enable_fdcan.rs
@@ -10,7 +10,7 @@ use stm32h7xx_hal::rcc::{rec, ResetEnable};
 use stm32h7xx_hal::{pac, prelude::*};
 
 fn enable_fdcan(rec: rec::Fdcan) {
-    rec.enable();
+    rec.enable().kernel_clk_mux(rec::FdcanClkSel::PLL1_Q);
 
     // rec is dropped here, and can never be changed again
 }
--- a/src/rcc/rec.rs
+++ b/src/rcc/rec.rs
@@ -40,7 +40,11 @@ impl Rcc {
 
 macro_rules! peripheral_reset_and_enable_control {
     ($($AXBn:ident, $axb_doc:expr => [
-        $( $p:ident $(kernel$(($Variant:ident))* $ccip:ident $clk_doc:expr)* ),*
+        $(
+            $( #[ $pmeta:meta ] )*
+                $p:ident
+            $( kernel$(($Variant:ident))* $ccip:ident $clk_doc:expr )*
+        ),*
     ];)+) => {
         paste::item! {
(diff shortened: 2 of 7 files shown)


richardeoin:
I got this stuck in my head, so came up with a prototype. The usage looks like:

```rust
#[entry]
fn main() -> ! {
    let dp = pac::Peripherals::take().unwrap();
    ...
    let rcc = dp.RCC.constrain();
    let prec = rcc.take_peripherals().unwrap();
    let _ccdr = rcc.sys_ck(100.mhz()).freeze(vos, &dp.SYSCFG);

    prec.FDCAN.enable();
    ...
}
```

What do you think @ryan-summers ? Perhaps it could be extended to peripherals in the HAL too, but would then need to change the API

```rust
let spi = dp.SPI1.spi(..., 
                      &ccdr, // Immutable reference to core clock state
                      prec.SPI1 // Ownership of reset + enable control
);
```

ryan-summers:
@richardeoin I really like this design and believe that it accomplishes the use cases I was thinking of. One question: Is there a reason we can't move the REC module to be contained within the `Ccdr` or a reason why we wouldn't want to do that? It seems like the `Ccdr` could be a "one stop shop" for all peripheral enable/reset/disable and clocking needs.

richardeoin:
I initially kept it separate because I was experimenting, but I think it could be merged in. The structures just need to be members of something that cortex_m already guarantees to exist once. `dp.RCC` is guaranteed to only exist once, so `rcc` is, so `ccdr` is.

`ccdr` would still have to appear in peripheral instantiation calls twice though

```rust
#[entry]
fn main() -> ! {
    let dp = pac::Peripherals::take().unwrap();
    ...
    let rcc = dp.RCC.constrain();
    let ccdr = rcc.sys_ck(100.mhz()).freeze(vos, &dp.SYSCFG);

    ccdr.FDCAN.enable();

    let _: &mut Ccdr = ccdr; // Compile error, value moved on line above

    // This does work
    let spi = dp.SPI1.spi(..., 
                      &ccdr.clocks, // Immutable reference to core clock state
                      ccdr.SPI1 // Ownership of reset + enable control
    );
}
```

The other thing I'd like to get into this API is the peripheral clock multiplexers ("kernel clocks"). This works beautifully when there is a bijective mapping between peripherals and clock multiplexers, but breaks down when multiple peripherals share a clock multiplexer. We could force the user to split those up manually, but it's a bit verbose.

```rust
#[entry]
fn main() -> ! {
    let dp = pac::Peripherals::take().unwrap();
    ...
    let rcc = dp.RCC.constrain();
    let ccdr = rcc.sys_ck(100.mhz()).freeze(vos, &dp.SYSCFG);

    // User still has to call split even if they don't use `clock_mux`
    let (rec_spi4, _) = ccdr.SPI45.clock_mux(PLL3_Q).split();

    let spi = dp.SPI4.spi(..., 
                      &ccdr.clocks, // Immutable reference to core clock state
                      rec_spi4 // Ownership of reset + enable control
    );
}
```

richardeoin:
Ok, this is now ready for review! @ryan-summers @hargoniX perhaps

The next step, for another PR I think, would be to remove the [accessors used internally by the crate](https://github.com/stm32-rs/stm32h7xx-hal/blob/d2112a1b4a3527673878949d90a1711f9f59b56a/src/rcc/mod.rs#L163-L180) and re-work all the peripheral constructors to take `... &ccdr.clocks, ccdr.peripheral.SPI1)`

examples/safe_enable_fdcan.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    // Constrain and Freeze power
+    let pwr = dp.PWR.constrain();
+    let vos = pwr.freeze();
+
+    // Constrain and Freeze clock
+    let rcc = dp.RCC.constrain();
+    let ccdr = rcc.sys_ck(100.mhz()).freeze(vos, &dp.SYSCFG);
+
+    enable_fdcan(ccdr.peripheral.FDCAN);
```
hargoniX, line 30: This example doesn't seem exactly useful to me like....it shows absolutely nothing to the user unless they read the correct bits from the proper registers, can we maybe make this an actual example?

Or is it more meant like an example on how to handle peripherals with the new approach?

src/rcc/rec.rs
```diff
@@ -0,0 +1,256 @@
...
+                    $(
+                        #[allow(missing_docs)]
+                        $( #[ $pmeta ] )*
+                        pub [< $p:upper >]: $p,
+                    )*
+                )+
+
+                // Private field making `Peripherals` non-exhaustive. We
+                // don't use `#[non_exhaustive]` so we can support older
+                // Rust versions.
```
hargoniX, line 84: Maybe add a TODO here so we can fix this once `#[non_exhaustive]`has been around for a while.

src/rcc/rec.rs
```diff
@@ -0,0 +1,256 @@
...
+    AHB1, "AMBA High-performance Bus (AHB1) peripherals" => [
+        Eth1Mac, Dma2, Dma1,
+        #[cfg(any(feature = "dualcore"))] Art
+    ];
+    AHB2, "AMBA High-performance Bus (AHB2) peripherals" => [
+        Sdmmc2, Hash, Crypt
+    ];
+    AHB3, "AMBA High-performance Bus (AHB3) peripherals" => [
+        Sdmmc1,
+        Qspi kernel d1ccip "QUADSPI kernel clock source selection",
```
ryan-summers, line 230: The syntax for the clock allowances on a peripheral don't make it clear, perhaps we could make this something like `Qspi: [kernel, d1ccip], description: "QUADSPI kernel clock source selection"` or something similar when the clocks are allowed to be configured? This may be restricted by the macro structure, but just thought I'd throw this in.

src/rcc/rec.rs
hargoniX: Oh and we should specify how long "a while" should take. So we dont just forget about it.

examples/safe_enable_fdcan.rs
richardeoin: It is a demonstration, indeed it doesn't actually do anything. It can be removed if it doesn't add anything to the example folder

src/rcc/rec.rs
richardeoin: We only need to bump the MSRV by one..

src/rcc/rec.rs
richardeoin: It is quite a serious macro unfortunately. The macro will support more punctuation, but I'm lazy and don't want lines to wrap (80 characters). Perhaps adding a colon after kernel makes sense?

richardeoin:
Great! 

There is one problem - this is unsound on the dual core parts. Marking the accessors with the `Send` trait means that they can be moved between cores, and hence both cores could make a Read-Modify-Write on the same register at the same time. The result of that is undefined. (Actually the ENR registers are separate for each core, but the resets and clock muxes are common)

I can only think of solving this with the `HSEM` peripheral - perhaps worth a try?

richardeoin:
Since we don't support the 2nd core yet, this can be merged already with a note about the potential unsoundness

bors r+

Issue #64, opened 2020-04-18
Support custom SPI configurations


ryan-summers:
The existing SPI configuration assumes a lot of default configuration values (e.g. 8 bit frames, 1 frame per transaction, no delay between CS and SCK assertions). 

I propose that we update the Spi constructors to accept a `spi::Config` structure instead of the `spi::Mode` parameter. This `Config` structure would contain all of the device-specific implementation. To construct the `Config`, a builder semantic can be used:

```
Let config = Spi::Config::new(Mode {polarity: Polarity::IdleHigh, phase: Phase::CaptureOnSecondTransition})
    .frame_size(8) // Configure 8-bit transactions
    .swap_miso_mosi() // Swap MOSI and MISO lines
    .cs_delay(100e-6) // 100uS delay from CS -> SCK
    .freeze();

let spi = spi::Spi1::new(dp.SPI1, (sck, miso, mosi), config, 10.mhz(), &clocks);
```

This allows us to initially provide a minimal configuration as to what users require and expand on it as needed without changing the API at all (e.g. backwards compatible).

I'd be happy to implement this and PR it, just wanted to get some feedback first.

### Background
I'm currently working on a [platform](https://github.com/quartiq/stabilizer) using SPI to talk to some ADCs and DACs where the channel is half-duplex. Some additional configurations that I need are:
1. MSSI support (programmable delay from CS assertion to transaction)
2. Configurable transaction length (16 bit frames)
3. Support for swapping MOSI/MISO lines

richardeoin:
Hello! This sounds very useful. I already have PR https://github.com/stm32-rs/stm32h7xx-hal/pull/61 for 16-bit frames that will be merged before v0.5, but would be keen to add even more functionality. That PR works well with the DAC124S085 for me, the PR also adds the documentation showing half-duplex usage.
```
let spi = dp.SPI1.spi((sck, spi::NoMiso, mosi), spi::MODE_0, 1.mhz(), &mut ccdr);
```


The general aim for this crate is to stick closely to the embedded-hal traits, to reduce the cognitive load for people switching between HALs (and allow device drivers to be interoperable). The [`FullDuplex`](https://docs.rs/embedded-hal/0.2.3/embedded_hal/spi/trait.FullDuplex.html) trait says that CS lines are out of scope, so I would not propose adding MSSI here. For the same reason, we should keep using the spi::MODE_n constants for configuration.

[...]

PR #68, opened 2020-04-19
Allow arbitrary SPI configurations

ryan-summers:
This PR adds a new `spi::Config` structure that can be used to provide arbitrary device-specific configurations for the SPI peripheral as discussed in #64. 

Because of the ability to configure the frame size, I have added a template parameter to the `read()` and `send()` API for `FullDuplex` implementation, but this is not necessarily the best approach (ref #61).

The existing API is unchanged due to the implementation of `From<spi::Mode>` for the new `spi::Config` structure, so applications will not need to be updated.

 @richardeoin Let me know what you think of this implementation - it's only preliminary at this point, but is the general idea I had in #64


src/spi.rs
```diff
@@ -266,29 +350,40 @@ macro_rules! spi {
...
 
                     // ssi: select slave = master mode
                     spi.cr1.write(|w| w.ssi().slave_not_selected());
 
+                    // Calculate the CS->transaction cycle delay bits.
+                    let mut cycle_delay = (config.cs_delay * spi_freq as f32) as u32;
+
+                    // The calculated cycle delay may not be more than 4 bits wide for the
+                    // configuration register.
+                    if cycle_delay > 0xF {
```
ryan-summers, line 364: I'm not a massive fan of this check, but I don't see a way to get at the maximum value in the MSSI register from the `svd2rust` PAC.

richardeoin: Hardware specific constants are completely fine here. We could even panic if the cycle delay isn't possible

src/spi.rs
```diff
@@ -266,29 +350,40 @@ macro_rules! spi {
...
                             .dsize()
-                            .bits(8 - 1) // 8 bit frames
+                            .bits(config.frame_size - 1)
                     });
 
                     // ssi: select slave = master mode
                     spi.cr1.write(|w| w.ssi().slave_not_selected());
 
+                    // Calculate the CS->transaction cycle delay bits.
+                    let mut cycle_delay = (config.cs_delay * spi_freq as f32) as u32;
```
richardeoin, line 360: This could be a good place to use [`ceil`](https://doc.rust-lang.org/std/primitive.f32.html#method.ceil). So that we promise a delay that equal or longer than specified, but never shorter.

src/spi.rs
```diff
@@ -55,6 +56,85 @@ where
...
+            frame_size: 8_u8,
+        }
+    }
+
+    /// Specify that the SPI MISO/MOSI lines are swapped.
+    ///
+    /// Note:
+    /// * This function updates the HAL peripheral to treat the pin provided in the MISO parameter
+    /// as the MOSI pin and the pin provided in the MOSI parameter as the MISO pin.
+    pub fn swap_mosi_miso(&mut self) -> &mut Self {
```
richardeoin, line 98: I think you can just say `pub fn swap_mosi_miso(self) -> Self {...}`. That removes the need for a `freeze()` method

richardeoin:
What do you want to do with #61 ? I'd like the docs changes to go in at least, perhaps it could be merged now so you can build on it?

ryan-summers:
@richardeoin I'm still trying to figure out if we want the frame size / FIFO read size to be constant for an implementation. It seems to me like the template over just `read()` and `send()` allows for more flexible FIFO access, but I don't know if that's necessarily better. I definitely think the docs from #61 should be merged, so let's handle #61 first and figure out the details later.

richardeoin:
The template over read() and send() is an embedded_hal thing, it allows a portable device driver to write 
```rust
impl MySemiSpiDAC{
    pub fn new(spi: impl embedded_hal::spi::FullDuplex<u16>) -> Self { ... }
}
```
and know its been given an SPI peripheral that deals in 16 bit words.

That's a good argument that a spi peripheral that has been `Config`ured for a word length != 8 or 16 should _not_ implement `FullDuplex`.

richardeoin:
As for which frame size to use, perhaps all SPI instances with configurable word length should be u16? I haven't read this part of the datasheet so not sure what the register write size should be.

Ok, I'll merge #61 with the proviso that it's expected to change again with this PR

ryan-summers:
Just as a followup: the SPI module on this family of chips is configurable from 4-32 bits per data frame, so it technically supports a configurable word length up to u32

src/spi.rs
ryan-summers: `ceil` appears to be part of `std::f32` as opposed to `core::f32`, so I don't think we can pull it in

src/spi.rs
ryan-summers: Should be fixed now to not use `freeze()`

ryan-summers:
@richardeoin The current implementation allows a custom frame size from 4-32 bits, and it's not entirely clear to me how that will play with the `embedded-hal` SPI traits, as they (appear to) assume that SPIs operate in byte-sized words.

I think the existing `embedded-hal` default implementations may choke when using e.g. 16-bit words with an 8-bit `FullDuplex` implementation (return invalid data?), but I can't think of a good way around that using a run-time word size, unless we implement default implements for `u8`, u16`, and `u32`

src/spi.rs
jordens: I would also advocate to `ceil`/`floor` all the cases where one typically supplies a datasheet min/max value and can reasonably expect the hal to honor that limit.
Without math in `core`, and the numeric cast being truncation we could still implement a dirty `ceil` here.

ryan-summers: Agreed - I don't think we should silently ignore the truncation (although forgot to indicate this), since the intended use case here is to adhere to some minimum hold time.

ryan-summers: A simple (dirty) workaround I have implemented is to add 1 if the cs_delay is specified as non-zero, which will ensure that we always have at least the number of clock cycles necessary (up to 15).

richardeoin:
The embedded-hal SPI trait should be clearer on this issue imo. The phrase is 
> Some SPIs can work with 8-bit and 16-bit words

I think the community has taken that to mean
* You can expect something that implements `FullDuplex<u8>` to make 8-bit transactions
* You can expect something that implements ` to make 16-bit transactions

Anything else is not really defined.

One way out of this is to have two spi types, which could be called `Spi<SPI, WORD>` and `SpiWithCs<SPI>`. Then only implement those

```
impl FullDuplex<u8> for Spi<$SPIX, u8> { .. } 
impl FullDuplex<u16> for Spi<$SPIX, u16> { .. } 
```
and nothing else

ryan-summers:
After thinking about this, I've come to the conclusion that it's easiest to use the existing word size implementation (and keep it out of the `Config` structure). This prevents us from having to duplicate code across multiple structures. Additionally, it's really corner-case behavior (and functionality that I don't need) to use something with a non-byte-aligned word.

I think we can safely kick the can down the road on that front. I would imagine a chip using a non byte-aligned transfer size would support byte-aligned transfers (and set the bits to dont-cares). If there's a use case in the future, we can address it then.

ryan-summers:
@richardeoin This should be ready for review now. There's not a whole lot in the configuration struct right now, but it provides us an API that can be expanded in the future without breaking the SPI constructor API. This allows us to release additional functionality to SPI without breaking existing code.

richardeoin:
Sorry I took a while to look at this! All looks good to me, and I finally got round to trying it with a TI DAC124S085 - it works great.

bors r+

PR #69, opened 2020-04-20
DAC implementation

richardeoin:
Implementation of a trait that is not in embedded-hal but is common to several of the
other stm32-rs crates

### DIFF
--- a/examples/dac.rs
+++ b/examples/dac.rs
@@ -1,0 +1,68 @@
+#![deny(warnings)]
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+extern crate panic_itm;
+
+use cortex_m;
+use cortex_m::asm;
+use cortex_m_rt::entry;
+use stm32h7xx_hal::hal::Direction;
+use stm32h7xx_hal::{pac, prelude::*};
+
+use stm32h7xx_hal::dac::DacOut;
+use stm32h7xx_hal::traits::DacPin;
+
+use cortex_m_log::println;
+use cortex_m_log::{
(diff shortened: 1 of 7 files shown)


richardeoin:
This still needs more testing or review

astraw:
This works for me on h743 rev v (nucleo board). I tested both channels. Thanks.

One feature that might be useful but not covered here is simultaneous updating to both DACs. Another is running the DAC without the output buffer. Perhaps these are out-of-domain for the HAL.

richardeoin:
Thanks for testing, that is very useful!

The `set_value` call should optimise down to a single register write for each channel, so at the instruction level sequential calls to `set_value` should be very similar to a single function call. But if there was a neat way to write simultaneous updates it could go in.

Selecting the output buffer would be a good addition though.

richardeoin:
I changed the API to add typestates here, as used in the ADC implementation. This removes the simple usage error of calling `calibrate` after `enable`. I also added an `enable_unbuffered` method to support usage without the output buffer.

astraw:
Sounds good.

As far as I read the manual, calibration is only relevant for the buffered operation. Perhaps extending the typestates further, calibration could only be possible in buffered mode?

But I am afraid we move beyond the A (abstraction) in HAL here - how many other devices are going to have similar behavior? On the other hand, this is also a crate for ease-of-use of the h7xx series.

richardeoin:
That's true, but also calibration doesn't break unbuffered mode. So I see no reason to restrict it. 

Perhaps re-naming `calibrate` to `calibrate_buffer` makes the hardware effect obvious?

astraw:
I like it.

PR #70, opened 2020-04-21
SMPS: Add documentation describing Run* mode, and in examples folder

richardeoin:
And a bugfix: In the case of the default mode, we should not modify CR3 at
all. Therefore the default mode should now work for dualcore parts with no
SMPS (if these even exist, I haven't checked).

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,26 @@
+Examples
+======
+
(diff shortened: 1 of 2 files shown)


richardeoin:
bors retry

PR #74, opened 2020-05-04
Erase pins from device driver types

### LINKED ISSUE #65: [RFC] Erase pin types from HAL driver types
When implementing an application using [RTFM](https://github.com/rtfm-rs/cortex-m-rtfm) and storing peripheral resources in the `Resource` structure, the type of the peripheral can be quite onerous to carry around as it contains generic parameters for the pins and peripheral instance.

Example:
```
let spi = hal::spi::Spi1::new(dp.SP2, (sck, miso, mosi), ...);
```

The `spi` variable has the following type:
```
hal::spi::Spi<
    hal::stm32::SPI2, (
        hal::gpio::gpiob::PB10<hal::gpio::Alternate<hal::gpio::AF5>>, 
        hal::gpio::gpiob::PB14<hal::gpio::Alternate<hal::gpio::AF5>>,
        hal::spi::NoMosi)
>
```

This type then changes based on which pins are used. This type is also required to be specified in the RTFM resource structure declaration, which results in a (fairly messy) structure definition that's non-intuitive.

Because the pin types are used only as a compile-time guarantee that they implement the necessary SPI alternate functions, I don't believe there's a need for the SPI module to carry around this type information after instantiation (although the SPI peripheral needs to maintain ownership of the pins). 

I propose that we update the SPI constructor to downgrade the pins to a generic "pin" type since they are never directly controlled by the SPI peripheral. This would remove pin types from the type signature and simplify it to something like `hal::spi::Spi<hal::stm32::SPI2>`, which is much more manageable (and we could potentially remove the SPI2 type as well with other work, although that's beyond scope here).

ryan-summers:
This PR fixes #65 by removing pin storage within device drivers, which also removes the pin type from the driver type. Additionally, it exposes an `unchecked` API that allows users to instantiate a driver without providing pins if they have complex requirements.

This addresses the following drivers:
* I2C
* SPI
* QEI
* UART

CC: @hargoniX @richardeoin

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -46,15 +46,18 @@ where
 }
 
 #[derive(Debug)]
-pub struct I2c<I2C, PINS> {
+pub struct I2c<I2C> {
     i2c: I2C,
-    pins: PINS,
 }
 
 pub trait I2cExt<I2C>: Sized {
-    fn i2c<PINS, F>(self, pins: PINS, freq: F, ccdr: &Ccdr) -> I2c<I2C, PINS>
+    fn i2c<PINS, F>(self, _pins: PINS, freq: F, ccdr: &Ccdr) -> I2c<I2C>
     where
         PINS: Pins<I2C>,
+        F: Into<Hertz>;
+
+    fn i2c_unchecked<F>(self, freq: F, ccdr: &Ccdr) -> I2c<I2C>
+    where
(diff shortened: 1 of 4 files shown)


richardeoin:
This covers all the implementations there were storing the pin types in a struct member. 

Closes #65

bors r+

Issue #35, opened 2019-08-14
[Bug] ADC initialization might panic


Jan561:
### Description and Reproduction
Currently, the ADC initialization might panic if following happens:
- Watchdog enabled
- ADC initialized and enabled
- Panic occurs, causing watchdog to restart the software
- ADC initialized again (and thus calibrated) -> fails because:
  - Reset isn't being triggered because the clock of the ADC is still running (see adc.rs, line 283)
  - Therefore, ADC is still enabled
  -> ADC calibration will panic

### Possible fix

Upon initialization, check if ADC is enabled and disable it then.

richardeoin:
Thanks for detailed explanation!

## Possible Fix

Yes, this would work. Even better - toggle the reset line of every peripheral (excluding the CPU) on initialisation. This is good practice imo.

## API flaw

The ultimate reason for this is that when `Adc::adc1(dp.ADC1, ..)` is called, the HAL has to decide if the user meant to reset both ADC1 and ADC2, or neither. The hardware does not allow them to be reset separately. 

It currently tries to guess based on the enabled state, but in this use case it gets your intentions wrong.

The only option is to change the API to make the intention unambiguous. You could remove `Adc::adc1()` and `Adc::adc2()` and replace them with a `Adc::dual(dp.ADC1, dp.ADC2, ..)` function.

Jan561:
Yeah, good points. Currently I'm working on DMA, but when I finished that I can head over to the ADC.

richardeoin:
Closed by #94

PR #75, opened 2020-05-11
implement free() for ADC

astraw:
The PR adds a `.free()` method to the ADCs in a style similar to many of the other peripherals.

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -587,6 +587,12 @@ macro_rules! adc_hal {
             }
 
             impl<ED> Adc<$ADC, ED> {
+
+                /// Releases the ADC peripheral
+                pub fn free(self) -> $ADC {
+                    self.rb
+                }
+
                 /// Save current ADC config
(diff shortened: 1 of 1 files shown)


richardeoin:
LGTM! It does add a little more complexity around fixing #35 (hardware requires that adc1 and 2 are reset together), but happy to include anyhow.

bors r+

astraw:
I'm not sure if this fully addresses your note, but given that ADC1 and 2 are linked, we could make the signature consume both adc1 and adc2 from the HAL like the following (or similar):

```
/// For ADC1
fn free(self, other: adc2) -> (ADC1, ADC2) {...}

/// For ADC2
fn free(self, other: adc1) -> (ADC1, ADC2) {...}
```

richardeoin:
#78 makes it impossible (in safe rust) to initialise both ADC1 and ADC2, because they will both consume the same singleton `rec::Adc12` marker type. I think once #78 is in, then the ADC1 vs 2 problem can be considered and fixed as a whole.

PR #76, opened 2020-05-14
Add quadspi

jordens:
@ryan-summers I went ahead and drafted this PR to be able to discuss it.


src/qspi.rs
```diff
@@ -0,0 +1,230 @@
...
+        let divisor = match spi_kernel_ck / spi_frequency {
+            1..=255 => {
+                spi_kernel_ck / spi_frequency - 1
+            },
+            _ => {
+                return Err(QspiError::InvalidFrequency);
+            },
+        };
+
+        // Write the prescaler and select flash bank 2 - flash bank 1 is currently unsupported.
```
jordens, line 83: Note to self: only because we use bank 2 on Stabilizer.

src/qspi.rs
```diff
@@ -0,0 +1,230 @@
...
+        regs.cr.write(|w| unsafe {
+            w.prescaler().bits(divisor as u8)
+             .fsel().set_bit()
+        });
+
```
jordens, line 88: I think you want to `cr.en()` the peripheral here and leave it enabled. Don't toggle it during reads/writes: the pertinent registers can be changed with the peripheral enabled and transactions are triggered not by enabling but by writes to the various data/instruction/address registers depeding on the mode.

src/qspi.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+    }
+
+    pub fn write(&mut self, addr: u8, data: &[u8]) -> Result<(), QspiError> {
+        if self.is_busy() {
+            return Err(QspiError::Busy);
+        }
+
+        // Clear the transfer complete flag.
+        self.rb.fcr.modify(|_ ,w| w.ctcf().set_bit());
```
jordens, line 137: Why do you `ctcf()` before and after each read and write? Should be only after obtaining and acknowledging a set `tcf()`, i.e. after each read/write.

src/qspi.rs
```diff
@@ -0,0 +1,230 @@
...
+        match ccdr.rb.d1ccipr.read().qspisel().variant() {
+            stm32::rcc::d1ccipr::QSPISEL_A::RCC_HCLK3 => Some(ccdr.clocks.hclk()),
+            stm32::rcc::d1ccipr::QSPISEL_A::PLL1_Q => ccdr.clocks.pll1_q_ck(),
+            stm32::rcc::d1ccipr::QSPISEL_A::PLL2_R => ccdr.clocks.pll2_r_ck(),
+            stm32::rcc::d1ccipr::QSPISEL_A::PER => ccdr.clocks.per_ck(),
+        }
+    }
+
+    pub fn configure_mode(&mut self, mode: QspiMode) -> Result<(), QspiError> {
+        if self.rb.sr.read().busy().bit_is_set() {
```
jordens, line 106: `self.busy()`

src/qspi.rs
ryan-summers: This is done just to ensure that the transfer complete flag is clear before starting the transaction at all. This prevents a latent, uncleared flag (e.g. caused by a user using the QSPI peripheral before instantiation) from causing a transfer to immediately be detected as complete. This is just ensuring that we have the QSPI peripheral in a "known-good" state and is redundant in 99% of cases, but it doesn't cause any harm.

src/qspi.rs
ryan-summers: After re-reading the documentation, I think you're right here. I'll remove the enables/disables outside of the constructor.

ryan-summers:
@richardeoin I've updated this to feature-gate the QSPI, but for some reason the QSPI example is failing to build and I can't figure out why. I'll try to take a look at this later.

src/qspi.rs
```diff
@@ -0,0 +1,548 @@
...
+            PF7<Alternate<AF9>>,
+            NoIo
+        ]
+        IO3: [
+            PA1<Alternate<AF9>>,
+            PD13<Alternate<AF9>>,
+            PF6<Alternate<AF9>>,
+            NoIo
+        ]
+    ],
```
richardeoin, line 188: This is a rather tedious task.. But the XML files in [modm-devices](https://github.com/modm-io/modm-devices/tree/develop/devices/stm32) are great for cross-checking.

Some corrections

Bank 1
``` 
        IO0: [
            PC9<Alternate<AF9>>,
            PD11<Alternate<AF9>>,
            PF8<Alternate<AF10>>
        ]
        IO1: [
            PC10<Alternate<AF9>>,
            PD12<Alternate<AF9>>,
            PF9<Alternate<AF10>>
        ]
```

Bank 2
```
        IO0: [
            PE7<Alternate<AF10>>,
            PH2<Alternate<AF9>>
        ]
        IO1: [
            PE8<Alternate<AF10>>,
            PH3<Alternate<AF9>>
        ]
```

src/qspi.rs
```diff
@@ -0,0 +1,548 @@
...
+                .adsize()
+                .bits(0)
+                .imode()
+                .bits(0)
+                .dcyc()
+                .bits(0)
+        });
+
+        let spi_frequency = frequency.into().0;
+        let divisor = match spi_kernel_ck / spi_frequency {
```
richardeoin, line 334: Elsewhere we round up the divisor, so that the resulting clock is less than or equal to specified. I think we need to do that here too.

So it would be `match (spi_kernel_ck + spi_frequency - 1) / spi_frequency`

You can also use [pattern binding](https://doc.rust-lang.org/stable/reference/patterns.html#identifier-patterns) to avoid repeating the calculation on the line below

richardeoin:
Continuing the discussion from the #rust-embedded matrix room, from this PR we're going to include more peripherals in the HAL, even if they don't have embedded-hal traits. However, they will be behind feature gates.

Do you have any thoughts @astraw @hargoniX or anyone else? It's a change of direction vs. what it currently says in the [README](https://github.com/stm32-rs/stm32h7xx-hal#other-stm32h7-crates)

astraw:
The comments here are purely to the question of merging things not in embedded-hal into this crate. I haven't reviewed the code in this PR.

If this would be considered as an incubator for eventual potential inclusion to embedded-hal, I would be positive on merging behind a feature gate. If we cannot honestly see eventual inclusion, then I'd still be positive, albeit slightly less so. This is because I don't think every such peripheral should spawn its own new crate - we increase discoverability and uniformity if we include things here, and I think that is overall good.

hargoniX:
Nope, no further thoughts. I think it is appropriate to include the bonus peripherals into the HAL behind feature gates in order to avoid build time issues.

Would also be willing to file some PRs in cass you dont want to merge all your crates on your own @richardeoin

jordens:
> from this PR we're going to include more peripherals in the HAL,
>  It's a change of direction vs. what it currently says in the README

RCC, PWR, DAC, EXTI, SAI, FLASH, Signature all already do that though, right? Would they also end up behind the feature gate? For RCC especially that doesn't seem to be what we want. And there it's not just core functionality but a pretty elaborate implementation.
Putting the non-e-h parts behind a feature gate is fine with me if it's really needed (for semver?) but I think in practice this would be signaled by the HAL user deciding between using an e-h trait vs the HAL's extended trait.

ryan-summers:
> RCC, PWR, DAC, EXTI, SAI, FLASH, Signature all already do that though, right? Would they also end up behind the feature gate? For RCC especially that doesn't seem to be what we want. And there it's not just core functionality but a pretty elaborate implementation.

Based on my understanding, they exist because they are required by something that _is_ implemented in the e-h (e.g. signature may be used for ADC calibration constants etc., RCC is used for clocking SPI/I2C/etc)

richardeoin:
@astraw Indeed, including peripherals is the least-worst option compared to spawning new crates, because of maintenance and discoverability. The user can still write their own implementation to use instead of the HAL one, whilst making use of the rest of the HAL.

@jordens / @ryan-summers It's been a rule with plenty of excuses, either because "core functionality" as you say or because "the other stm32-rs crates have it".

jordens:
To fix that, we should improve the rule to something like "This crate only includes implementations that are compatible with `embedded-hal`" where compatibility means superset and no contradiction.

src/qspi.rs
ryan-summers: Great catch. I was using the STM32H74x series datasheet for this and forgot to go back and look at the 75x series. Fix incoming.

src/qspi.rs
ryan-summers: Ah, I was unaware about pattern binding! That's a nice feature.

ryan-summers:
@richardeoin All review comments should be addressed (I can't mark them as resolved because I didn't originally spawn this PR).

astraw:
@ryan-summers I just gave you admin rights on the repo. Perhaps you can mark comments as resolved now, but in any case, welcome aboard!

src/qspi.rs
```diff
@@ -0,0 +1,548 @@
...
+//! simplest case, this can be accomplished with just the peripheral and the GPIO pins.
+//!
+//! ```
+//! use stm32h7xx_hal::qspi;
+//!
+//! // Get the device peripherals and instantiate IO pins.
+//! let dp = ...;
+//! let (sck, io0, io1, io2, io3) = ...;
+//!
+//! let mut qspi - dp.QUADSPI.bank1(&ccdr.clocks, (sck, io0, io1, io2, io3), 3.mhz(),
```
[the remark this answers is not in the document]
ryan-summers: Fixed in https://github.com/stm32-rs/stm32h7xx-hal/pull/76/commits/172a69b2eed35031ac014e7760bbc69c4eead8c6#diff-7a546c5c633c8b017f7bb8fa952164fbL20

src/qspi.rs
ryan-summers: Fixed in 172a69b2eed35031ac014e7760bbc69c4eead8c6#diff-7a546c5c633c8b017f7bb8fa952164fb

src/qspi.rs
ryan-summers: Fixed in 172a69b2eed35031ac014e7760bbc69c4eead8c6

richardeoin:
@ryan-summers Thanks!

Nightly should be optional in CI, but it looks like we need an [extra flag to allow that](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error).

src/qspi.rs
richardeoin: I think it does still need the `- 1` in the match arm, or in the bitfield assignment below. The register mapping is:

reg = 0; clk / 1
reg = 1; clk / 2
and so on

The bound can be `1..=256`

ryan-summers: Fixed the off-by-one and bounds in 93b7d3de890b31beff5885711376178d50a7a747

src/qspi.rs
```diff
@@ -0,0 +1,552 @@
...
+        self.rb.ccr.modify(|_, w| unsafe {
+            w.fmode().bits(0b01).instruction().bits(addr)
+        });
+
+        // Write the instruction bits to force the read to start. This has to be done after the
+        // transaction is enabled to indicate to the peripheral that we are ready to start the
+        // transaction, even though these bits should already be set.
+        self.rb
+            .ccr
+            .modify(|_, w| unsafe { w.instruction().bits(addr) });
```
ryan-summers, line 481: This should be the address bits.

src/qspi.rs
```diff
@@ -0,0 +1,552 @@
...
+        prec: rec::Qspi,
+    ) -> Self
+    where
+        T: Into<Hertz>,
+        PINS: PinsBank2,
+    {
+        Self::qspi_unchecked(regs, clocks, frequency, prec, Bank::Two)
+    }
+
+    pub fn qspi_unchecked<T>(
```
ryan-summers, line 284: Align arg order with other peripherals.

ryan-summers:
@richardeoin I just noticed a bug in the read function and want to test the read fix on a board at home. I'll comment back later today after I test the update.

ryan-summers:
I have tested the read/write example at both debug and release configurations. I found a bug with the release configuration, as it looks like the peripheral wasn't de-asserting the busy signal quick enough, which resulted in erroneous error indications. I updated the busy check to a while loop to account for this.

Here's a logic analyzer capture of the QSPI read/write example in `Release` configuration. The write cycle can be seen to write 0xAA, 0x00, 0xFF. During the read, as there is no slave connected, all signals remain high.
image

examples/qspi.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    // Constrain and Freeze power
+    let pwr = dp.PWR.constrain();
+    let vos = pwr.freeze();
+
+    // Constrain and Freeze clock
+    let rcc = dp.RCC.constrain();
+    let ccdr = rcc
+        .sys_ck(96.mhz())
+        .pll1_q_ck(48.mhz())
```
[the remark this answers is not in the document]
ryan-summers: This was purely copy-pasted from `examples/spi.rs`, so the custom PLL1 Q clock doesn't actually have any meaning here. I can clear it up.

ryan-summers:
We also discovered a different potential silicon errata with this peripheral where it appears the QSPI peripheral doesn't release it's drive on a read until a half clock cycle too late. Documentation is available at https://github.com/quartiq/stabilizer/issues/101

examples/qspi.rs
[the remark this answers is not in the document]
ryan-summers: Should be fixed in 1218ec6b8a9b88baa0cf56e18a6e4a7a02e3f8f5

PR #79, opened 2020-05-24
DFSDM

richardeoin:
Beginning of DFSDM support. Needs work in upstream PAC crate first, register addresses from SVD are incorrect.

### DIFF
--- a/examples/dfsdm.rs
+++ b/examples/dfsdm.rs
@@ -1,0 +1,39 @@
+// #![deny(warnings)]
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+extern crate panic_itm;
+
+use cortex_m;
+use cortex_m_rt::entry;
+use stm32h7xx_hal::{pac, prelude::*};
+
+#[entry]
+fn main() -> ! {
+    let dp = pac::Peripherals::take().unwrap();
+
+    // Constrain and Freeze power
+    let pwr = dp.PWR.constrain();
+    let vos = pwr.freeze();
+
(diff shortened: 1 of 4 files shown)


richardeoin:

David-OConnor:
Hi. I just implemented DFSDM, and am getting ready to test it on H7 (https://github.com/David-OConnor/stm32-hal/blob/main/src/dfsdm.rs)

Which PAC fields are incorrect? Are they still incorrect?

mattico:
@David-OConnor neat, I'm glad someone is working on a cross-stm32 HAL (besides [embassy](https://github.com/embassy-rs/embassy)). It seemed silly to duplicate all this work when the STM32 chips are similar enough that a lot can be reused.

Looking at the [register offsets](https://docs.rs/stm32h7/0.13.0/stm32h7/stm32h743v/dfsdm/struct.RegisterBlock.html) and §30.8.16 of RM0433 R7, all of the field offsets except for DFSDM_CHCFG0R1 are wrong. Looks like all of the names are wrong, too and some of the fields are missing? Looking at the stm32-rs nightlies this wasn't fixed by the updated SVDs.

image

The registers are grouped by channel. All of the different types of registers for one channel are next to each other. However in the PAC the registers are grouped by type. e.g. all of the DFSDM_CHCFGxR1 registers are one after another in memory.

David-OConnor:
Fix inbound https://github.com/stm32-rs/stm32-rs/pull/637

Offsets only appear to be off for H743 and H743V. Fixing reg names across H7.

PR #81, opened 2020-05-31
Implementation for LCD-TFT display controller

richardeoin:
Somewhat replicates the implementation from the [F7 HAL](https://github.com/stm32-rs/stm32f7xx-hal/blob/master/src/ltdc.rs), but with a re-usable traits that should apply to a wide range of displays and at least stm32 microcontrollers.


richardeoin:
This would be a good candidate to be shared between f4/f7/h7, as well as the FMC

richardeoin:

PR #82, opened 2020-06-06
Note about programming Arduino H7 board

richardeoin:
This looks like a nice H7 board (if a bit expensive?), but a quick scan of the [schematic](https://content.arduino.cc/assets/Arduino-PortentaH7-schematic-V1.0.pdf) shows that the usual programming interfaces are only on difficult-to-solder high density connectors. The stm32 bootloader is only on the *other* USB port (USB_FS).

I think a little expectations management is therefore appropriate, especially to warn new users. If someone has a board to test, we can then update with more details.


astraw:
I looked at the docs a little and agree with your analysis. Perhaps the MKR connector offers some possibilities but still the boot pin would need to be toggled, and indeed even this seems available only on the high density connector.

bors r+

PR #83, opened 2020-06-17
Add support for MCO pins, also small improvements to rcc

richardeoin:
MCO = Micro-Controller Out pins, PA8 and PC9

### DIFF
--- a/examples/mco.rs
+++ b/examples/mco.rs
@@ -1,0 +1,71 @@
+#![deny(warnings)]
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+extern crate panic_itm;
+
+use cortex_m;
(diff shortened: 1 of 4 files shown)


hargoniX:
While we're at fixing the cortex-m-rtfm fail from the CI, it might be useful to just switch to the new name cortex-m-rtic as well? (it got renamed)

richardeoin:
I'll make a new PR for the

richardeoin:
RTFM actually broke two things here: 

* The rename breaks the examples, because we have `#![deny(warnings)]` 
* The latest rtfm/rtic breaks our MSRV (1.40.0)

I've switched the CI to github actions so it's not painfully slow. It only tests against the latest stable rust now.

PR #91, opened 2020-06-27
i2c: Fix i2c bus clock frequency calculation

richardeoin:
This is quite a tricky area but the previous version was flawed in several ways:
* The floating-point representation of `presc` was used for further calculation,
whilst the hardware divides by integers only.
* The `assert!` statements for out of range values of out-of-range values were
commented, probably because they were being hit regularly.
* Some almost-magic numbers (514, 387 ?)

The HAL now panics if the bus frequency is too slow to be achieved with the
maximum division ratio (8192). For a max(vos1) PCLK4 = 100MHz this corresponds
to a bus frequency of 12.2kHz. Previously it just set some other semi-random bus
frequency faster than specified.

It now asserts that the ratio between `i2cclk` and `freq` must be 4 or greater;
this is assumed by the standard-mode calculation.

The `SDADEL` and `SCLDEL` calculations still need review. The reference manual
has several complex requirements for these. However, they are now limited to 2
and 4 respectively, matching the configuration examples in RM0433 Rev 7
47.4.10. This also stops them overflowing.

Tested with a few devices at 100kHz and 400kHz bus clocks.

### DIFF
--- a/src/i2c.rs
+++ b/src/i2c.rs
@@ -34,18 +34,30 @@ pub enum Error {
 }
 
 /// A trait to represent the SCL Pin of an I2C Port
-pub trait PinScl<I2C> {}
+pub trait PinScl<I2C> {
+    fn set_open_drain(self) -> Self;
+}
 
 /// A trait to represent the SDL Pin of an I2C Port
-pub trait PinSda<I2C> {}
+pub trait PinSda<I2C> {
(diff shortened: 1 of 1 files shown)


richardeoin:
It was very useful having I2C in the HAL for a some time now! Would be good to review `SDADEL` and `SCLDEL` in future, but this can go in for now

bors r+

PR #92, opened 2020-07-04
Fixes #72: Timer first cycle invalid

### LINKED ISSUE #72: Timer first cycle is invalid
When the timer is configured, the first cycle of the timer will have an incorrect period. After the first cycle, all following periods will be correct.

This occurs because the ARR register is buffered and doesn't become "Active" (referred as the shadow register in the datasheet) until an update event occurs (either by software or hardware).

As such, the timer `start()` routine needs to make a call to 
```rust
self.tim.egr.write(|w| w.ug.set_bit());
```
to force an update of the ARR register. There's another method of accomplishing this (bypassing the shadow register), but I don't think that's what we want.

ryan-summers:
This PR fixes #72 by triggering a timer update event to ensure the ARR register is updated.

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -162,6 +162,10 @@ macro_rules! hal {
                     // Set PSC and ARR
                     self.set_freq(timeout);
 
+                    // Generate an update event to force an update of the ARR register. This ensures
+                    // the first timer cycle is of the specified duration.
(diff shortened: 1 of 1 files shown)


richardeoin:
Looks good to me!

bors r+

Issue #93, opened 2020-07-04
Prec structures preclude using ADC1 and ADC2 simultaneously 


ryan-summers:
The existing Prec implementation precludes using ADC1 and ADC2 simultaneously because only a single Prec structure is generated for both ADC implementations.

ryan-summers:
@richardeoin Perhaps we could update the ADC constructor to just borrow the ADC PREC structure, which would allow it to be used more than once? Curious to hear your thoughts. We could potentially also generate duplicate PREC structures for ADC1 and ADC2, but I feel like that would obfuscate the fact that they're essentially sharing resources.

richardeoin:
I think this is the only case where a single reset line is used to control two peripherals that have separate definitions in the SVD.

We could fix this the in ADC HAL by initialising both ADC1 and ADC2 at the same time. The usage would then look like:

```
Adc::adc12(dp.ADC1, dp.ADC2, ccdr..., ...) -> (adc1, adc2)
```

richardeoin:
So if you were just using ADC1:

```
let (adc1, _) = Adc::adc12(dp.ADC1, dp.ADC2, ccdr..., ...);

ryan-summers:
I like that concept as it makes it very clear to the user that the peripherals are intertwined.

richardeoin:
Indeed, it forces any alternative ADC HAL implementations to acknowledge that ADC1 and ADC2 are linked

ryan-summers:
I'd be happy to write up a PR for this change. It'd be nice to get some of the `stabilizer` code base as back-on-track with the HAL as possible.

richardeoin:
I'm most of the way through making the change, will push a PR shortly. I'll open the branch to collaborators though, if there's anything you want to add.

ryan-summers:
Fixed in https://github.com/stm32-rs/stm32h7xx-hal/pull/94

PR #94, opened 2020-07-04
Add method to initialise ADC1 and ADC2 together

richardeoin:
See #93

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -1,11 +1,13 @@
 //! Analog to Digital Converter (ADC)
+//!
+//! ADC1 and ADC2 share a reset line. To initialise both of them, use the
+//! `adc12` method.
 
 use crate::hal::adc::{Channel, OneShot};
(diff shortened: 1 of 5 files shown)


src/adc.rs
```diff
@@ -317,46 +395,37 @@ macro_rules! adc_hal {
...
                              prec: rec::$Rec, clocks: &CoreClocks
                 ) -> Self {
-                    let mut s = Self {
-                        rb: adc,
-                        sample_time: AdcSampleTime::default(),
-                        resolution: Resolution::SIXTEENBIT,
-                        lshift: AdcLshift::default(),
-                        _enabled: PhantomData,
-                    };
+                    let mut a = Self::default_from_rb(adc);
```
ryan-summers, line 398: Would recommend a rename from `a` -> `adc` (or something similar. Benefit of rebinding the `adc` is that we consume it here as well.)

richardeoin: Agreed, thanks!

PR #96, opened 2020-07-11
Updated pwr to handle vos0

mtthw-meyer:
Looking at how VOS0 scaling needs to be implemented the current structure is going to need to be broken a bit I opted to create a new freeze_vos0 function to not break all the existing code.

> Setup PWR...
> Setup RCC...
> 
> stm32h7xx-hal example - RCC
> 
> hclk = 240 MHz
> sys_ck = 480 MHz

RM0433 rev 7
> VOS0 activation/deactivation sequence
> The system maximum frequency can be reached by boosting the voltage scaling level to
> VOS0. This is done through the ODEN bit in the SYSCFG_PWRCR register.
> The sequence to activate the VOS0 is the following:
> 
> 1. Ensure that the system voltage scaling is set to VOS1 by checking the VOS bits in
> PWR D3 domain control register (PWR D3 domain control register (PWR_D3CR))
> 
> 2. Enable the SYSCFG clock in the RCC by setting the SYSCFGEN bit in the
> RCC_APB4ENR register.
> 
> 3. Enable the ODEN bit in the SYSCFG_PWRCR register.
>   
> 4. Wait for VOSRDY to be set.
> Once the VCORE supply has reached the required level, the system frequency can be
> increased. Figure 31 shows the recommended sequence for switching VCORE from VOS1 to
> VOS0 sequence.

### DIFF
--- a/examples/vos0.rs
+++ b/examples/vos0.rs
@@ -7,7 +7,7 @@ extern crate panic_itm;
 
 use cortex_m;
 use cortex_m_rt::entry;
-use stm32h7xx_hal::{pac, rcc, prelude::*};
+use stm32h7xx_hal::{pac, prelude::*, rcc};
 
 use cortex_m_log::println;
 use cortex_m_log::{
@@ -23,11 +23,13 @@ fn main() -> ! {
     // Constrain and Freeze power
     println!(log, "Setup PWR...                  ");
     let pwr = dp.PWR.constrain();
-    let rcc = dp.RCC.constrain();
-    let vos = pwr.freeze_vos0(&rcc, &dp.SYSCFG);
+    let vos = pwr.vos0(&dp.SYSCFG).freeze();
 
     // Constrain and Freeze clock
+    // The PllConfigStrategy::Normal strategy uses the medium range VCO which has a maximum of 420 Mhz
+    // Switching to PllConfigStrategy::Iterative sets the VCO to wide range to allow this clock to reach 480 Mhz
     println!(log, "Setup RCC...                  ");
+    let rcc = dp.RCC.constrain();
     let ccdr = rcc
         .sys_ck(480.mhz())
         .pll1_strategy(rcc::PllConfigStrategy::Iterative)
@@ -39,7 +41,7 @@ fn main() -> ! {
 
     // HCLK
     println!(log, "hclk = {} MHz", ccdr.clocks.hclk().0 as f32 / 1e6);
-    // assert_eq!(ccdr.clocks.hclk().0, 50_000_000);
+    assert_eq!(ccdr.clocks.hclk().0, 240_000_000);
 
     // SYS_CK
     println!(log, "sys_ck = {} MHz", ccdr.clocks.sys_ck().0 as f32 / 1e6);
--- a/examples/vos0.rs
+++ b/examples/vos0.rs
@@ -36,7 +36,7 @@ fn main() -> ! {
         .freeze(vos, &dp.SYSCFG);
 
     println!(log, "");
-    println!(log, "stm32h7xx-hal example - RCC");
+    println!(log, "stm32h7xx-hal example - VOS0");
     println!(log, "");
 
     // HCLK
--- a/examples/vos0.rs
+++ b/examples/vos0.rs
@@ -1,0 +1,49 @@
+#![deny(warnings)]
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+extern crate panic_itm;
+
+use cortex_m;
+use cortex_m_rt::entry;
+use stm32h7xx_hal::{pac, rcc, prelude::*};
+
+use cortex_m_log::println;
+use cortex_m_log::{
+    destination::Itm, printer::itm::InterruptSync as InterruptSyncItm,
+};
+
+#[entry]
+fn main() -> ! {
+    let cp = cortex_m::Peripherals::take().unwrap();
+    let dp = pac::Peripherals::take().expect("Cannot take peripherals");
+    let mut log = InterruptSyncItm::new(Itm::new(cp.ITM));
+
+    // Constrain and Freeze power
+    println!(log, "Setup PWR...                  ");
+    let pwr = dp.PWR.constrain();
+    let rcc = dp.RCC.constrain();
+    let vos = pwr.freeze_vos0(&rcc, &dp.SYSCFG);
+
+    // Constrain and Freeze clock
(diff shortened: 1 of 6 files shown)


examples/vos0.rs
```diff
@@ -0,0 +1,49 @@
...
+    println!(log, "Setup PWR...                  ");
+    let pwr = dp.PWR.constrain();
+    let rcc = dp.RCC.constrain();
+    let vos = pwr.freeze_vos0(&rcc, &dp.SYSCFG);
+
+    // Constrain and Freeze clock
+    println!(log, "Setup RCC...                  ");
+    let ccdr = rcc
+        .sys_ck(480.mhz())
+        .pll1_strategy(rcc::PllConfigStrategy::Iterative)
```
richardeoin, line 33: This isn't needed or explained here, should be removed

examples/vos0.rs
```diff
@@ -0,0 +1,49 @@
...
+        .freeze(vos, &dp.SYSCFG);
+
+    println!(log, "");
+    println!(log, "stm32h7xx-hal example - RCC");
+    println!(log, "");
+
+    // HCLK
+    println!(log, "hclk = {} MHz", ccdr.clocks.hclk().0 as f32 / 1e6);
+    // assert_eq!(ccdr.clocks.hclk().0, 50_000_000);
```
richardeoin, line 42: Worth setting the new value and uncommenting this line

src/pwr.rs
```diff
@@ -236,4 +238,69 @@ impl Pwr {
...
+        // We have now entered Run mode. See RM0433 Rev 7 Section 6.6.1
+
+        // go to VOS1 voltage scale for high performance
+        self.rb.d3cr.write(|w| unsafe { w.vos().bits(0b11) });
+        while self.rb.d3cr.read().vosrdy().bit_is_clear() {}
+
+        // Enable overdrive for maximum clock
+        // Syscfgen required to set enable overdrive
+        rcc.rb.apb4enr.modify(|_, w| w.syscfgen().enabled());
+        unsafe { syscfg.pwrcr.modify(|_, w| w.oden().bits(1)) };
```
richardeoin, line 301: You probably want `w.oden().set_bit()` here

examples/vos0.rs
mtthw-meyer: The rcc::PllConfigStrategy::Normal strategy uses the medium range VCO which has a maximum of 420 Mhz for PLL1, and hence won't work at the 480Mhz sys_ck (it causes a panic). Iterative uses the wide range VCO which goes much higher.

examples/vos0.rs
mtthw-meyer: Yes, I just forgot.

src/pwr.rs
mtthw-meyer: Doesn't exist for this register for some reason.
```
unsafe { syscfg.pwrcr.modify(|_, w| w.oden().set_bit()) };
    |                                                      ^^^^^^^ method not found in `stm32h7::stm32h743v::syscfg::pwrcr::ODEN_W<'_>`
```

mtthw-meyer:
> As well as the inline comments, I think it would be good to avoid the code duplication if possible. Particularly as this is quite fragile and complex code (see the comments!) You should be able to add another builder method to `Pwr` for setting the vos instead.
> 
I need to get rcc: &Rcc, syscfg: &SYSCFG into Pwr somehow. You want me to add a fn vos0(rcc: &Rcc, syscfg: &SYSCFG) that takes those and have it hold a a reference until Pwr goes out of scope then instead?

examples/vos0.rs
richardeoin: Sorry, in that case it is needed. At some point it would be nice to add another strategy that uses the wide VCO

src/pwr.rs
richardeoin: Ah the problem is it does exist on some parts 
https://docs.rs/stm32h7/0.11.0/stm32h7/stm32h747cm7/syscfg/pwrcr/struct.ODEN_W.html#method.set_bit
But not on others
https://docs.rs/stm32h7/0.11.0/stm32h7/stm32h743/syscfg/pwrcr/struct.ODEN_W.html#method.set_bit

The `bits` method is the opposite, which is why the CI is failing on it.

Ultimately it would be worth extending this PR https://github.com/stm32-rs/stm32-rs/pull/366/files

But for now you can write
```rust
#[cfg(any(feature = "dualcore"))]
syscfg.pwrcr.modify(|_, w| w.oden().set_bit());
#[cfg(not(any(feature = "dualcore")))]
unsafe { syscfg.pwrcr.modify(|_, w| w.oden().bits(1)) };
```

richardeoin:
> I need to get rcc: &Rcc, syscfg: &SYSCFG into Pwr somehow. You want me to add a fn vos0(rcc: &Rcc, syscfg: &SYSCFG) that takes those and have it hold a a reference until Pwr goes out of scope then instead?

We want to write an API that makes it obvious what memory is being accessed, but we don't want to get too caught up in the memory aliasing rules.

Perhaps a good trade-off would be

```rust
impl Pwr {
    pub fn vos0(mut self, _: &SYSCFG) -> Self { self.vos_configuration =..; self }
}
```

Then when you need to, break the aliasing rules
```rust
unsafe { // unsafe: enable bit in SYSCFG
    &(*RCC::ptr()).apb4enr.modify(|_, w| w.syscfgen().enabled());
    &(*SYSCFG::ptr()).pwrcr.modify(|_, w| w.oden().bits(1));
}
```

mtthw-meyer:
Failing CI due to attributes on if expressions. That landed in 1.43 and the CI ran on 1.41.

richardeoin:
Nice, thanks for the changes. Good point about this being `revision_v` too.

You could get around the old stable problems by moving the if statement into a non-`pub` function and gating the call to that? Otherwise you can increase the MSRV to 1.43 in (README.md, bors.toml, ci.yml) if needed

mtthw-meyer:
Could, I think it's better to just go to 1.43, it's such a useful conditional compile flag to have.

richardeoin:
Ah you can specify which feature gates an example is tested for at the bottom of Cargo.toml

```
[[example]]
name = "vos0"
required-features = ["revision_v"]
```

I guess it works to put the `cfg` gate in the example itself, but people tend to copy+paste that into their own code

mtthw-meyer:
Aha, so there is a way to do that.

richardeoin:
Looks good! I think I have to merge manually because of the bors changes..

PR #99, opened 2020-07-17
Sai i2s v0.2.x 1

### LINKED ISSUE #90: No implementation for SAI I2S
As far as I can tell the only implemented functionality for SAI is PDM. The daisy uses SAI I2S for audio I/O to the codec chip. If I can figure it out I may take a shot at implementing it myself.

mtthw-meyer:
Implemented I2S for SAI1-4.
Added traits to split DMA channels.

Example runs on daisy seed board successfully.

Closes #90

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,10 +36,6 @@ sdio-host = { version = "0.4", optional = true }
 stm32-fmc = { version = "0.2", optional = true }
 synopsys-usb-otg = { version = "^0.2.2", features = ["cortex-m"], optional = true }
 
-generic-array = "0.14.2"
-as-slice = "0.1.3"
-stable_deref_trait = { version = "1.1.1", default-features = false }
-
 [dependencies.smoltcp]
 version = "0.6.0"
 default-features = false
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ features = ["stm32h743", "rt", "quadspi", "sdmmc", "fmc", "ethernet", "phy_lan87
 targets = ["thumbv7em-none-eabihf"]
 
 [dependencies]
+embedded-hal = "0.2.4"
 cortex-m = "^0.6.2"
 cortex-m-rt = "^0.6.12"
 stm32h7 = "^0.12.1"
@@ -41,10 +42,6 @@ version = "0.6.0"
 default-features = false
 features = ["ethernet", "proto-ipv4", "proto-dhcpv4", "socket-tcp", "socket-raw"]
 optional = true
-
-[dependencies.embedded-hal]
-git = "https://github.com/eldruin/embedded-hal.git"
-branch = "i2s-v0.2.x"
 
 [dev-dependencies]
 cortex-m-rtic = "0.5.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,6 @@ features = ["stm32h743", "rt", "quadspi", "sdmmc", "fmc", "ethernet", "phy_lan87
 targets = ["thumbv7em-none-eabihf"]
 
 [dependencies]
-embedded-hal = "0.2.4"
 cortex-m = "^0.6.2"
 cortex-m-rt = "^0.6.12"
 stm32h7 = "^0.12.1"
@@ -37,11 +36,19 @@ sdio-host = { version = "0.4", optional = true }
 stm32-fmc = { version = "0.2", optional = true }
 synopsys-usb-otg = { version = "^0.2.2", features = ["cortex-m"], optional = true }
 
+generic-array = "0.14.2"
+as-slice = "0.1.3"
+stable_deref_trait = { version = "1.1.1", default-features = false }
+
 [dependencies.smoltcp]
 version = "0.6.0"
 default-features = false
 features = ["ethernet", "proto-ipv4", "proto-dhcpv4", "socket-tcp", "socket-raw"]
 optional = true
+
+[dependencies.embedded-hal]
+git = "https://github.com/eldruin/embedded-hal.git"
+branch = "i2s-v0.2.x"
 
 [dev-dependencies]
 cortex-m-rtic = "0.5.3"
--- a/src/sai/i2s.rs
+++ b/src/sai/i2s.rs
@@ -5,7 +5,7 @@
 use core::convert::TryInto;
 
 use crate::rcc::{rec, CoreClocks, ResetEnable};
-use crate::sai::{GetClkSAI, Sai, SaiChannel, INTERFACE};
+use crate::sai::{GetClkSAI, Sai, SaiChannel, CLEAR_ALL_FLAGS_BITS, INTERFACE};
 use crate::stm32;
 use crate::stm32::{SAI1, SAI2, SAI3, SAI4};
 use crate::time::Hertz;
@@ -90,16 +90,17 @@ pub enum I2SCompanding {
     MuLaw = 0b11,
 }
 
-pub enum I2SCompliment {
+#[derive(Copy, Clone)]
+pub enum I2SComplement {
     Ones = 0,
     Twos = 1,
 }
 
-impl I2SCompliment {
-    fn as_bool(&self) -> bool {
-        match &self {
-            I2SCompliment::Ones => false,
-            I2SCompliment::Twos => true,
+impl From<I2SComplement> for bool {
+    fn from(value: I2SComplement) -> Self {
+        match value {
+            I2SComplement::Ones => false,
+            I2SComplement::Twos => true,
         }
     }
 }
@@ -151,12 +152,13 @@ pub struct I2SChanConfig {
     oversampling: bool,
     master_clock_disabled: bool,
     companding: I2SCompanding,
-    compliment: I2SCompliment,
+    complement: I2SComplement,
     protocol: I2SProtocol,
     mono_mode: bool,
     mute_repeat: bool,
     mute_counter: u8,
     tristate: bool,
+    frame_size: Option<u8>,
 }
 
 impl I2SChanConfig {
@@ -176,12 +178,13 @@ impl I2SChanConfig {
             oversampling: false,
             master_clock_disabled: false,
             companding: I2SCompanding::Disabled,
-            compliment: I2SCompliment::Ones,
+            complement: I2SComplement::Ones,
             protocol: I2SProtocol::MSB,
             mono_mode: false,
             mute_repeat: false,
             mute_counter: 0,
             tristate: false,
+            frame_size: None,
         }
     }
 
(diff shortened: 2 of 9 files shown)


mtthw-meyer:
I'm done with this if you think it's ready for review.

src/sai/i2s.rs
```diff
@@ -0,0 +1,748 @@
...
+
+    fn try_send(
+        &mut self,
+        left_word: u32,
+        right_word: u32,
+    ) -> nb::Result<(), Self::Error> {
+        let chan = &self.rb.cha;
+        match chan.sr.read().flvl().variant() {
+            Val(sr::FLVL_A::FULL) => Err(nb::Error::WouldBlock),
+            Val(sr::FLVL_A::QUARTER4) => Err(nb::Error::WouldBlock),
```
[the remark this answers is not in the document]
mtthw-meyer: The FIFO is 8 words long. A write consists of 2 words, in stereo mode at least. Therefore you need to wait for 3/4s to ensure 2 words are available for writing. That's how I read the docs anyway. I'll add a comment to explain it.

src/sai/i2s.rs
```diff
@@ -0,0 +1,748 @@
...
+                            slave,
+                            0,
+                            bit_rate,
+                        );
+                    }
+
+                    per_sai
+                }
+
+                pub fn enable(&mut self) {
```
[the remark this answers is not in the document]
mtthw-meyer: Nice, totally missed those helper functions.

Cargo.toml
```diff
@@ -22,19 +22,25 @@ features = ["stm32h743", "rt"]
...
 cortex-m = "^0.6.2"
 cortex-m-rt = "^0.6.12"
 stm32h7 = "0.11.0"
 void = { version = "1.0.2", default-features = false }
 cast = { version = "0.2.3", default-features = false }
 nb = "0.1.2"
 paste = "0.1.18"
+generic-array = "0.14.2"
+as-slice = "0.1.3"
+stable_deref_trait = { version = "1.1.1", default-features = false }
```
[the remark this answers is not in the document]
mtthw-meyer: Leftover from when I was looking into DMA implementation.

src/sai/i2s.rs
```diff
@@ -0,0 +1,748 @@
+//! # Serial Audio Interface - Inter-IC Sound
+//!
+//! Inter-IC Sound.
+//!
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
mtthw-meyer: Removed

richardeoin:
I finally got round to trying this with the WM8994. It actually requires a huge amount of configuration over I2C to get it to do anything, but so far I've been using it as an ADC.

The "I2S" interface from the datasheet:

Screenshot 2020-07-21 at 17 59 03
The main point is that if configured to use the actual "I2S" interface it follows the Phillips I2S specification where the DACDAT/ADCDAT are available one cycle later. That means that `FRCR`->`FSOFF` needs to be set. The first bit offset in `SLOTR` isn't quite what's needed - that masks the bit ahead of the MSB but the LSB is still lost.

Could `FSOFF` be added to `I2SChanConfig` with some sensible name?

Also the WM8994 defaults to having 16-bit audio words but 64-bit frames. It does work with this implementation by changing the AIF1 BCLK in Register 771, then it has 16-bit words and 32-bit frames. But I wonder if it would make more sense for `I2SChanConfig::new` to specify the frame length and just assume the number of slots is 2 for now?

The current code doesn't make much sense for slots != 2. slots = {0,1} would go very badly (frame_length = 0) whilst my understanding is that > 2 slots is only practically useful where some of them are disabled, so other devices on the bus can insert data. We could add support for slots != 2 at some point in the future, but I think those are quite specialised use cases.

mtthw-meyer:
I'll look into those changes. There are ALOT of settings and it's not clear which ones are useful to be accessible.

src/sai/i2s.rs
```diff
@@ -0,0 +1,748 @@
...
+                .bits(slot_size as u8)
+                .nbslot()
+                .bits(config.slots - 1)
+                .sloten()
+                .bits(slot_en_bits)
+        });
+        match mode {
+            I2SMode::Master => audio_ch
+                .im
+                .modify(|_, w| w.ovrudrie().enabled().wckcfgie().enabled()),
```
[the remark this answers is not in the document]
richardeoin: For clarification it isn't needed in this passthrough example, because the `FREQIE` is set on the slave channel. However I think if you were using one (master) channel only, then it is needed if you want to feed/send data from an interrupt.

richardeoin:
Thanks for those updates! Indeed managing all the configuration options is difficult, it's really only by experimenting that you get an idea of which ones are useful.

A nice method of managing all the configuration options is the 'builder pattern', there's a good example in the spi config @ryan-summers [added recently](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/spi.rs#L127). The new function can still take arguments, but things that have obvious defaults / are optional can configured by a builder method instead.

src/sai/i2s.rs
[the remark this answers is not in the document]
mtthw-meyer: All the interrupts should be accessible from the listen/unlisten methods now.

mtthw-meyer:
Builder is in place now with most of the configuration exposed. I cleaned up the listen/unlisten methods to work by channel since they were locked into channel A and exposed some more controls at the sai level.

src/sai/i2s.rs
```diff
@@ -0,0 +1,877 @@
...
+    MuLaw = 0b11,
+}
+
+pub enum I2SCompliment {
+    Ones = 0,
+    Twos = 1,
+}
+
+impl I2SCompliment {
+    fn as_bool(&self) -> bool {
```
richardeoin, line 99: I would have expected to see `#[derive(Copy, Clone)]` on I2SComplement, and then to see this written as `impl From<I2SCompliment> for bool`

Not an essential change, would just make it more idiomatic

src/sai/i2s.rs
```diff
@@ -0,0 +1,877 @@
...
+                .prtcfg()
+                .free()
+                .ds()
+                .bits(bit_rate as u8)
+                .lsbfirst()
+                .bit(if config.protocol == I2SProtocol::LSB {
+                    true
+                } else {
+                    false
+                })
```
richardeoin, line 618: You can just write `.bit(config.protocol == I2SProtocol::LSB)`

src/sai/mod.rs
```diff
@@ -122,34 +142,99 @@ macro_rules! sai_hal {
...
-                    self.rb.cha.clrfr.write(|w| {
-                        // FIFO request interrupt is cleared by
-                        // reading or writing data
-                        w
-                    });
+                /// Clears all interrupts on the `channel`
+                pub fn clear_all_irq(&mut self, channel: SaiChannel) {
+                    match channel {
+                        SaiChannel::ChannelA => &self.rb.cha.clrfr.reset(),
+                        SaiChannel::ChannelB => &self.rb.chb.clrfr.reset(),
```
richardeoin, line 199: This doesn't quite do what you expect. The `reset` method ([from svd2rust](https://docs.rs/svd2rust/0.17.0/svd2rust/)) writes the reset value as defined in the SVD to the register. In this case the reset value is all zeros, which doesn't clear any of the flags.

mtthw-meyer:
I can expose that as well, I'm not sure anyone will need it but might as well.

mtthw-meyer:
Closes #90

eldruin:
Alright, I'll mark this implementation as finished in the issue. Thank you for all your work!

src/sai/mod.rs
```diff
@@ -122,34 +144,101 @@ macro_rules! sai_hal {
...
+                        SaiChannel::ChannelA => &self.rb.cha,
+                        SaiChannel::ChannelB => &self.rb.chb,
+                    };
+                    match event {
+                        Event::Overdue              => ch.im.write(|w| w.ovrudrie().set_bit()),
+                        Event::Muted                => ch.im.write(|w| w.mutedetie().set_bit()),
+                        Event::WrongClock           => ch.im.write(|w| w.wckcfgie().set_bit()),
+                        Event::Data                 => ch.im.write(|w| w.freqie().set_bit()),
+                        Event::AnticipatedFrameSync => ch.im.write(|w| w.afsdetie().set_bit()),
+                        Event::LateFrameSync        => ch.im.write(|w| w.lfsdetie().set_bit()),
```
richardeoin, line 159: I happened to be looking at this again, should these not be `modify` rather than `write`? I would expect multiple calls to `listen()` to set all the events, not just the last one.

src/sai/mod.rs
```diff
@@ -122,34 +144,101 @@ macro_rules! sai_hal {
...
-                        Event::Data => {
-                            // Enable FIFO request interrupt
-                            self.rb.cha.im.modify(|_, w| w.freqie().clear_bit());
-                        }
+                        Event::Overdue              => ch.clrfr.write(|w| w.covrudr().set_bit()),
+                        Event::Muted                => ch.clrfr.write(|w| w.cmutedet().set_bit()),
+                        Event::WrongClock           => ch.clrfr.write(|w| w.cwckcfg().set_bit()),
+                        Event::Data                 => (), // Cleared by reading/writing data
+                        Event::AnticipatedFrameSync => ch.clrfr.write(|w| w.cafsdet().set_bit()),
+                        Event::LateFrameSync        => ch.clrfr.write(|w| w.clfsdet().set_bit()),
```
richardeoin, line 193: These are correctly `write` though

src/sai/mod.rs
mtthw-meyer: What's the difference? Sounds like modify should be for changed vs ones that are just written like flags?

mtthw-meyer: Yes, each flag set requires a call to listen, what do you mean? The example only registers the data flag currently.

richardeoin: Ah, modify specifically refers to "Read-Modify-Write" - [svd2rust explains it well](https://en.wikipedia.org/wiki/Read%E2%80%93modify%E2%80%93write). The outcome for `write` is that all the bits not in the closure are cleared.

richardeoin: To clarify with an example, I would expect

```rust
let mut sai = ...;

sai.listen(Event::Data);
sai.listen(Event::WrongClock);
```

to set both the Data and WrongClock flags.

src/sai/mod.rs
mtthw-meyer: Got it. Did a pass on the modules I touched for this and fixed them.

richardeoin:
I'd like to see this merged for the next [stm32h7xx-hal release](https://github.com/stm32-rs/stm32h7xx-hal/milestone/6). Until embedded-hal catches up, we can do by copying the required traits into the [traits](https://github.com/stm32-rs/stm32h7xx-hal/tree/master/src/traits) directory.

@mtthw-meyer does that sound like a reasonable plan to you?

mtthw-meyer:
Sounds fine to me.

richardeoin:
Done and rebased onto master

PR #100, opened 2020-07-20
Implemented initial fractional strategy

### LINKED ISSUE #98: Add PLL strategy for fractional clock
The fractional divisor allows setting of precise clocks during runtime using the rcc_pllxfracr register. It needs to be enabled during startup. I purpose adding two new PLL strategies that enable the fractional divisor. One for just below and just above a target value. (e.g. for audio I want a clock that is as close to possible as 48 KHz * 256 without going _under_).

I think the existing iterative strategy could be used to set the base clock and then tweak the rcc_pllxfracr from there.

PllConfigStrategy::FractionalGreaterThan
PllConfigStrategy::FractionalLessThan

mtthw-meyer:
I've implemented a fractional strategy. The math could definitely use double checking. There are some corner cases that may need additional work such as the pll3_r_ck below being below the target value instead of above, but that would require adjusting the other clocks i think, so it may just be a oddity.

Is there a good way to confirm the output clock values short of throwing a scope on it ( I don't have one)?

```
PllConfigStrategy::FractionalNotLess
sys_ck = 400 MHz
pll3_p_ck = 12288002 MHz (target was 12_288_000)
pll3_q_ck = 6144001 MHz (target was 6_144_000)
pll3_r_ck = 3023546 MHz (target was 3024000)
```

Closes #98

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ features = ["const-fn"]
 panic-itm = "~0.4.1"
 cortex-m-rtic = "0.5.3"
 cortex-m-log = { version = "~0.6", features = ["itm"] }
+rtt-target = { version = "0.2.0", features = ["cortex-m"] }
 
 [features]
 default = ["unproven"]
--- a/examples/fractional-pll.rs
+++ b/examples/fractional-pll.rs
@@ -1,0 +1,59 @@
+#![deny(warnings)]
+#![deny(unsafe_code)]
+#![no_main]
+#![no_std]
+
+extern crate panic_itm;
+
+use rprintln as println;
+use rtt_target::{rprintln, rtt_init_print};
+
+use cortex_m_rt::entry;
+use stm32h7xx_hal::rcc;
+use stm32h7xx_hal::{pac, prelude::*};
+
+#[entry]
+fn main() -> ! {
+    rtt_init_print!();
+    let dp = pac::Peripherals::take().unwrap();
+
+    // Constrain and Freeze power
+    println!("Setup PWR...                  ");
+    let pwr = dp.PWR.constrain();
+    let vos = pwr.freeze();
+
+    // Constrain and Freeze clock
+    println!("Setup RCC...                  ");
+    let rcc = dp.RCC.constrain();
+    let ccdr = rcc
+        .sys_ck(400.mhz())
+        .pll3_strategy(rcc::PllConfigStrategy::FractionalNotLess)
+        .pll3_p_ck((48_000 * 256).hz())
+        .pll3_q_ck((48_000 * 128).hz())
+        .pll3_r_ck((48_000 * 63).hz())
+        .freeze(vos, &dp.SYSCFG);
+
+    println!("");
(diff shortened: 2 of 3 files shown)


richardeoin:
Great! In terms of checking, maybe we can extend the existing unit test routines a little more? Particularly for 
q and r ck? Testing on hardware is also possible, a digital counter/logic analyser would be ideal for that rather than an oscilloscope. Combined with sigrok many people like [these analysers](https://sigrok.org/wiki/Hobby_Components_HCTEST0006).

The corner case with `r_ck`happens because p:q:r must always be in an integer ratio (that's fixed in hardware) and the PLL strategy only runs for the requested `p_ck`. Once the `p_ck` is fixed then q and r are calculated [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/rcc/pll.rs#L228) and the integer divisor is always rounded up. Perhaps you could change that to round down for `FractionalNotLess`?

mtthw-meyer:
Extending the tests sounds like a good idea. Unsure how I would go about implementing either of those hardware solutions.

richardeoin:
I re-wrote part of the example to use pll2 and mco2, and then used pulseview to count the pulses on PC9.

|| Value in `_mco2_ck` | Measured | Error
|---|---|---|---
|`mco2_from_hse`| 6_250_000 | 6_248_460 | -246ppm
|`mco2_from_pll2_p_ck`| 6_144_000 | 6_142_490 | -246ppm

Screenshot 2020-07-21 at 21 09 37

This mostly tells us that there's a 246ppm error between the oscillator on my board and my PC. But also shows that the fractional strategy is at least somewhat working

richardeoin:
A  method of testing this with minimal hardware (you need 1 wire) would be to set one timer peripheral to output a divided version of the PLL clock, then use the capture functionality of another timer peripheral to measure it. The HAL timer implementation definitely doesn't support that yet

mtthw-meyer:
Ok. I have a cheap logic analyzer otw I can test the clocks with hopefully this weekend. Seems like a useful tool to have around.

mtthw-meyer:
Ran a couple of tests. Found and fixed a bug. Sampled the MCO2 clock and it seems to be in the right general range.

richardeoin:
I added some more documentation to the new functions (`calc_`...), and fixed the name of the 2nd variable in `calc_ck_div`. That highlights another bug, that `if target_ck < calc_ck(...)` always evaluates to true here.

In fact I was confused why the `calc_ck_div` takes both the VCO output frequency and N divider value. I think there's some confusion between the target and actual/achieved VCO output frequencies (and it was a problem before this PR). In particular the Q and R output dividers should be calculated using the actual VCO output frequency.

richardeoin:
Ok, I think that's resolved it. This has changed the outcome of the fractional_not_less test, the Q divider is now 156 (Q CK = 6.144001MHz) rather than 155 (Q CK = 6.183639MHz) before. I think this is correct.

mtthw-meyer:
Looks good. Built and ran just fine on my daisy board still.

PR #101, opened 2020-07-21
CI: Continue if nightly fails

richardeoin:
Also provide clippy lints. Only `deny`/`forbid` lints will fail the build, unless `#[deny(warnings)]` is also specified.


richardeoin:
It should mean that bors will accept this PR, although github still shows the red cross

bors r+

richardeoin:
If we make nightly a separate job, then perhaps bors will ignore it?

bors r+

PR #103, opened 2020-07-23
Add setter methods for group kernel clock multiplexers.

richardeoin:
This is the last piece of functionality that was missing from the Peripheral REC implementation. This implementation requires all peripherals to be available (through a `&mut`) in order to set group kernel clocks. This ensures it won't affect any active peripherals.

It could be made slightly more flexible by also allowing changes when all the peripherals in the group are available, but that would be more complicated.


src/rcc/rec.rs
```diff
@@ -208,16 +224,16 @@ macro_rules! peripheral_reset_and_enable_control {
...
-
                     $(          // Individual kernel clocks
                         #[doc=$clk_doc]
                         /// kernel clock source selection
                         pub type [< $pk ClkSel >] =
                             rcc::[< $ccip r >]::[< $pk:upper SEL_A >];
                     )*
+
+                    $(          // Group kernel clocks
+                        impl [< $pk_g ClkSelGetter >] for $p {}
```
ryan-summers, line 235: Would recommend referring to the `paste` crate to inform the reader.

src/rcc/rec.rs
```diff
@@ -246,6 +262,32 @@ macro_rules! peripheral_reset_and_enable_control {
...
+                                /// Modify the kernel clock for
+                                #[doc=$clk_doc_g]
+                                /// . See RM0433 Section 8.5.8.
+                                ///
+                                /// It is possible to switch this clock
+                                /// dynamically without generating spurs or
+                                /// timing violations. However, the user must
+                                /// ensure that both clocks are running. See
+                                /// RM0433 Section 8.5.10
+                                pub fn [< kernel_ $pk_g:lower _clk_mux >](&mut self, sel: [< $pk_g ClkSel >]) -> &mut Self {
```
ryan-summers, line 277: Would recommend referring to the `paste` crate to inform the reader, this is a heft macro.

PR #107, opened 2020-07-24
Implemented logger that supports ITM, RTT, and semihosting

### LINKED ISSUE #7: Examples use ITM
May cause a hang when running examples, if debugger is removed or mis-configured.

Feature gating ITM usage is a possible solution.

### LINKED ISSUE #104: [Proposal] Log wrapper
For my own project I wrapped the log crate to allow output via RTT or none via feature. A feature gated logger implementing several implementations sounds really useful. Could include ITM, RTT, etc. Leaving no features selected makes the macro empty and it get's compiled out.

Is this the right level of crate for this or should it go somewhere else?

[Example](https://gist.github.com/mtthw-meyer/0c1d4977cfbb848916f92830c35c852c)

mtthw-meyer:
Implemented logger that supports ITM, RTT, and semihosting.
Updated examples to use the log crate.
I have tested it with RTT and none.
Closes #104 
Closes #7

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,10 +37,17 @@ version = "0.2.5"
 features = ["const-fn"]
 
 [dev-dependencies]
-panic-itm = "~0.4.1"
 cortex-m-rtic = "0.5.3"
-cortex-m-log = { version = "~0.6", features = ["itm"] }
+log = "0.4.11"
+panic-halt = "0.2.0"
+cfg-if = "0.1.10"
 rtt-target = { version = "0.2.0", features = ["cortex-m"] }
+panic-rtt-target = { version = "0.1.0", features = ["cortex-m"] }
+lazy_static = { version = "1.4.0", features = ["spin_no_std"] }
+cortex-m-log = { version = "~0.6", features = ["itm", "semihosting", "log-integration"] }
+panic-itm = { version = "~0.4.1" }
+panic-semihosting = { version = "0.5.3" }
+cortex-m-semihosting = { version = "0.3.5" }
 
 [features]
 default = ["unproven"]
@@ -64,6 +71,9 @@ stm32h743v = ["stm32h7/stm32h743v", "device-selected", "revision_v", "singlecore
 stm32h753v = ["stm32h7/stm32h753v", "device-selected", "revision_v", "singlecore"]
 stm32h750v = ["stm32h7/stm32h743v", "device-selected", "revision_v", "singlecore"]
 stm32h747cm7 = ["stm32h7/stm32h747cm7", "device-selected", "revision_v", "dualcore", "cm7", "dsi"]
+log-itm = []
+log-rtt = []
+log-semihost  = []
 
 
 [profile.dev]
--- a/examples/rtic.rs
+++ b/examples/rtic.rs
@@ -13,6 +13,8 @@ use stm32h7xx_hal::gpio::{Edge, ExtiPin, Floating, Input};
 use stm32h7xx_hal::gpio::{Output, PushPull};
 use stm32h7xx_hal::prelude::*;
 
+use panic_halt as _;
+
 #[app(device = stm32h7xx_hal::stm32, peripherals = true)]
 const APP: () = {
     struct Resources {
--- a/examples/rtic.rs
+++ b/examples/rtic.rs
@@ -3,7 +3,6 @@
 #![no_std]
 #![no_main]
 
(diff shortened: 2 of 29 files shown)


Cargo.toml
```diff
@@ -30,15 +30,23 @@ void = { version = "1.0.2", default-features = false }
 cast = { version = "0.2.3", default-features = false }
 nb = "0.1.2"
 paste = "0.1.18"
+log = "0.4.11"
+panic-halt = "0.2.0"
```
hargoniX, line 34: Does panic-halt really have to be in our default enabled dependencies? I personally don't like the idea of having a default panic implementation as a HAL when usually the user wants to determine that implementation in their crate themselves.

src/logger.rs
```diff
@@ -0,0 +1,87 @@
...
+                inner: semihosting::InterruptOk::<_>::stdout().expect("Get Semihosting stdout"),
+            };
+        }
+
+        pub fn init() {
+            cortex_m_log::log::init(&LOGGER).unwrap();
+        }
+    }
+    else {
+        use panic_halt as _;
```
hargoniX, line 84: See my first comment for this.

src/logger.rs
```diff
@@ -0,0 +1,87 @@
+cfg_if::cfg_if! {
```
hargoniX, line 1: I also do not see anything stm32h7 specific in this file (which is the only file that actually contains library code and not examples as far as I can see? ) so I don't believe this should live in stm32h7xx-hal, after all we are building a Hardware Abstraction Layer and this code does not exactly abstract hardware away, it provides some minimal logic around code that is already general across Cortex-M chips.

examples/rtic.rs
```diff
@@ -3,7 +3,6 @@
 #![no_std]
 #![no_main]
 
-extern crate panic_itm;
```
richardeoin, line 6: Every examples needs a `#[panic_handler]` to compile. For the examples that don't use logging, then I'd say `use panic_halt as _;` is fine

Cargo.toml
richardeoin: Agreed, having panic implementation as a required dependency is not helpful

richardeoin:
I really like using the `log` crate in the examples, because log works across so much of the rust ecosystem. And the example are simpler + more flexible! However having the log implementation as part of the crate is a problem as @hargoniX highlighted.

That can be easily solved by moving `src/logger.rs` -> `examples/logger.rs`, and replacing `use stm32h7xx_hal::logger;` with `mod logger;`. You'd need to document that it's not a standalone example, and tell cargo not to try to build it individually. This should work:

```
[[example]]
name = "logger"
required-features = ["!"] # logger is used by the other examples, not a standalone example
```

Or even move it to `examples/utilities/logger.rs` (for example) to make clear what it is.

Then all the added dependencies can actually be dev-dependencies instead.

mtthw-meyer:
Sure, I can shuffle things around to make that work.

mtthw-meyer:
Yes, `log` is practically a std library.

For building examples I see two options.

1. Leave in features for logging that only effect the examples. Downside is this recompiles the entire crate when any of the features are changed. Upsides are general compatibility with other tools and ease of use.

2. Use cfg flags which require each example to be compiled separately with "cargo rustc --features=stm32h750v,rt --examples -- --cfg log_rtt" and do not work with other tools (e.g. cargo embed can't handle -- --cfg but works fine with --features). I think this only works with cargo rustc invocations.. Advantage being the crate doesn't get recompiled.

Given the limitations of the 2nd approach I favor the 1st. Cargo simply isn't good at handling passing around a bunch of cfg flags yet. Unless there is another approach I am unaware of.

richardeoin:
LGTM! I think the 1st option is the best approach too. It is a bit awkward to have feature flags that don't affect the crate source itself, but it wins on compatibility/ease of use.

The only other thing it would be nice to add is more documentation in `examples/README.md`

mtthw-meyer:
The ITM and semishosting still need to be tested.

richardeoin:
I tested semihosting and ITM, they both work. I've added some more docs for ITM.

richardeoin:
Rebased onto master

mtthw-meyer:
Looks good!

richardeoin:
I fixed the recently added examples that weren't using the logger

richardeoin:
Ok, let's merge this! Thanks for your work @mtthw-meyer 

bors r+

PR #108, opened 2020-07-24
Add USB and example

richardeoin:
The examples require release mode to run correctly.

~~Blocked on:~~
* ~~https://github.com/stm32-rs/synopsys-usb-otg/pull/4~~
* ~~https://github.com/stm32-rs/stm32-rs/pull/401~~
* ~~https://github.com/stm32-rs/stm32h7xx-hal/pull/103~~

There are actually two USB peripherals:

| Name | Type | Base Address | Pin Name
| --- | --- | --- | --- 
| `OTG1_HS` | HS | 0x4004_0000 | `OTG_HS`
| `OTG2_HS` | HS | 0x4008_0000 | `OTG_FS`

~~This PR only adds support for `OTG1_HS`. [synopsys-usb-otg](https://github.com/stm32-rs/synopsys-usb-otg) could support `OTG2_HS` too, except it currently hard-codes the base address (0x4004_0000). It would be interesting to see if this could be fixed with no runtime cost by adding a `const fn` to the `UsbPeripheral` trait. If anyone is interested in picking this up I would be happy to help.~~

Supports both `OTG1_HS` and `OTG2_HS` with https://github.com/stm32-rs/synopsys-usb-otg/pull/5

~~Otherwise I will look at merging this once the blocking PRs are resolved.~~

It would also be interesting to understand why release mode is required for the example. Likely the inlining in release mode is needed to reduce the latency when responding to events, but more experimentation is needed.

### DIFF
--- a/src/usb_hs.rs
+++ b/src/usb_hs.rs
@@ -1,0 +1,90 @@
+//! USB OTG peripherals
+//!
+//! Requires the `usb_hs` feature.
+//!
+//! Note that only full-speed mode is supported,
+//! external high-speed PHY is not supported.
+
+use crate::rcc;
+use crate::stm32;
+
+// use crate::gpio::{
+//     gpioa::{PA11, PA12},
+//     AF10,
+// };
+use crate::gpio::{
+    gpiob::{PB14, PB15},
+    Alternate, AF12,
+};
+
+use crate::time::Hertz;
+
+pub use synopsys_usb_otg::UsbBus;
+use synopsys_usb_otg::UsbPeripheral;
+
+pub struct USB1 {
+    pub usb_global: stm32::OTG1_HS_GLOBAL,
+    pub usb_device: stm32::OTG1_HS_DEVICE,
+    pub usb_pwrclk: stm32::OTG1_HS_PWRCLK,
+    pub pin_dm: PB14<Alternate<AF12>>,
+    pub pin_dp: PB15<Alternate<AF12>>,
+    pub prec: rcc::rec::Usb1Otg,
+    pub hclk: Hertz,
+}
+
+// pub struct USB2 {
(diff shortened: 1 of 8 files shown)


Disasm:
I released `synopsys-usb-otg` v0.2.2 with the relevant changes.

src/usb_hs.rs
```diff
@@ -0,0 +1,93 @@
...
+                    // Reset USB peripheral
+                    rcc.ahb1rstr.modify(|_, w| w.$rst().set_bit());
+                    rcc.ahb1rstr.modify(|_, w| w.$rst().clear_bit());
+                });
+            }
+
+            fn ahb_frequency_hz(&self) -> u32 {
+                // For correct operation, the AHB frequency should be higher
+                // than 30MHz. See RM0433 Rev 7. Section 57.4.4
+                assert!(self.hclk.0 > 30_000_000);
```
Disasm, line 77: BTW, this check is already present in https://github.com/stm32-rs/synopsys-usb-otg/blob/33dc554a8c2ce15e15f577b36d04de7ed3c85034/src/bus.rs#L282-L283

richardeoin: Ok, thanks for finding that. For `HIGH_SPEED` = true, it actually uses the match statement above. Perhaps this is fine because we're only running the bus in Full-Speed mode anyhow?

Disasm: Oh, it's a mistake. The match statement should be executed for `HIGH_SPEED` = false. Actually, inside the driver from CubeIDE this TRDT value is calculated for enumeration speed, not for the maximum speed. As far as I understand, this >30MHz AHB frequency bound is suggested (in 57.4.4) to *guarantee* a correct operation, because you can't tell in advance what speed your device will enumerate into.

PR #110, opened 2020-07-24
Added pwm channel info to pwm::Pins

mlamoore:
Before, it was possible to pass any tuple of TIMx pins to TIMx.pwm(...)
and it would succeed: you could pass two different pins for CH1 or pins
for CH3 and CH4 and either way you would get PWM channels 1 and 2
which would end up writing to the wrong peripheral registers and trying
to use PWM channels with no pins configured.

To fix this, pwm::Pins<TIM> was changed to pwm::Pins<TIM, CHANNEL> and
a pin is now associated with a specific PWM channel.

Tuples of pins and channels are now marked with which channels they
are associated, such as (C1, C2, C3) or (C4, C2)

Pins is only implemented for tuples of channels with no repeated
channels, and each Pins implementation uses the correct Channels tuple
based on the channels for the tuples of pins it is implemented for.

The implementation of Pins<TIM, CHANNEL> for tuples of pins is now
provided by the macro pins_tuples! and includes every valid pin
combination in any possible order (60 total: 6 combinations of 2
channels with 2 orders each, 4 combinations of 3 channels with 6
orders each, and 1 combination of 4 channels with 24 orders)

A tuple of pins passed to pwm(...) now is checked to make sure that
the right pins are associated with the
right channels.

### DIFF
--- a/src/pwm.rs
+++ b/src/pwm.rs
@@ -30,7 +30,8 @@
 //!   let (c0, c1, c2, c3) = device.TIM1.pwm(
 //!       pins,
 //!       100.hz(),
-//!       &mut ccdr
+//!       prec,
+//!       &clocks
 //!   );
 //!
 //!   // Set the duty cycle of channel 0 to 50%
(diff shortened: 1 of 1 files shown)


mlamoore:
I feel like Pins is more of an internal implementation detail; that comment was added more to make the implementation easier to understand.

I like the idea of not adding the redundant generic type parameter to Pins so I tried that, but I run into the issue that I have multiple potentially conflicting implementations of Pins. The new type parameter makes these theoretically conflicting implementations of Pins into separate concrete parameterized implementations, so I think it's easiest to keep it the way I submitted this PR.

That said, I'm pretty new to Rust and I'm definitely open to suggestions.

richardeoin:
Pins is still available outside the crate by necessity (it's part of the public API), so ideally it should still be at least partially documented. The Rust API guidelines have a [more advanced guide](https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed) to how to deal with these traits like that, but perhaps a full `Sealed` implementation would be unnecessarily complex here.

Ok, the bound on the associated type would have made sense for fixing the original problem only. Because we now have many possible permutations, we need to make the `pwm()` and `$timX()` methods generic over the possible Pins. The second generic type parameter allows that easily, so great!

mlamoore:
I added doc comments on Pins, PwmExt, Pwm, and C1-C4. I also fixed the example code in the doc which had wrong parameters to pwm(...).

I'm working on some more advanced PWM features for power electronics and motor control, but that'll be a separate pull request and will be more exploratory.

richardeoin:
Perfect, thanks!

bors r+

PR #113, opened 2020-07-27
i2c: Extend i2c::Write implementation method and add events

richardeoin:
Currently the `write()` method returns once the final byte is written to the FIFO, not when the transaction finishes. Although this saves one byte of time, it is somewhat awkward because:

* Calling another i2c method during this time will fail.
* I2C events are not raised until the transfer ends.

At some point we will need to re-write all of this for async and DMA, but for now I think this is less confusing.


richardeoin:
👍 

bors r+

PR #114, opened 2020-07-30
Add low_power method to PREC structure

richardeoin:
Actually quite a simple addition. This method could be added to the `ResetEnable` trait, but since it is intended for the user and not the HAL, I haven't done this.

Also implement LPTIM[1-5] and add an example that uses this.

* Breaking change to the type of TimerExt (but hopefully no-one was using that).
* Removes `ResetEnable` implementation for HSEM as this doesn't support any low power modes (somewhat obviously), but HSEM needs more consideration anyhow (particularly for dual-core parts).

Blocked on:
* https://github.com/stm32-rs/stm32-rs/pull/404

### DIFF
--- a/examples/rtic_low_power.rs
+++ b/examples/rtic_low_power.rs
@@ -1,0 +1,151 @@
+//! Timers in RTIC with low power mode.
+//!
+//! After the end of init the CPU transitions into CStop mode, and D1/D2
+//! (aka. CD) transition into DStop mode.
+//!
+//! However we set the run_d3 flag, and enable Autonomous mode on the LPTIM3
+//! PREC struture. Therefore LPTIM3 continues to run and fires an interrupt that
+//! wakes the core. Following each interrupt the core returns to CStop mode.
+//!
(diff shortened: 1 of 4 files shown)


richardeoin:
bors retry

richardeoin:
bors try

richardeoin:
@mattico This PR touches some of the same area as #144 because it implements `TimerExt` for the low-power timers (LPTIM). 

For now `timer_tick` is unimplemented for the low-power timers. Does you think it would be useful to implement it?

mattico:
Yes I think implementing that would be useful, though I don't need it at the moment so I don't care if you don't implement it in this PR :) 

Is there some limitation of the LPTIMs that makes it difficult?

richardeoin:
The LPTIMs require a specific initialisation sequence (for example, CFGR must be set **before** enabling the timer, ARR must be set **after**), so it's not difficult but rather a little tedious to check against the RM that it's correct. I'll leave this unimplemented for now.

PR #115, opened 2020-08-02
Improvements to serial

richardeoin:
* Allow baud rate in Hertz/KiloHertz as well as Bps
* Accept a baud rate instead of a configuration struct. In the case that a baud rate is specified, the defaults for the other parameters (8N1) are used.
* Rename `usart` / `usart_unchecked` extension methods to `serial` / `serial_unchecked`. The method names elsewhere match the module name, and also some peripherals are `uart` not `usart`, so I think `serial` is an improvement. Keep the old methods for backwards compatibility.

No breakage, but I have updated the example.

### DIFF
--- a/examples/serial.rs
+++ b/examples/serial.rs
@@ -6,11 +6,11 @@
 use cortex_m_rt::entry;
 #[path = "utilities/logger.rs"]
 mod logger;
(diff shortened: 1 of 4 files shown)


richardeoin:
Good point about removing these eventually. I added a depreciation notice that starts from the next release

PR #116, opened 2020-08-02
Add FMC controller via stm32-fmc crate

richardeoin:
The [stm32-fmc](https://github.com/stm32-rs/stm32-fmc) crate is still a work in progress.

Todo:

- [ ] Add example

### DIFF
--- a/src/fmc.rs
+++ b/src/fmc.rs
@@ -2,7 +2,9 @@
 
 // From stm32_fmc
 use stm32_fmc::FmcPeripheral;
-use stm32_fmc::{PinsSdram, Sdram, SdramChip, SdramPinSet, SdramTargetBank};
+use stm32_fmc::{
+    AddressPinSet, PinsSdram, Sdram, SdramChip, SdramPinSet, SdramTargetBank,
+};
 
 use crate::rcc::{rec, rec::ResetEnable, CoreClocks};
 use crate::stm32;
@@ -45,7 +47,12 @@ pub trait FmcExt: Sized {
     fn fmc(self, prec: rec::Fmc, clocks: &CoreClocks) -> FMC;
 
     /// A new SDRAM memory via the Flexible Memory Controller
-    fn sdram<BANK: SdramPinSet, PINS: PinsSdram<BANK>, CHIP: SdramChip>(
+    fn sdram<
+        BANK: SdramPinSet,
+        ADDR: AddressPinSet,
+        PINS: PinsSdram<BANK, ADDR>,
+        CHIP: SdramChip,
+    >(
         self,
         pins: PINS,
         chip: CHIP,
--- a/src/fmc.rs
+++ b/src/fmc.rs
@@ -1,0 +1,278 @@
+//! HAL for Flexible memory controller (FMC)
+
+// From stm32_fmc
+use stm32_fmc::FmcPeripheral;
+use stm32_fmc::{PinsSdram, Sdram, SdramChip, SdramPinSet, SdramTargetBank};
+
+use crate::rcc::{rec, rec::ResetEnable, CoreClocks};
+use crate::stm32;
+use crate::time::Hertz;
+
+use crate::gpio::gpioa::PA7;
+use crate::gpio::gpiob::{PB5, PB6, PB7};
+use crate::gpio::gpioc::{PC0, PC2, PC3, PC4, PC5, PC6, PC7, PC8};
+use crate::gpio::gpiod::{
+    PD0, PD1, PD10, PD11, PD12, PD13, PD14, PD15, PD3, PD4, PD5, PD6, PD7, PD8,
+    PD9,
+};
+use crate::gpio::gpioe::{
+    PE0, PE1, PE10, PE11, PE12, PE13, PE14, PE15, PE2, PE3, PE4, PE5, PE6, PE7,
+    PE8, PE9,
+};
+use crate::gpio::gpiof::{
+    PF0, PF1, PF11, PF12, PF13, PF14, PF15, PF2, PF3, PF4, PF5,
+};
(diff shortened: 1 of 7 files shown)


src/fmc.rs
```diff
@@ -0,0 +1,278 @@
...
+
+        FMC {
+            fmc: self,
+            _prec: prec,
+            fmc_ker_ck,
+        }
+    }
+}
+
+unsafe impl Sync for FMC {}
```
ryan-summers, line 92: Why do we need to unsafely implement Sync here? Shouldn't the compiler automatically assign sync to the FMC if it's applicable?

I know there are cases where unsafely implementing the Sync is okay (e.g. I've done it for `shared-bus-rtic`), but it could use some documentation?

src/fmc.rs
```diff
@@ -0,0 +1,278 @@
...
+use crate::gpio::gpioh::{
+    PH10, PH11, PH12, PH13, PH14, PH15, PH2, PH3, PH5, PH6, PH7, PH8, PH9,
+};
+use crate::gpio::gpioi::{PI0, PI1, PI10, PI2, PI3, PI4, PI5, PI6, PI7, PI9};
+use crate::gpio::{Alternate, AF12, AF9};
+
+/// Storage type for Flexible Memory Controller and its clocks
+pub struct FMC {
+    fmc: stm32::FMC,
+    _prec: rec::Fmc,
```
ryan-summers, line 38: Why even bother storing the PREC if you're not going to use it? Alternatively, I think you should be able to use it in the `enable()` function instead of taking direct access to the RCC.

src/fmc.rs
```diff
@@ -0,0 +1,278 @@
...
+    }
+}
+
+unsafe impl Sync for FMC {}
+unsafe impl FmcPeripheral for FMC {
+    const REGISTERS: *const () = stm32::FMC::ptr() as *const ();
+
+    fn enable(&mut self) {
+        // unsafe: we own this via the PREC structure
+        let rcc = unsafe { &*stm32::RCC::ptr() };
```
ryan-summers, line 98: I'm not understanding why we have to unsafely take access to the RCC. Wasn't the PREC structure designed to implement the necessary functions below? E.g. enable the peripheral and reset it?

richardeoin: So the methods in [`ResetEnable`](https://docs.rs/stm32h7xx-hal/0.6.0/stm32h7xx_hal/rcc/rec/trait.ResetEnable.html) require us to own the PREC structure (`self`). However the `FmcPeripheral` only gives us a mutable reference (`&mut self`) so we can't call them.

This highlights that the PREC `reset`/`enable` methods are only usable at state transitions when we own the PREC alone, or we are constructing / destructing a struct that contains it.

Perhaps the problem is in the `FmcPeripheral` trait? One option would be to remove the `enable` method entirely. Instead we would document that implementing the `FmcPeripheral` trait requires the AHB bus clock to be enabled already (and we would use PREC to enable it in `FmcExt::new`).

src/fmc.rs
richardeoin: svd2rust implements only implements `Send` for register blocks, and thus FMC will only be automatically marked `Send` but not `Sync`.

Perhaps there's no reason for the `FmcPeripheral` trait to require us to implement `Sync` at all. There's no obvious reason to share it between threads: the use case is to `.init()` it once and that's it.

richardeoin:
Thanks for the comments @ryan-summers! I think these comments are relevant to the design of the stm32-fmc library as well as this PR (which was somewhat expected, since this is the first implementation).

src/fmc.rs
ryan-summers: Yeah, I think the right approach may be to remove the `Sync` requirement from the trait in `stm32-fmc`. The downside of this is that we can't then share the FMC peripheral within various RTIC tasks, but I'm not sure if that's an actual downside, since I wouldn't expect that to be a use case.

I see where this is coming from though, so I could understand a desire to leave it in as well (e.g. tasks may want to dynamically reconfigure the FMC chip?)

src/fmc.rs
ryan-summers: As a workaround to these restrictions, I think you could wrap the `ResetEnable` (PREC) in an [`Option`](https://doc.rust-lang.org/core/option/enum.Option.html). You can then [`take()`](https://doc.rust-lang.org/core/option/enum.Option.html#method.take) it during the enable section to have full ownership of the PREC and then return it into the option at the end.

However, I wanted to comment that this feels like more of a workaround to the fact that `ResetEnable` doesn't take a mutable ref. What's the reasoning behind passing ownership of the `ResetEnable` and consuming `self` instead of just acceptable a mutable ref?

ryan-summers:
@richardeoin I'm definitely seeing how some of these comments are more applicable to `stm32-fmc` after some further review - take a look at some of my feedback on the PREC structure implementations - it may be worth investigating changing the API to use borrowed refs to the PREC instead of taking ownership? That would solve use cases like this one where a struct keeps hold of the PREC.

src/fmc.rs
richardeoin: Agreed that removing the requirement for `Sync` is the right approach - [the use case for `Sync` in RTIC](https://rtic.rs/0.5/book/en/by-example/types-send-sync.html) is actually quite small and I don't think applies here. I'll make a PR to remove it

src/fmc.rs
richardeoin: Agreed this is a good point - I made a separate issue to track this - https://github.com/stm32-rs/stm32h7xx-hal/issues/117

richardeoin:
I've committed one solution that works with the current PREC API, but it'll be worth looking at it again when #117 is resolved

PR #119, opened 2020-08-06
Fix Delay clock divider on Rev V chips

mattico:
image

image


richardeoin:
Looks good to me 👍  I recall we only checked AN5312 for changes, so maybe why this was missed.

bors r+

Issue #102, opened 2020-07-21
Reincorporate stm32h7-ethernet


ryan-summers:
As per the conversation on Matrix and in https://github.com/stm32-rs/stm32h7xx-hal/pull/76, it has been proposed that the various stm32h7 peripheral drivers should be contained within this crate to prevent many interdependent crates in the h7 ecosystem. Instead, peripherals that are not directly related to the [embedded-hal](https://github.com/rust-embedded/embedded-hal) should be placed behind feature gates.

This issue encompasses re-incorporating the `stm32h7-ethernet` crate into this crate.

@hargoniX mentioned that he was planning on getting to this in the near future.

hargoniX:
So has there been any further discussion as to how exactly we will proceed with this?

ryan-summers:
I haven't seem any changes recently, I would assume that it's safe to move forward?

I remember there was some discussion about potentially incorporating `stm32h7-ethernet` into `stm32-eth`, but I think we could also target that as something for the future. It sounds like it would be a non-trivial change.

richardeoin:
Yeah, it'd be great to move this forward.

Integration with `stm32-eth` would be nice, but there would still need to be some H7-specific code. I'm happy to support either approach (`stm32-eth` or here).

hargoniX:
Regarding the smoltcp logging @richardeoin added in the stm32h7-ethernet crate, do we want this to be inside the HAL as well? As of now we do not do any logging so I'm not quite sure whether we need/want it?

ryan-summers:
We talked a bit about logging in Matrix a bit ago - it might not be a bad idea to add a `log` feature flag? But it might also be fine to remove the log messages for now pending a decision on how to log in this crate.

richardeoin:
There is only minimal logging in the stm32h7-ethernet crate, so no problem removing it. There's just one call to `warn!` in `src/ethernet.rs` which could become a `debug_assert!` instead.

The examples could then depend on `smoltcp/log`, that's not a problem imo.

PR #121, opened 2020-08-09
Ethernet merge

hargoniX:
This PR adds stm32h7-ethernet into the HAL behind the ethernet feature flag, trying to introduce as little diff as possible in the process of doing so.

Closes: #102


memory.x
```diff
@@ -45,3 +45,25 @@ _stack_start = ORIGIN(RAM) + LENGTH(RAM);
...
+    } > SRAM1
+  .sram2 (NOLOAD) : ALIGN(4) {
+    *(.sram2 .sram2.*);
+    . = ALIGN(4);
+    } > SRAM2
+  .sram3 (NOLOAD) : ALIGN(4) {
+    *(.sram3 .sram3.*);
+    . = ALIGN(4);
+    } > SRAM3
+} INSERT AFTER .bss;
```
richardeoin, line 69: These are needed for any examples involving DMA, but I think we should be more selective about which ones we allow to be used. My suggestion would be:

* Add a comment explaining that these are used for examples
* Remove `.itcm`
* Replace `.sram1` and `.sram2` with a comment explaining that they are commonly used for the CM4 stack/heap, so should not be used in examples
* Add `.sram4`

That gives one RAM on each of the bus matrixes, which should be fine.

richardeoin: Also, adding these invalidate the line L23 `/* Section definitions below this line are not used by stm32h7xx-hal */`

ryan-summers:
@hargoniX / @richardeoin After getting this in, I would like to hear your potential thoughts on the following:
1. Should we pull the `phy` support out (or provide a generic trait such that users can provide their own phy support if required)?
2. Let's clean up the pins definition - we should be able to take a `Pins` structure in the constructor similar to other peripherals
3. I'd really like to clean up the `unsafe`-ness here, but I'm not familiar as to why its originally there.

Curious to hear your thoughts here - we can break this out into an issue if necessary.

richardeoin:
1. I agree it would be nice to pull out the PHY support. The `PHY` trait is there already in ethernet/mod.rs, but currently we use the feature flags to choose an implementation, rather than allowing the user to pass generic object that implements `PHY`. Definitely fixable.

2. Good point, I'd missed that we don't offer a way to check the pins. Definitely we should keep the same `new` /`new_unchecked` pattern.

3. I think there's always going to be some unsafe because of the DMA interaction. But we can definitely reduce it, especially in the public API. A good place to start would be [`ethernet_init`](https://docs.rs/stm32h7-ethernet/0.2.0/stm32h7_ethernet/fn.ethernet_init.html), I don't think it actually needs to be unsafe. But I think users would need unsafe to take the `DesRing` from static storage?

richardeoin:
@ryan-summers Do you want to merge this now, and then continue with those 3 points in another PR?

ryan-summers:
Let's merge as-is and open issues for updates

richardeoin:
Ok!

bors r+

Issue #95, opened 2020-07-11
ST have renamed D3 to SRD on new A3/B3 parts


adamgreig:
Hi all,

This issue really affects the PACs but I expect it will have an impact on the HAL, and the HAL developers are probably in a good position to comment on it, so I thought I'd ask here first.

It's been pointed out that ST have started calling "SmartRun", "SmartRun Domain", and "SRD" what used to be called "D3". Our SVDs for the A3/B3/B0 parts use "D3", for example in `PWR:CPUCR:RUN_D3` which is now called `RUN_SRD` in the reference manual. It seems like the older H7 parts are not renamed (yet?).

What do you think we should do about it? Options seemingly include:

* Do nothing, keep D3, which maintains easier compatibility with other H7 parts for clock and memory setup, but disagrees with the reference manual
* Rename D3 to SRD in the SVDs, so HAL and user code will have to handle these devices separately
* Other?

I assume future H7 devices will use the new names, which might suggest we stick to the reference manuals and at some point have to rename the current SVDs.

hargoniX:
I think disagreeing with the Reference Manual is usually a bad idea, especially naming wise. Imagine if someone wanted to look into the HAL for a potential bug fix but they find terminology regarding registers that is not mentioned in the reference manual, that doesn't seem like a particularly good idea for people who aren't familiar with the naming issue.

richardeoin:
Yup, I'm ok with renaming the SVDs and cleaning up the difference here in feature gates

richardeoin:
Closed by #122

PR #122, opened 2020-08-10
Add support for stm32h7b3,  stm32h7a3, stm32h7b0

richardeoin:
Not checked in detail, but builds against the nightly stm32h7 PAC.

TODO:

- [x]  Add rm0455-specific clock things to rcc, several changes there
- [x]  Rename d3ccip -> srdccip (ref #95), fix field naming in that register
- [x]  Strip `SDMMC_` prefix from sdmmc registers
 ~~Add new peripherals to REC~~  Needs to be done upstream first
~~Look again at missing items on 7b0~~
- [x]  More cross-referencing with RM0455 / part datasheets
- [x]  Test peripherals with 'awkward' changes: ADC, DAC - **Tested examples on NUCLEO-H7A3ZI-Q**
- [x]  Remove dep on nightly stm32h7 PAC


richardeoin:
I tested all the applicable examples on a NUCLEO-H7A3ZI-Q. There's still plenty of missing peripherals, but this PR is good enough to allow the existing HAL to be used on these parts.

src/rcc/mod.rs
```diff
@@ -485,6 +490,49 @@ impl Rcc {
             },
         };
 
+        // See RM0455 Rev 3 Table 15
+        #[cfg(feature = "rm0455")]
+        let (wait_states, progr_delay) = match vos {
+            // VOS 0 range VCORE 1.25V - 1.35V
+            Voltage::Scale0 => match rcc_aclk_mhz {
+                0..=41 => (1, 0), // Errata? 7A3 Rev Z did not run at 0 wait-states
```
hargoniX, line 498: Might be worth to investigate further before we merge this?

src/delay.rs
```diff
@@ -62,24 +62,16 @@ impl DelayUs<u32> for Delay {
...
-        #[cfg(all(feature = "dualcore", feature = "cm7"))]
-        let mut total_rvr =
-            u64::from(us) * u64::from(self.clocks.c_ck().0 / 8_000_000);
-
-        // CM4 dervived from HCLK
-        #[cfg(all(feature = "dualcore", feature = "cm4"))]
-        let mut total_rvr =
-            u64::from(us) * u64::from(self.clocks.hclk().0 / 8_000_000);
+        let mut total_rvr = if cfg!(not(feature = "revision_v")) {
+            // See errata ES0392 §2.2.3. Revision Y does not have the /8 divider
```
hargoniX, line 66: Unrelated side note, how does stuff like this happen? Do they just forget their divider?

hargoniX (commented):
I don't think I alone can approve this PR, this definitely requires more than one person + ideally a bunch of real world hardware test before being merged...after all its a 1k+ diff.

richardeoin:
Thanks for looking though this @hargoniX. I'm not expecting anyone to completely verify this, that would be pretty much impossible without real hardware anyhow. I think my checks with the NUCLEO-H7A3ZI-Q are sufficient for checking functionality.

Your comments are useful, really I want to discuss/check this is going in the right direction for supporting the different sub-families.

I think we should merge this sooner than later, because it touches so much of the code it'll need manually rebasing with almost every merged PR. Also I'll try to merge it when the number of other PRs is low, because many open PRs will need rebasing on top of it.

src/rcc/mod.rs
richardeoin: Honestly for me this doesn't matter. It would only be important to squeeze out the last bit of performance at low clock speeds. The next step would be to check if ST's example code ever tries setting zero-wait states, and if it does something different.

examples/adc.rs
```diff
@@ -42,7 +42,11 @@ fn main() -> ! {
         .freeze(pwrcfg, &dp.SYSCFG);
 
     // Switch adc_ker_ck_input multiplexer to per_ck
+    #[cfg(not(feature = "rm0455"))]
     let d3ccipr = &unsafe { &*pac::RCC::ptr() }.d3ccipr;
+    #[cfg(feature = "rm0455")]
+    let d3ccipr = &unsafe { &*pac::RCC::ptr() }.srdccipr;
```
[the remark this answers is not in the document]
richardeoin: We do have this now, but I guess the example was never updated. I'll do that now.

src/delay.rs
richardeoin: I'd guess that design components are simulated/verified separately to reduce the complexity, with functional models standing-in for the rest of the chip in each simulation. And this was just missed at some point, then only noticed again when running code on real silicon. Interesting that it took until the second spin "Y" before this was found, so likely the first spin "Z" wasn't even capable of running all their bring-up code (or it wasn't before the deadline to tape-out the second spin).

mattico:
I read through this all, looks good to me. A bit unfortunate that we need so much `#[cfg]` but it is what it is.

Makes me want to set up a CI runner with a bunch of Nucleo boards and defmt's test harness.

src/sai/i2s.rs
```diff
@@ -584,14 +599,17 @@ macro_rules! i2s {
 }
 
 i2s! {
-    SAI2, Sai2: [i2s_sai1_ch_a, i2s_sai1_ch_b],
-    SAI3, Sai3: [i2s_sai2_ch_a, i2s_sai2_ch_b],
-    SAI1, Sai1: [i2s_sai3_ch_a, i2s_sai3_ch_b],
+    SAI1, Sai1: [i2s_sai1_ch_a, i2s_sai1_ch_b],
```
richardeoin, line 602: @mtthw-meyer I guess the mismatch between the peripheral number and method name number is a just mistake. If so I'll just fix here.

PR #124, opened 2020-08-12
Use rcc_aclk for flash wait state calculations

### LINKED ISSUE #120: Flash clock and wait states
To calculate the number of wait states for the flash, we use RM0433 Rev 7. Table 17

image

This table implies the calculation should be based on the AXI Interface clock (= `rcc_hclk`). However the text immediately above it says it should be based on `sys_ck`. Indeed the block diagram (Fig. 5) shows only `sys_ck` entering the flash sub-system.

image

image

Which clock should we use for this calculation? (see Figure 49.)

* sys_ck
* sys_d1cpre_ck ( = sys_ck / 1)
* rcc_hclk ( = sys_d1cpre / 2 by default)

[Currently we use sys_d1cpre_ck](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/rcc/mod.rs#L684), so the flash is always set to 7 wait states for sys_d1cpre_ck > 225MHz. 

TODO:

- [ ] Decide and use correct clock
- [ ] Add flash wait state table for VOS0

richardeoin:
* Add flash wait state table for VOS0
* Add rcc_aclk and use for flash wait state calculations. Closes #120

### DIFF
--- a/src/rcc/mod.rs
+++ b/src/rcc/mod.rs
@@ -441,8 +441,18 @@ impl Rcc {
         // See RM0433 Table 13. FLASH recommended number of wait
         // states and programming delay
         let (wait_states, progr_delay) = match vos {
+            // VOS 0 range VCORE 1.26V - 1.40V
+            Voltage::Scale0 => match rcc_aclk_mhz {
+                0..=69 => (0, 0),
+                70..=139 => (1, 1),
+                140..=184 => (2, 1),
+                185..=209 => (2, 2),
+                210..=224 => (3, 2),
+                225..=240 => (4, 2),
+                _ => (7, 3),
+            },
             // VOS 1 range VCORE 1.15V - 1.26V
-            Voltage::Scale0 | Voltage::Scale1 => match rcc_aclk_mhz {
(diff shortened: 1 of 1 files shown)


src/rcc/mod.rs
```diff
@@ -436,13 +437,27 @@ macro_rules! ppre_calculate {
...
     fn flash_setup(rcc_aclk: u32, vos: Voltage) {
         use crate::stm32::FLASH;
-        let rcc_aclk_mhz = rcc_aclk / 1_000_000;
+        // ACLK in MHz, round down and subtract 1 from integers. eg.
+        // 61_999_999 -> 61MHz
+        // 62_000_000 -> 61MHz
+        // 62_000_001 -> 62MHz
+        let rcc_aclk_mhz = (rcc_aclk - 1) / 1_000_000;
 
         // See RM0433 Table 13. FLASH recommended number of wait
```
mattico, line 446: Nit: RM0433 Rev 7 Table 17

richardeoin: Thanks! I think that was written for Rev 6

PR #125, opened 2020-08-14
Add SPI enable disable

mattico:
With the addition of these functions, it becomes quite easy to share an SPI peripheral between multiple devices which require different configuration. You can `disable()`, `free()`, then pass ownership between the different devices which can then configure the SPI peripheral.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -358,7 +358,7 @@ pub trait SpiExt<SPI, WORD>: Sized {
         freq: T,
         prec: Self::Rec,
         clocks: &CoreClocks,
-    ) -> Spi<SPI, Disabled, WORD>
+    ) -> Spi<SPI, Enabled, WORD>
     where
         PINS: Pins<SPI>,
         T: Into<Hertz>,
@@ -370,7 +370,7 @@ pub trait SpiExt<SPI, WORD>: Sized {
         freq: T,
         prec: Self::Rec,
         clocks: &CoreClocks,
-    ) -> Spi<SPI, Disabled, WORD>
+    ) -> Spi<SPI, Enabled, WORD>
     where
         T: Into<Hertz>,
         CONFIG: Into<Config>;
@@ -394,7 +394,7 @@ macro_rules! spi {
 	    $(
             // For each $TY
             $(
-                impl Spi<$SPIX, Disabled, $TY> {
+                impl Spi<$SPIX, Enabled, $TY> {
                     pub fn $spiX<T, CONFIG>(
                         spi: $SPIX,
                         config: CONFIG,
@@ -437,7 +437,7 @@ macro_rules! spi {
                         spi!(DSIZE, spi, $TY); // modify CFG1 for DSIZE
 
                         // ssi: select slave = master mode
-                        spi.cr1.write(|w| w.ssi().slave_not_selected().spe().disabled());
+                        spi.cr1.write(|w| w.ssi().slave_not_selected());
 
                         // Calculate the CS->transaction cycle delay bits.
                         let cycle_delay: u8 = {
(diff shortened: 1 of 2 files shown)


src/spi.rs
```diff
@@ -551,6 +551,31 @@ macro_rules! spi {
...
+                        self.spi.ifcr.write(|w| w.modfc().clear());
+                    }
+
+                    /// Disables the SPI peripheral. Any SPI operation is
+                    /// stopped and disabled, the internal state machine is
+                    /// reset, all the FIFOs content is flushed, the MODF
+                    /// flag is cleared, the SSI flag is cleared, and the
+                    /// CRC calculation is re-initialized. Clocks are not
+                    /// disabled.
+                    pub fn disable(&mut self) {
```
hargoniX, line 566: I'm assuming that this puts the peripheral in a state where it can no longer send data since it's disabled? If that is the case with the system proposed here it would be easily possible to create an SPI device that can be given to a driver but not actually used by since the peripheral itself is disabled, I don't think that is a particularly good idea.

If that's the case the correct thing to do would be to represent this in the type system IMHO, for example via a type state transition or similar.

mattico:
Could have `disable()` disassemble the device like `free()`

hargoniX:
You could, but in that case your enable() method would need to only take an SPI register struct which means that our API allows to enable a device even if it hasn't been configured before, so that API is unsound as well. I guess a type state pattern would be acceptable to use here but maybe there is other ideas?

mlamoore:
I think a type state pattern makes sense here. Then only Spi<Enabled>
implements the SPI hal trait and calling .disable() gives you a
Spi<Disabled> that can't be passed to drivers that need the hal trait.

richardeoin:
Great, I like the PR. Agreed the enabled/disabled typestate pattern would really improve this, just to add that we already do this for the ADC, the DAC, and there's another implementation hiding in [#114](https://github.com/stm32-rs/stm32h7xx-hal/pull/114/files#diff-dde82bd7fc015253d6e2e7f289dec7baR112) also. Hopefully that's enough examples to copy.

W.r.t disabling the SPI peripheral in `free()`, we do this for some other peripherals (SAI..) but it's not consistent yet. I think disabling in the `free()` method is an improvement (but it's okay for this PR to leave it unchanged)

mattico:
For my use-case all I really need is to disable the peripheral in `free()` which I didn't do because I thought it would be inconsistent with other peripherals.

I'll implement the typestate enable/disable for consistency, though.

mattico:
Most of the SPI functions work whether enabled or disabled, but the typestate helps ensure that you can `free()` and reconstruct an SPI without problems.

https://github.com/stm32-rs/stm32h7xx-hal/pull/125/files#diff-7ee16ebd4be60979ac97f399eca0cc89R46-R53

It seems that type inference doesn't work through `enable()` so you have to annotate the word type. If you try to chain the constructor into `enable()` it loses track of the `pac::SPI` type as well. If anybody knows a way to tweak the generic impls to solve this I'd love to hear it.

hargoniX:
I dont feel good about the fact that the SPI constructor now returns a disabled SPI peripheral which needs to be explicitly enabled afterwards, that is a breaking change to old lib code which we do not have to perform if the SPI constructor returned a default enabled SPI object.

If there is a particular reason we want default disabled here Im missing out on please do explain it to me.

mattico:
I made it Disabled for consistency with the ADC, but the ADC actually has things which need to be done while disabled so I suppose it's the outlier.

hargoniX:
In that case @richardeoin what would be your opinion on consistency with the ADC vs breaking the SPI API beyond the type states?

src/spi.rs
```diff
@@ -99,6 +99,15 @@ pub enum Error {
...
 }
 
+/// Enabled SPI peripheral (type state)
+pub struct Enabled;
+/// Disabled SPI peripheral (type state)
+pub struct Disabled;
+
+pub trait ED {}
+impl ED for Enabled {}
+impl ED for Disabled {}
```
richardeoin, line 109: Is this trait really useful? (I know the other examples have it..)

It's only possible to instantiate `Spi<_,_,_>` in this module, so we can ensure that only `Enabled` and `Disabled` are used here. The generic `impl<EN>` below works just fine without the trait bound.

Then I think this trait can just be removed.

richardeoin:
@hargoniX  I think the initial typestate can be `Enabled` or `Disabled` depending on what's obvious for the API. In this case `Enabled` makes more sense to me (and `Disabled` making more sense for the ADC is also fine)

PR #128, opened 2020-08-26
examples: fix for ITM logging in examples that use CorePeripherals

richardeoin:
ITM logging as added by PR #107

`stm32::CorePeripherals::take()` ([re-exported here](https://github.com/stm32-rs/stm32-rs-nightlies/blob/master/stm32h7/src/stm32h743v/mod.rs#L701)) will only return `Some` once.

Use `steal()` instead, and assume the user won't try to access the ITM
peripheral from any of the examples.


richardeoin:
Using `steal` requires unsafe for obvious reasons, and thus breaks examples marked `#![deny(unsafe_code)]`. I don't see any way around that, so the solution would be to remove the `deny`.

PR #129, opened 2020-08-27
Add a set_timeout method for timers

dflemstr:
I have a use-case where I need to set a specific timer timeout (using interrupts).

Of course it would be possible to convert the delay into a frequency, but then it would involve one division (delay -> Hz) and another one back (Hz -> delay).  It seemed more natural to support duration delays directly.

It might also make sense to use `CountDown<Time=Duration>` as well, but since there is already a `CountDown<Time=Hz>` I couldn't add a conflicting one.  However it might make sense to reconsider whether it makes sense to measure a duration in Hz, as is currently the case in this crate?

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -227,11 +227,7 @@ macro_rules! hal {
                     let frequency = self.timeout.0;
                     let ticks = clk / frequency;
 
-                    let psc = u16((ticks - 1) / (1 << 16)).unwrap();
-                    self.tim.psc.write(|w| { w.psc().bits(psc) });
-
-                    let arr = u16(ticks / u32(psc + 1)).unwrap();
-                    self.tim.arr.write(|w| unsafe { w.bits(u32(arr)) });
+                    self.set_timeout_ticks(ticks);
                 }
 
                 pub fn set_timeout(&mut self, timeout: core::time::Duration) {
@@ -246,8 +242,11 @@ macro_rules! hal {
                     )
                     .unwrap_or(u32::max_value());
 
+                    self.set_timeout_ticks(ticks.max(1));
+                }
 
-                    let psc = u16((ticks.max(1) - 1) / (1 << 16)).unwrap();
+                fn set_timeout_ticks(&mut self, ticks: u32) {
+                    let psc = u16((ticks - 1) / (1 << 16)).unwrap();
                     self.tim.psc.write(|w| { w.psc().bits(psc) });
 
                     let arr = u16(ticks / u32(psc + 1)).unwrap();
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -3,6 +3,7 @@
 // TODO: on the h7x3 at least, only TIM2, TIM3, TIM4, TIM5 can support 32 bits.
 // TIM1 is 16 bit.
 
+use core::convert::TryFrom;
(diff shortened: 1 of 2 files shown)


src/timer.rs
```diff
@@ -234,6 +234,26 @@ macro_rules! hal {
                     self.tim.arr.write(|w| unsafe { w.bits(u32(arr)) });
                 }
 
+                pub fn set_timeout(&mut self, timeout: core::time::Duration) {
```
hargoniX, line 237: time::Duration might surely be a thing to look into, however as of now we usually use our self defined type Milliseconds from time.rs, probably we could look into replacing it with time::Duration though.

However as of now I'd prefer to keep the code consistent unless someone has a good point against it.

hargoniX (changes requested):

src/timer.rs
hargoniX: In fact I would not be surprised if Rust could optimize these conversion lots of the time.

richardeoin:
Thanks for the PR! 

I'd agree with @hargoniX that that we should include the [`Milliseconds`](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/time.rs#L23) type in this. I think the solution is to implement `Into<core::time::Duration> for Milliseconds` in [time.rs](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/time.rs). In that case `set_timeout` could accept either `core::time::Duration` or `Milliseconds`.

I'd also suggest implementing `Into<core::time::Duration>` for all the frequency types. In that case it would be possible to swap over to `CountDown<Time=core::time::Duration>` without breaking any existing code.

Could the four lines `let psc = ...; let arr = ...;` go into their own private method? That would avoid the code duplication.

dflemstr:
I didn't notice the `Into<MilliSeconds> for Hertz` impl in time.rs, but for my use case I need sub-millisecond / more than 1000 Hz timers, so either of these approaches won't work for me unfortunately.  What would be your suggestion for an API that supports more than 1000 ticks per second/setting the timeout to <1ms?

dflemstr:
Maybe it could make sense to have a `Ticks<Clock>` type, to avoid having to perform the division more than once? Just a thought!

richardeoin:
It's clear that you need a sub-millisecond timeout, but that shouldn't mean writing an API that is broken for simpler use cases. Also bear in mind that for this HAL we're not worried about 1 or 2 integer divides, the usability of the API is much more important.

> What would be your suggestion for an API that supports more than 1000 ticks per second/setting the timeout to <1ms?

I explained this above. I'll just add it to the PR now to avoid confusion.

It's possible an intermediate type like `Ticks` might simplify things, it would be interesting to try.

dflemstr:
Now I understand what you meant, not sure how I misread it in hindsight, thanks!

dflemstr:
I'd prefer to merge this PR as is and perhaps try the `Ticks<..>` type as a separate PR if the integer division becomes a problem (which I doubt)!

richardeoin:
Agreed! @hargoniX feel free to merge this

hargoniX:
Or not i guess :p

jordens:
I added you through the stm32h7 team. Try again.

hargoniX:
Miracle!

PR #133, opened 2020-09-04
Added time types for microseconds and nanoseconds

mlamoore:
Operations like specifying timer dead time or pulse width or time delay
make more sense as times rather than frequencies. These time durations
can be measured in (tens or hundreds of) nanoseconds, so it makes sense
to have units to specify durations or delays measured in ns or us.

This PR only adds MicroSeconds and NanoSeconds and conversions between
them and MilliSeconds. I would like to have the 'advanced' timer functions
I'm working on take NanoSeconds as a parameter.

### DIFF
--- a/src/time.rs
+++ b/src/time.rs
@@ -24,6 +24,14 @@ pub struct MegaHertz(pub u32);
 #[derive(PartialEq, PartialOrd, Clone, Copy, Debug)]
 pub struct MilliSeconds(pub u32);
 
+/// MicroSeconds
+#[derive(PartialEq, PartialOrd, Clone, Copy, Debug)]
+pub struct MicroSeconds(pub u32);
+
+/// NanoSeconds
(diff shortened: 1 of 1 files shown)


richardeoin:
Thanks, makes sense. I guess the use case would be something like:

```rust
pwm_channel_1.set_dead_time(50.ns());
```

It's worth looking at #129 which uses `core::time::Duration` for the timers. Could you implement `Into<Duration>` for MicroSeconds and NanoSeconds also? Should be simple with the from_micros/from_nanos methods.

mlamoore:
That's basically what I want it for, except dead time is shared between all
the channels of a timer so I'll have to set it as part of configuring the
timer for PWM. I'm envisioning this:

let channels = tim.advanced_pwm(...).with_deadtime(200.ns()).finalize();

I'll take a look at that other change and implement Into<Duration> and
update this PR.

richardeoin:
Looks good to me!

bors r+

PR #135, opened 2020-09-21
gha: Cache cargo registry/index and target directories between builds

richardeoin:


richardeoin:
Approximately 50% reduction in build time, more on the device flag it happens to cache. Could extend to cache every device flag, but that's unnecessary for now.

bors r+

PR #136, opened 2020-09-22
Add conversions from us/ns to Hertz

mattico:


hargoniX (approved):
LGTM, although every HAL adding their own src/timer.rs which provides a U32Ext trait seems kinda dumb...we should definitely be able to do better than this, but I guess that might be a change we can do another time;

mattico:
There's the embedded-time crate which is being integrated into rtic maybe. We could use it as well.

richardeoin:
+1 on looking into the embedded-time crate. But for now this PR is useful on its own.

bors r+

PR #139, opened 2020-09-24
Impl Copy, Clone, PartialEq for enums

mattico:
I was annoyed when an enum didn't impl Copy so I decided to add these impls to all enums that could use them. I chose `Copy, Clone, PartialEq` since I thought those would be uncontroversial. `Debug` could be useful as well, and some of the enums already impl it, but having lots of `Debug` impls slows compile times and bloats binaries. I didn't add impls to error types since maybe they might want to include some information which would conflict with the impls.

### DIFF
--- a/src/pwr.rs
+++ b/src/pwr.rs
@@ -86,7 +86,7 @@ pub struct Pwr {
 /// Generated when the PWR peripheral is frozen. The existence of this
 /// value indicates that the voltage scaling configuration can no
 /// longer be changed.
-#[derive(PartialEq)]
+#[derive(Copy, Clone, PartialEq)]
 pub enum VoltageScale {
     Scale0,
     Scale1,
--- a/src/pwr.rs
+++ b/src/pwr.rs
@@ -86,7 +86,7 @@ pub struct Pwr {
 /// Generated when the PWR peripheral is frozen. The existence of this
 /// value indicates that the voltage scaling configuration can no
 /// longer be changed.
-#[derive(Copy, Clone, PartialEq)]
+#[derive(PartialEq)]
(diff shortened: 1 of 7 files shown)


src/pwr.rs
```diff
@@ -86,7 +86,7 @@ pub struct Pwr {
 /// Generated when the PWR peripheral is frozen. The existence of this
 /// value indicates that the voltage scaling configuration can no
 /// longer be changed.
-#[derive(PartialEq)]
```
mattico, line 89: Currently only one `VoltageScale` is created which is then passed to `Rcc::freeze()` and never seen again. It doesn't seem to harm anything to make this `Copy`, though.

richardeoin: The mechanics of `VoltageScale` could do with some more work. This idea is that the PWR HAL sets the voltage scale, and the RCC HAL uses this to configure the core and flash correctly for that voltage scale. If PWR returned `VoltageScale::Scale3` but then you passed `VoltageScale::Scale1` to `Rcc::freeze()` that would be bad - likely you'd read corrupt instructions from flash and hardfault.

Currently we allow this, but perhaps we should prevent it. One solution is to declare `VoltageScale` as a struct with  private members (and that does not implement Clone/Copy). However I'd left this as future work - for example there's no functions for changing voltage scale dynamically yet.

Unless you want to work on this, I would leave the current declaration as-is.

richardeoin:
Apart from `VoltageScale`, looks good to me. Thanks!

src/pwr.rs
mattico: I'll leave that to another PR since it's a breaking change.

PR #140, opened 2020-09-24
Update dependencies

mattico:
This is a **breaking change** due to changing the re-exported `Nb` types.


hargoniX:
I'd be in favour of this breaking change since it introduces 1.0 versioons of depndencies which we will have to migrate to sooner or later.

If we upgrade to 1.0 versions though, it might be nicer if we did that all in one release, so one major breaking change in dependencies and then not anymore for a long time.

PR #141, opened 2020-09-24
Ensure VoltageScale remains frozen from Pwr::freeze() to Rcc::freeze()

mattico:
This ensures that the `VoltageScale` configured by `Pwr::freeze()` cannot be changed when it is passed to `Rcc::freeze()`, while still allowing access to the `VoltageScale` value.

See https://github.com/stm32-rs/stm32h7xx-hal/pull/139#discussion_r494572457

This is a **breaking change**.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -13,13 +13,13 @@ PWR. For instance if your board uses the internal LDO to supply VCORE, you
 can replace:
 
 ```rust
-let vos = pwr.freeze();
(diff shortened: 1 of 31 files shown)


richardeoin:
Looks good, thanks. Agreed that the enum can `#[derive(Copy, Clone, PartialEq)]` now, and the rename to `pwrcfg` makes sense.

bors r+

PR #142, opened 2020-09-25
Add MCUDev DevEBox and WeAct MiniSTM32H7 to README

AlyoshaVasilieva:
Adds the MCUDev DevEBox and WeAct MiniSTM32H7 to the readme, relatively cheap development boards.

I've done the basic blinky test for both of these boards. In the case of the DevEBox, I tested the STM32H743VIT6 version.

I'd prefer to link to a documentation source for the DevEBox, rather than their Taobao store, but I cannot find any site they run other than their store. There's a [MicroPython board definition repo by mcauser](https://github.com/mcauser/MCUDEV_DEVEBOX_H7XX_M) that documents some aspects of the board.

I didn't add it to the readme but the WeAct board has read-out protection enabled and I had to manually issue a mass erase command to be able to connect to it. STM32CubeProg's read unprotect command just disconnected the board from USB with no other effect.


richardeoin:

PR #143, opened 2020-09-29
Implement RTC

mattico:
The RTC is a bit strange due to the backup power domain. I tried varying amounts of integration with the standard RCC, REC, PWR methods, but the implementation and semantics conflicted. Instead I added `Backup` to manage the backup power domain. Right now it's just the RTC but hypothetically it could help manage the backup SRAM and the LSE.

~~I started out with a minimal `Date`/`Time` implementation but found myself implementing more and more functionality so I replaced it with the `time` crate. It has a `no_std` mode but always requires an allocator. This isn't an issue for me but I might re-introduce the simple `Date`/`Time` and maybe genericize functions that take one of those as an argument.~~

I moved to the `chrono` crate since it doesn't require allocation.

Interested in any feedback you have.

### Blocked on:

- [x] https://github.com/time-rs/time/issues/280
- [x] RTC HSE divider is configurable according to CubeMX
- [x] stm32h7 (https://github.com/stm32-rs/stm32-rs/pull/446, https://github.com/stm32-rs/stm32-rs/pull/445)

### DIFF
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -18,7 +18,6 @@ mod logger;
 #[entry]
 fn main() -> ! {
     logger::init();
-    let cp = cortex_m::Peripherals::take().unwrap();
     let dp = pac::Peripherals::take().unwrap();
 
     // Constrain and Freeze power
@@ -45,9 +44,17 @@ fn main() -> ! {
         rtc::RtcClock::Lsi,
         &ccdr.clocks,
     );
-    rtc.set_date_time(Utc.ymd(2000, 1, 1).and_hms(0, 0, 0));
+
+    // TODO: Get current time from some source
+    let now = NaiveDate::from_ymd(2000, 1, 1).and_hms(0, 0, 0);
+
+    rtc.set_date_time(now);
     rtc.listen(rtc::Event::Wakeup);
     rtc.enable_wakeup(10);
+
+    unsafe {
+        pac::NVIC::unmask(interrupt::RTC_WKUP);
+    }
 
     loop {
         info!("Time: {}", rtc.time().unwrap());
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -22,7 +22,7 @@ static RTC: Mutex<RefCell<Option<rtc::Rtc>>> = Mutex::new(RefCell::new(None));
 #[entry]
 fn main() -> ! {
     logger::init();
-    let dp = pac::Peripherals::take().unwrap();
+    let mut dp = pac::Peripherals::take().unwrap();
 
     // Constrain and Freeze power
     info!("Setup PWR...                  ");
@@ -54,7 +54,7 @@ fn main() -> ! {
 
     rtc.set_date_time(now);
     rtc.enable_wakeup(10);
-    rtc.listen(rtc::Event::Wakeup);
+    rtc.listen(&mut dp.EXTI, rtc::Event::Wakeup);
 
     unsafe {
         pac::NVIC::unmask(interrupt::RTC_WKUP);
@@ -79,7 +79,8 @@ fn RTC_WKUP() {
         let mut rc = RTC.borrow(cs).borrow_mut();
         let rtc = rc.as_mut().unwrap();
         let date_time = rtc.date_time().unwrap();
-        rtc.unpend(rtc::Event::Wakeup);
+        let exti = unsafe { &mut pac::Peripherals::steal().EXTI };
+        rtc.unpend(exti, rtc::Event::Wakeup);
         date_time
     });
     info!("Time: {}", date_time);
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -3,10 +3,12 @@
 #![no_main]
 #![no_std]
 
+use core::cell::RefCell;
+
 use log::info;
 
 use chrono::prelude::*;
-use cortex_m::asm;
+use cortex_m::{asm, interrupt::Mutex};
 use cortex_m_rt::entry;
 
 use pac::interrupt;
@@ -14,6 +16,8 @@ use stm32h7xx_hal::{pac, prelude::*, rtc};
 
 #[path = "utilities/logger.rs"]
 mod logger;
+
+static RTC: Mutex<RefCell<Option<rtc::Rtc>>> = Mutex::new(RefCell::new(None));
 
 #[entry]
 fn main() -> ! {
@@ -49,21 +53,34 @@ fn main() -> ! {
     let now = NaiveDate::from_ymd(2001, 1, 1).and_hms(0, 0, 0);
 
     rtc.set_date_time(now);
-    rtc.listen(rtc::Event::Wakeup);
     rtc.enable_wakeup(10);
+    rtc.listen(rtc::Event::Wakeup);
 
     unsafe {
         pac::NVIC::unmask(interrupt::RTC_WKUP);
     }
 
+    info!("Time: {}", rtc.date_time().unwrap());
+
+    cortex_m::interrupt::free(|cs| {
+        RTC.borrow(cs).replace(Some(rtc));
+    });
+
     loop {
-        info!("Time: {}", rtc.time().unwrap());
-        rtc.unpend(rtc::Event::Wakeup);
+        // Some debuggers have issues working when the MCU goes to standby/sleep.
+        // You may want to change this to asm::nop() or try to enable low-power debugging.
         asm::wfi();
     }
 }
 
 #[interrupt]
 fn RTC_WKUP() {
-    // Just cause the asm::wfi() above to complete so the loop runs one more time
+    let date_time = cortex_m::interrupt::free(|cs| {
+        let mut rc = RTC.borrow(cs).borrow_mut();
+        let rtc = rc.as_mut().unwrap();
+        let date_time = rtc.date_time().unwrap();
+        rtc.unpend(rtc::Event::Wakeup);
+        date_time
+    });
+    info!("Time: {}", date_time);
 }
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -1,0 +1,60 @@
+//! Example of configuring the real-time clock using the internal LSI oscillator
+
+#![no_main]
+#![no_std]
+
+use log::info;
+
+use chrono::prelude::*;
+use cortex_m::asm;
+use cortex_m_rt::entry;
+
+use pac::interrupt;
+use stm32h7xx_hal::{pac, prelude::*, rtc};
+
+#[path = "utilities/logger.rs"]
+mod logger;
+
+#[entry]
+fn main() -> ! {
+    logger::init();
+    let cp = cortex_m::Peripherals::take().unwrap();
+    let dp = pac::Peripherals::take().unwrap();
+
+    // Constrain and Freeze power
+    info!("Setup PWR...                  ");
+    let pwr = dp.PWR.constrain();
+    let mut pwrcfg = pwr.freeze();
+
+    // Take the backup power domain
+    let backup = pwrcfg.backup().unwrap();
+
+    // Constrain and Freeze clock
+    info!("Setup RCC...                  ");
+    let rcc = dp.RCC.constrain();
+
+    let ccdr = rcc.sys_ck(100.mhz()).freeze(pwrcfg, &dp.SYSCFG);
+
+    info!("");
+    info!("stm32h7xx-hal example - RTC");
+    info!("");
+
+    let mut rtc = rtc::Rtc::open_or_init(
+        dp.RTC,
+        backup.RTC,
+        rtc::RtcClock::Lsi,
+        &ccdr.clocks,
+    );
+    rtc.set_date_time(Utc.ymd(2000, 1, 1).and_hms(0, 0, 0));
+    rtc.listen(rtc::Event::Wakeup);
+    rtc.enable_wakeup(10);
+
+    loop {
+        info!("Time: {}", rtc.time().unwrap());
+        rtc.unpend(rtc::Event::Wakeup);
+        asm::wfi();
+    }
+}
+
+#[interrupt]
+fn RTC_WKUP() {}
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -46,7 +46,7 @@ fn main() -> ! {
     );
 
     // TODO: Get current time from some source
-    let now = NaiveDate::from_ymd(2000, 1, 1).and_hms(0, 0, 0);
+    let now = NaiveDate::from_ymd(2001, 1, 1).and_hms(0, 0, 0);
 
     rtc.set_date_time(now);
     rtc.listen(rtc::Event::Wakeup);
--- a/examples/rtc.rs
+++ b/examples/rtc.rs
@@ -64,4 +64,6 @@ fn main() -> ! {
 }
 
 #[interrupt]
-fn RTC_WKUP() {}
+fn RTC_WKUP() {
+    // Just cause the asm::wfi() above to complete so the loop runs one more time
+}
--- a/src/rcc/backup.rs
+++ b/src/rcc/backup.rs
@@ -1,0 +1,133 @@
+//! Backup Power Domain Reset, Enable, and Clock Control
+//!
+//! Reset, enable and Clock functionality for Backup Power Domain
+//! peripherals. The backup power domain can run separately and be
+//! powered separately from the other domains; these peripherals may
+//! be enabled even after a core reset.
+//!
+//! # Example
+//!
+//! ```
+//! ...
+//! // Constrain and Freeze power
+//! let pwr = dp.PWR.constrain();
+//! let mut pwrcfg = pwr.freeze();
+//!
+//! // Take the backup power domain
+//! let backup = pwrcfg.backup().unwrap();
+//!
+//! if backup.RTC.is_enabled() {
+//!     println!("RTC Enabled");
+//! }
+//! ```
+
+/// Backup Power Domain Peripheral Reset and Enable Control
+#[allow(non_snake_case, missing_docs)]
+#[non_exhaustive]
+pub struct BackupREC {
+    #[cfg(feature = "rtc")]
+    pub RTC: Rtc,
+}
+
+impl BackupREC {
+    /// Creates a new `Backup`
+    ///
+    /// # Safety
+    ///
+    /// Must only be called once, after the backup domain write
+    /// protection is disabled.
+    pub(crate) unsafe fn new_singleton() -> Self {
+        Self {
+            #[cfg(feature = "rtc")]
+            RTC: Rtc {
+                _marker: core::marker::PhantomData,
+            },
+        }
+    }
+}
+
+#[cfg(feature = "rtc")]
+pub use rtc::{Rtc, RtcClkSel};
+
+#[cfg(feature = "rtc")]
+mod rtc {
+    use crate::rcc::rec::ResetEnable;
+    use crate::stm32::RCC;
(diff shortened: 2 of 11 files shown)


hargoniX:
Considering that many peripherals such as for example the ADC do allow multiple clock sources in theory (as of now we limited it to one, but the clock source is actually connected to a MUX we could configure if we so desired inside the chip) would it possibly make sense to have an enum (or a set of structs which implement some trait) which represents clocks we have available to us in general? Since I feel like if we were to actually fully expose the H7 as a Rust interface this is definitely a thing we need to have and reimplementing it over and over again wouldn't be beneficial.

An basic idea without thinking it too much through in depth:

1. Build structs that represent all of our clocks such as the LSI, LSE, HSI etc. etc. These structs would carry information such as the frequency but also more specialised things such as for example the bypass thing for LSI.
2. Possibly implement a way to represent divided clocks as structs, this could be easily done once const generics are a thing actually since we can then simply have a type Divided<CLOCK, DIV> with CLOCK being one of the clock structs mentioned from 1 and DIV being the divider as a const generic. However we don't actually have to wait for const generics since most clocks I've seen so far are usually divided by a multiple of 2 so we can probably get away with providing zero sized PhantomData structs for 2,4,8,16,32,64 and maybe 128.

3. For peripherals that do allow to configure which clock source is their input simply have a function argument of the type T where T is constrainted to implement a marker trait which marks a certain (possibly divided clock source) as usable for this peripheral.

This has the advantage of having a centralized representation of clocks we can use across the chip. Furthermore the Rust compiler as well as docs.rs can still easily tell the user which clocks are allowed for this peripheral:

docs.rs: Check who implementes the peripheral specific marker trait
rustc: If the type does not match, rustc will usually print implementations of the trait it has found for the user so they get docs as a compiler error so to speak.

The disadvantage is of course that we increase our type signature, function signatures etc. by a bit which is not at all an issue for me but I have talked to people that would rather prefer if we did not do that.

Any thoughts on this? Or am I just totally overthinking it at this point xD?

mattico:
Since you didn't mention it, I'll point you to our existing facilities for changing peripheral clocks: https://docs.rs/stm32h7xx-hal/0.7.1/stm32h7xx_hal/rcc/rec/struct.Cec.html#method.kernel_clk_mux (I didn't know/remember this existed until I started working on the RTC). Looking at the implementing macro you can see that many peripherals have no clock mux, many have group clock muxes where they share a clock source, and some have individual clock muxes:

https://github.com/stm32-rs/stm32h7xx-hal/blob/e6da7dc7bf9cd9f85e4f84a980a94e8cbf18be34/src/rcc/rec.rs#L313-L404

I think it's a reasonable abstraction once you know it exists. I think it'd certainly be possible to improve with some typesystem things, maybe represent the tree-like structure better, but I'm a bit wary about the complication. I remember being a rust noob and finding such things very confusing, and rustdoc is not great at exposing them.

Back to the RTC, the reasons I didn't integrate with this existing clock system are:

1. The RTC RCC registers are completely different from every other peripheral. It took [some ugly hacks](https://github.com/stm32-rs/stm32h7xx-hal/commit/13f34fbfd7818bb23a7911e1c961669b296be48f#diff-c0b461220d32c92f125c1eba26e1277f) to work which didn't seem worthwhile for just one peripheral, so I decided to split it out from the `PeripheralRec` struct.
2. The RCC initialization is designed around bringing up a blank-slate reset system. The LSE lives in the backup domain and thus isn't reset unless you reset the whole backup domain (LSE and RTC). There are also lots of specific requirements around configuring the LSE clock, backup domain write protection, and RTC, like you can only set the RTC clock mux one time per backup domain reset, and you can't enable LSECSS until after the RTC is using the LSE.
3. I didn't think that anything else used the LSE clock. Looking at the STM32CubeMX clock configuration tab (why doesn't the reference manual have a clock tree chart!?) now I see that MCO1, CEC, LPTIMs, LPUARTs, and UARTs all can use the LSE. However I think these use-cases are very rare (nobody has asked for them yet) and the RTC is a much more primary use-case for the LSE. I think it's reasonable to ignore that use-case for now given the implementation and interface complexities they'd introduce. There are many places in the HAL where we choose to support the common use-case at the expense of flexibility and implementability.

I can imagine a system where the LSE is moved into the `Backup` which is passed to RCC::freeze() which populates the `lse_ck` in `CoreClocks` so other peripherals can use it. The main issue is reset. LSE and RTC are reset together so they really want to be initialized together. If you can think of a good interface for this I'd love to hear it.

richardeoin:
This is really useful! I haven't looked at it all yet, but looks reasonable so far.

I think implementing `ResetEnable` separately for the backup domain is fine, especially since the `peripheral_reset_and_enable_control!` macro is complicated enough already. [The documentation](https://docs.rs/stm32h7xx-hal/0.7.1/stm32h7xx_hal/rcc/rec/index.html) already does a reasonable job at describing which clocks are possible for each peripheral, although it's certainly possible to improve it more.

Indeed it would be nice to share the LSE state with the CCDR struct somehow, but agreed that's not essential for this PR.

src/backup.rs
```diff
@@ -0,0 +1,107 @@
+//! Backup Power Domain
```
richardeoin, line 1: I think it makes sense to move this file into the rcc folder. You can then have `pub use` statements in rcc/mod.rs to export `Backup` and (conditionally) `rtc`.

src/backup.rs
```diff
@@ -0,0 +1,107 @@
...
+        /// Modify a kernel clock for this
+        /// peripheral. See RM0433 Section 8.5.8.
+        ///
+        /// **NOTE**: This can only be written one time per peripheral reset.
+        /// Check `get_kernel_clk_mux()` to see if the write succeeded.
+        pub fn kernel_clk_mux(&mut self, sel: RtcClkSel) {
+            // unsafe: Owned exclusive access to this bitfield
+            interrupt::free(|_| {
+                let ccip = unsafe { &(*RCC::ptr()).bdcr };
+                ccip.modify(|_, w| w.rtcsel().variant(sel));
```
richardeoin, line 96: I would rename these variables `ccip` -> `bdcr`

src/rtc.rs
```diff
@@ -0,0 +1,639 @@
...
+
+        // Exit initialization mode
+        rtc.isr.modify(|_, w| w.init().free_running_mode());
+
+        Rtc { reg: rtc, prec }
+    }
+
+    /// Reads the value of a 32-bit backup register
+    ///
+    /// **Panics:** if reg is greater than 31
```
richardeoin, line 229: Nice, really good to document these edge cases. There is actually a [standard markdown format](https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure) for this, which is slightly different. Maybe worth using that for consistency?

src/rcc/backup.rs
```diff
@@ -0,0 +1,111 @@
...
+
+        #[inline(always)]
+        /// Modify a kernel clock for this
+        /// peripheral. See RM0433 Section 8.5.8.
+        ///
+        /// # NOTE
+        ///
+        /// This may only be written one time per peripheral reset.
+        /// Check `get_kernel_clk_mux()` to see if the write succeeded.
+        pub fn kernel_clk_mux(&mut self, sel: RtcClkSel) {
```
mattico, line 96: I changed this from `self` to `&mut self` to make `Rtc::handle_lse_css()` only need to take `&mut self`, but it seems safe to me?

mattico:
Thanks for the review! I'll leave this as draft until `stm32h7` 0.13 then I'll rebase.

mattico:
Actually it may be a while until 0.13 so I'll revert to the uglier implementation that works with the undocumented registers for now

src/rcc/backup.rs
richardeoin: Ok, that's reasonable (and certainly not "unsafe" from a memory safety point-of-view).

The reason for requiring `self` is it forces the parent to be mutated when the kernel clock is mutated. You could also argue that actually `handle_lse_css()` should also take `self`, and it should return some new type called `RtcFromLsi` or similar. Thinking in terms of a state machine, `Rtc` and `RtcFromLsi` are the states and `handle_lse_css()` is a state transition - this kind of thinking is common in languages like Haskell and Erlang.

So don't think there's a right-or-wrong answer. Requiring `self` does make it more difficult to use, especially if you're thinking imperatively. But it makes the program easier to reason about if you're thinking in terms of states.

mattico: That makes a lot of sense. I can envison an `Rtc` parameterized by the clock source. The reset situation makes changing between clock states a bit tricky. I'd like to use this more to better understand what kind of interface would be helpful.

richardeoin:
Could you add the new `rtc` feature gate to the CI please? It's [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/.github/workflows/ci.yml#L23) and [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/.github/workflows/nightly.yml#L22), see what https://github.com/stm32-rs/stm32h7xx-hal/pull/81 does for an example. Otherwise the example doesn't get built in CI unfortunately.

examples/rtc.rs
```diff
@@ -0,0 +1,60 @@
...
+    info!("stm32h7xx-hal example - RTC");
+    info!("");
+
+    let mut rtc = rtc::Rtc::open_or_init(
+        dp.RTC,
+        backup.RTC,
+        rtc::RtcClock::Lsi,
+        &ccdr.clocks,
+    );
+    rtc.set_date_time(Utc.ymd(2000, 1, 1).and_hms(0, 0, 0));
```
richardeoin, line 48: This doesn't compile for me:
```
error[E0308]: mismatched types
  --> examples/rtc.rs:48:23
   |
48 |     rtc.set_date_time(Utc.ymd(2000, 1, 1).and_hms(0, 0, 0));
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `chrono::NaiveDateTime`, found struct `chrono::DateTime`
   |
   = note: expected struct `chrono::NaiveDateTime`
              found struct `chrono::DateTime<chrono::Utc>`
```

Perhaps using [NaïveDate](https://docs.rs/chrono/0.4.19/chrono/naive/struct.NaiveDate.html) would be work?

`NaiveDate::from_ymd(2000, 1, 1).and_hms(0, 0, 0)`?

examples/rtc.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    loop {
+        info!("Time: {}", rtc.time().unwrap());
+        rtc.unpend(rtc::Event::Wakeup);
+        asm::wfi();
+    }
+}
+
+#[interrupt]
+fn RTC_WKUP() {}
```
richardeoin, line 60: Doesn't this need to be enabled in the NVIC? That needs to be done for [other examples](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/examples/tick_timer.rs#L55).

examples/rtc.rs
```diff
@@ -0,0 +1,60 @@
...
+use pac::interrupt;
+use stm32h7xx_hal::{pac, prelude::*, rtc};
+
+#[path = "utilities/logger.rs"]
+mod logger;
+
+#[entry]
+fn main() -> ! {
+    logger::init();
+    let cp = cortex_m::Peripherals::take().unwrap();
```
richardeoin, line 21: `cp` should be removed if unused

src/rtc.rs
```diff
@@ -0,0 +1,723 @@
...
+                .bits(mnu)
+                .st()
+                .bits(st)
+                .su()
+                .bits(su)
+        });
+
+        let year = date_time.year();
+        let yt = ((year - 2000) / 10) as u8;
+        let yu = ((year - 2000) % 10) as u8;
```
richardeoin, line 358: What happens if `year` = 2000? I think `calendar_initialized` returns `None`, because the year field in the register is 0.

What happens if `year` < 2000 or `year` > 2099 ? I think we should handle those cases. I'd say `let year = year % 100;` is OK, although then value read back by the `date()` method will be different. Maybe combine that with a `debug_assert`?

mattico: The RTC date register only supports years 2000-2099. I'll add an assert there, though good catch.

richardeoin: Ok, great. I still think `calendar_initialized` returns None for `year` = 2000, which is the case in the example.

src/rtc.rs
```diff
@@ -0,0 +1,729 @@
...
+    /// crystal this value has a resolution of 1/256 of a second.
+    pub fn subseconds(&self) -> Option<f32> {
+        self.calendar_initialized()?;
+        self.wait_for_sync();
+        let ss = f32(self.reg.ssr.read().ss().bits());
+        let prediv_s = f32(self.reg.prer.read().prediv_s().bits());
+        Some((prediv_s - ss) / (prediv_s + 1.0))
+    }
+
+    fn ss_to_us(&self, ss: u16) -> u32 {
```
richardeoin, line 482: It looks like this method returns milliseconds, not microseconds

src/rtc.rs
mattico: Right, the `INITS` just compares the year value to 0 to see if the calendar is initialized. So let's not allow the year 2000.

src/rtc.rs
mattico: Yeah, I remember now. I was trying to avoid overflowing u32 but also avoid needing u64 divide since `__aeabi_uldivmod` is many kilobytes. I'll have to think about what to do here.

richardeoin: Ok, yes that is tricky. You could split the multiply to take advantage of as much u32 precision as possible, like so:

```rust
assert!(ss <= prediv_s); // See RM0433 Rev 7 46.6.10, shift operations not supported

debug_assert!(prediv_s < 32_768);
let tens_us = ((prediv_s - ss) * 100_000) / (prediv_s + 1);

tens_us * 10
```

mattico: You figured that out faster than me, should've come back and checked. I'll add an assert for the possibility of ss > prediv_s.

richardeoin:
The example still doesn't work for me unfortunately. It actually needs to be
```rust
    loop {
        info!("Time: {}", rtc.time().unwrap());
        let date = rtc.date().unwrap(); // Added this line
        rtc.unpend(rtc::Event::Wakeup);
        asm::wfi();
    }
```

It appears to be due to RM0433 Rev 7. 46.3.9 Third Paragraph. The read to `DR` needs to come last. One solution could be a dummy read of `DR` after every read of `SSR` or `TR` (there are several functions with these reads).

richardeoin:
The first two paragraphs also talk about the required frequency of `pclk4`. It would be nice to `assert!` if PCLK4 is less than seven times the RTC clock frequency (fRTCCLK). That could reasonably happen when using the HSE as a source.

mattico:
I really should've tested the example after I made those changes, but I left my nucleo board at home... I'll get it working. Thanks for the review!

richardeoin:
I think it's nearly there!

After the example is working, I'll finish reviewing it 👍

mattico:
Reading through that section of the RM again I see a few options:

1. Fix the order of the shadow register reads and clear the RSF after every read like the RM says. This limits you to one read every RTC clock cycle which is probably 32KHz. I think that could be annoying or surprising for some applications that are trying to read the RTC often for timestamps, etc. and suddenly all their CPU is used waiting around for the RTC. Two solutions to that:
    A. Return `nb::Result` and let the user do work while the RTC is not ready to read. This pushes the complexity to the user, and how many people are set-up to handle async operations properly right now?
    B.  Read the whole time/date in each member function that reads any component, and store the result. Subsequent reads will check the RSF and if it's not set use the memoized value.
2. Disable the shadow register and read directly from the RTC TR/DR/SSR. You now have to read twice to make sure the registers didn't change during your read. *"While BYPSHAD=1, instructions which read the calendar registers require one extra APB
cycle to complete."* So that effectively quadruples the read time since we need to read twice and it takes twice (?) as long, not that it really matters.

It looks like both 1.B and 2 would work, but there's one more thing:

> After waking up from low-power mode (Stop or Standby), RSF must be cleared by software.
The software must then wait until it is set again before reading the RTC_SSR, RTC_TR and
RTC_DR registers.

We don't support low-power mode in the HAL, though it's pretty common to use `wfi()`. We could implement some sort of callback system that tells the RTC when we wake up from sleep/standby; maybe there's a register we can read somewhere that will tell us.

But it would be so much easier to just disable the shadow register. I can't see any reason not to. RSF is convenient to wait for shift and initialization to complete but we can check those flags manually if we have to.

mattico:
There were two additional issues which caused the example to not work:
1. RTC interrupts are routed through EXTI to NVIC, presumably so they can wakeup the core or D3. You need to enable the interrupts in EXTI as well. Thanks ST for not mentioning this anywhere at all.
2. The RTC_WKUP interrupt handler didn't unpend the interrupt so it just ran forever. 🤦‍♂️

@richardeoin this should be ready for review again when you have the time.

richardeoin:
Great, thanks for fixing up the example.

I think that disabling the shadow registers and reading directly (*BYPSHAD=1*) is a good solution. It does mean every read takes a little longer but it's worth it. Both to avoid blocking waiting for RSF to go high, and needing to do things on wakeup. (BTW for the example in release mode, I measure about 500 cycles for the `rtc.data_time().unwrap()` call in the interrupt. That's fine)

I just have one comment about `enable_wakeup`, otherwise this is ready to merge.

src/rtc.rs
```diff
@@ -0,0 +1,849 @@
...
+        if time.ht().bits() == 0 && time.hu().bits() == 0 {
+            return Err(DstError::CannotSubtract);
+        }
+        self.reg
+            .cr
+            .modify(|_, w| w.sub1h().set_bit().bkp().clear_bit());
+        Ok(())
+    }
+
+    /// Configures the wakeup timer to trigger periodically after `interval` seconds
```
richardeoin, line 618: I think the intention is to trigger periodically *every* `interval` seconds. So I would use the word 'every' since it's repeating

src/rtc.rs
```diff
@@ -0,0 +1,849 @@
...
+            let interval = u16(interval - 1 << 16)
+                .expect("Interval was too large for wakeup timer");
+            self.reg.wutr.write(|w| unsafe { w.wut().bits(interval) });
+        } else {
+            self.reg
+                .cr
+                .modify(|_, w| unsafe { w.wucksel().bits(0b100) });
+            let interval =
+                u16(interval).expect("Interval was too large for wakeup timer");
+            self.reg.wutr.write(|w| unsafe { w.wut().bits(interval) });
```
richardeoin, line 641: Section 49.6.6
> the WUTF flag is set every (WUT[15:0] + 1) ck_wut cycles

I think we want to set the flag every `interval` cycles, so it needs a - 1 here.

mattico: That would explain why the example woke up every 11 seconds... ;)

richardeoin:
@mattico What do you think of the rustfmt check in CI? Having consistent formatting is nice, and it means there should never be any formatting noise in diffs. However it's quite annoying when submitting a PR, unless you have rustfmt-on-save in your editor (a personable choice imo).

Could it be an optional check? Even better would be that bors runs rustfmt when merging the PR, but I don't know if bors supports that.

mattico:
@richardeoin I like having the code consistently formatted and I think it's good to have a PR check.

The reason I have trouble with it is that I primarily develop stm32h7hh-hal in a submodule of my firmware project which uses nightly. The nightly rustfmt formats differently from the stable rustfmt that CI uses so I don't use format-on-save and need to remember to `cargo fmt`. I am not good at remembering :)

PR #144, opened 2020-09-29
Timer: Allow configuring tick frequency

mattico:
I want to be able to create a timer which counts up at a specific frequency.  I think this solution is pretty good.

It's a bit awkward since the existing constructors and methods are built around the reload-overflow timer, but the HAL makes it be that way. I can foresee confusion if someone creates a tick_timer than calls `start` or `wait` on it.

I thought about creating a separate timer type for this, we already do for `Pwm`, but it seems silly to duplicate 80% of that code just so we can configure the timer slightly differently.

Perhaps an overhaul of the timer system is in order, but I don't have the energy for that right now.


richardeoin:
Thanks, looks good to me 👍 

Would you also consider adding an example or top-level documentation showing how to use this and particularly the `apply_freq()` method? I think that would help more people discover and use it.

richardeoin:
Great! Just needs a rustfmt again I think.

PR #149, opened 2020-10-13
Use embedded-time for duration and rate types

### LINKED ISSUE #138: embedded-time Crate
Rather than piecemeal implementing time functionality we could use the embedded-time crate which is rather complete and may help interoperability if it becomes widely adopted. Cc #136

mattico:
Closes #138 

A pretty mechanical translation. I added the `Baudrate` and `Timeout` traits to retain the argument-level flexibility we have. I also changed some arguments to use `TryInto` because embedded_time doesn't impl `Into` for fallible conversions so you couldn't have e.g. a `Megahertz` argument as it couldn't convert to `Hertz`.

The consequence of this flexibility is needing to type-annotate the integers. Rustc can't figure out the traits and defaults to `i32`.

Haven't implemented `embedded_time::Clock` for anything yet.

Interested in your thoughts.

## Blocked On
 - [ ] embedded-time 0.11 https://github.com/FluenTech/embedded-time/pull/78


hargoniX:
Since this is a big but very subtile breaking change (after all mhz() only changes to Mhz) I'd suggest we definitely add about this in the changelog so everyone knows what is suddenly wrong with their code.

richardeoin:
Indeed, PRs can optionally add their own changenote entry under `Unreleased` in [CHANGELOG.md](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/CHANGELOG.md) as part of the PR.

src/timer.rs
```diff
@@ -20,9 +21,73 @@ use void::Void;
 use crate::rcc::{rec, CoreClocks, ResetEnable};
 use crate::stm32;
 use crate::stm32::rcc::{d2ccip2r, d3ccipr};
-use crate::time::Hertz;
+use crate::time::{self, Duration, Hertz, Nanoseconds, Rate};
 use stm32h7::Variant::Val;
 
+/// Used to configure periodic timer timeouts using either a duration or a frequency
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum Timeout {
```
richardeoin, line 29: This enum seems generally useful, so it could actually go in `src/time.rs` too

richardeoin:
Thanks @mattico, it's really interesting to see how this looks. Unfortunately I don't think it's quite ready to use. There is actually a really high bar for this, because so far all our required dependencies are either trivial (cast/void) or maintained by the Rust Embedded WG. That's important as we would be forcing everyone to use a crate that's not necessarily mature just yet.

Some specific items:
 * There's many occurrences of `.try_into().unwrap()`. The user probably isn't thinking about possible panics in converting their time/frequency argument when calling our API, so I think the user should have to explicitly make the `try_into()` call themselves. Also using fallible methods within an infallible `From` implementation (at the top of timer.rs) is a problem. Can these `try_into` methods never actually fail in real usage? (because of [this](https://github.com/FluenTech/embedded-time/blob/master/src/fixed_point.rs#L36) perhaps?). Then that should be documented.
 * The extension type methods are not `snake_case`, which is a clear violation of the [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case). I think we should follow the API guidelines for external APIs and examples at the very least.

It's possible that embedded-time isn't the solution for the HAL's internal types. Perhaps instead we could include an example that implements `TryInto<time::NanoSeconds>` for `embedded_time::Duration` (or similar), and then show how that allows interoperability with embedded-time? There's also some nice independent ideas here, like the `Timeout` enum which simplifies the `set_timeout` method on the timers. That could be implemented separately.

mattico:
I had similar thoughts while implementing this, but I put in the work and wanted to hear your opinion. I think it's sensible to only adopt these if they become popular and not before.

PR #151, opened 2020-10-14
Add split_without_reset method to GPIO, and docs

richardeoin:
A quick survey shows the other stm32-rs HALs vary on if they reset the gpio peripheral in the `split` call.

Enable, no Reset:
[f4](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/gpio.rs#L274)
[f7](https://github.com/stm32-rs/stm32f7xx-hal/blob/master/src/gpio.rs#L123)

Enable and Reset:
[f1](https://github.com/stm32-rs/stm32f1xx-hal/blob/master/src/gpio.rs#L291)
[f3](https://github.com/stm32-rs/stm32f3xx-hal/blob/master/src/gpio.rs#L293)
[l4](https://github.com/stm32-rs/stm32l4xx-hal/blob/master/src/gpio.rs#L216)

I think offering both options is reasonable.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -5,7 +5,8 @@ use core::marker::PhantomData;
 use crate::rcc::ResetEnable;
 use crate::stm32::{EXTI, SYSCFG};
 
-/// Extension trait to split a GPIO peripheral in independent pins and registers
+/// Extension trait to split a GPIO peripheral into independent pins and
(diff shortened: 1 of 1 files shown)


hargoniX (approved):
Seems reasonable to me as well!

Issue #130, opened 2020-08-31
Ethernet 


richardeoin:
- [ ] ETH_MACMIIAR_CR is only valid for HCLK 150-250 MHz
- [ ] MDIO interface only supports PHYs with address 0 - #130 
- [ ] Add documentation explaining how to use PHY feature flags - #130 
- [ ] Add ethernet to CI - #130 

See also https://github.com/stm32-rs/stm32h7xx-hal/pull/121

richardeoin:
Closed by #152

PR #152, opened 2020-10-14
ethernet: Remove phy feature flags, allow user to implement their own PHY

richardeoin:
This is a breaking change, because the call to `new_unchecked` no longer resets and initialises the PHY. Now the user has to do that themselves.

* Keep the existing LAN8742A and KSZ8081R implementations available to use
* Add a builder method `set_phy_address` to configure the SMI address for the PHY

### DIFF
--- a/src/ethernet/eth.rs
+++ b/src/ethernet/eth.rs
@@ -30,8 +30,7 @@ use smoltcp::{
     wire::EthernetAddress,
 };
 
-use crate::ethernet::ETH_PHY_ADDR;
-use crate::ethernet::{StationManagement, PHY};
+use crate::ethernet::StationManagement;
 
 // 6 DMAC, 6 SMAC, 4 q tag, 2 ethernet type II, 1500 ip MTU, 4 CRC, 2
 // padding
@@ -399,6 +398,8 @@ pub struct EthernetDMA<'a> {
 ///
 pub struct EthernetMAC {
     eth_mac: stm32::ETHERNET_MAC,
+    eth_phy_addr: u8,
+    clock_range: u8,
 }
 
 /// Create and initialise the ethernet driver.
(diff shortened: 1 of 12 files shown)


richardeoin:
Ref #130

richardeoin:
Closes #130

src/ethernet/eth.rs
```diff
@@ -648,31 +638,58 @@ pub unsafe fn new_unchecked(
     });
 
     // MAC layer
-    let mut mac = EthernetMAC { eth_mac };
-    mac.phy_reset();
-    mac.phy_init(); // Init PHY
+
+    // Set the MDC clock frequency in the range 1MHz - 2.5MHz
+    let hclk_mhz = clocks.hclk().0 / 1_000_000;
```
hargoniX, line 643: I feel like we should be doing this by converting the hertz into megahertz here via an explicit mhz() call but it doesn't seem like that is implemented as of now so it probably doesn't matter. However doing all unit conversion through the type system instead of like this seems like a better idea to me since it's probably going to save us from doing some mistakes in the future.

hargoniX (approved):
LGTM! The type system part in the comment is certainly not a must have.

src/ethernet/eth.rs
richardeoin: Hmm, I'd want to be able to write something like:

```rust
let hclk_mhz: MegaHertz = clocks.hclk().into();

match hclk_mhz.0 { ... }
```

But that's a "lossy" conversion that isn't currently implemented. I had a look at what [embedded-time does](https://github.com/FluenTech/embedded-time/blob/master/src/rate.rs#L585): beneath much indirection it implements conversions in both directions, and actually conversions like MegaHertz -> Hertz are fallible (`TryFrom`) because they could overflow, whilst the opposite Hertz -> MegaHertz is infallible.

In any case I don't see a better way of addressing it in this PR; but the discussion could be continued in another issue.

Issue #80, opened 2020-05-26
DMA API


hargoniX:
During today's embedded WG meeting James mentioned that:
> If you see anyone working on DMA stuff, ping them and see if they could move their work over to the new APIs :)

Referring to these https://github.com/ra-kete/dma-poc APIs.

So @jan561 should maybe try and look into directing his efforts towards getting DMA running towards these APIs, if enough succesul implementations show up they might be added to embedded-hal soon :tm: which would be a big plus embedded Rust.

mtthw-meyer:
I except I'll end up with an implementation of DMA for #90.

hargoniX:
Since embedded-dma has been taken over by the embedded wg the new way to go is going to be to implement that!

mtthw-meyer:
Ok. Is there a issue I can track for that?

richardeoin:
Yup! This is led by the working group, so you need to look there https://github.com/rust-embedded/wg/issues/486

ryan-summers:
`embedded-dma` is now officially supported by the HAL team, so I think it would be safe to use. I have a project that's looking to use memory -> peripheral for QSPI transactions, so I may have some time to devote to this.

mtthw-meyer:
Nice. Having a working use case would make it easier to implement for SAI.

richardeoin:
Agreed! A small [dma module](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/dma.rs) was merged as part of #99, but I think that can just be overwritten with an `embedded-dma` implementation.

mtthw-meyer:
Yeah, that was me taking a stab at getting some basic DMA stuff in. I was originally going to support DMA in the SAI implementation but it proved to be too much at once.

antoinevg:
I also mentioned this on #153 but for reference here's a link to my proof-of-concept example using the SAI peripheral with DMA:

https://github.com/antoinevg/stm32h7xx-hal/blob/sai-dma/examples/sai-dma-passthru.rs

richardeoin:
Closed by #153

PR #153, opened 2020-10-17
Implement DMA based on embedded-dma traits

richardeoin:
DMA implementation based on the [embedded-dma](https://docs.rs/embedded-dma/0.1.2/embedded_dma/) traits and the existing implementation in the [stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal/tree/master/src/dma). ~Currently some parts are very similar - the `Stream` trait and its implementation are completely identical similar, and much of `dma/mod.rs` is similar. However it remains to be seen how much will stay the same when implementing the BDMA and/or MDMA.~ However some changes are required since we'd like to support the BDMA and MDMA also.

Implementation / examples for:
- [x] Memory-to-Memory
- [x] SPI TX
- [x] I2C RX

TODO:
- [x] BDMA implementation
- [ ] Confirm that owning the target peripheral is the right API design
- [x] ~Do the compiler fences also need matching synchronisation~ Memory barriers for the Cortex-M7
- [x] Add an example using RTIC - Thanks ryan-summers!

Contributions to this branch are very welcome, especially implementations and examples for other peripherals.

Ref #80


antoinevg:
I've just pushed a proof-of-concept example using the SAI peripheral with DMA:

https://github.com/antoinevg/stm32h7xx-hal/blob/sai-dma/examples/sai-dma-passthru.rs

I coded this up before you pushed this branch so I'll probably need to go back and update it for your changes!

antoinevg:
Please excuse me if I'm a little off-track here but I'm still very new to embedded-hal implementation!

**re: Soundness of taking a mutable reference to the target peripheral, rather than consuming it completely**

Are we assuming that a given peripheral will only ever have a single DMA stream associated with it?

The SAI peripheral, in particular, has two channels: A & B

Each of these channels has its own DMA stream associated with it.

Usually configured as one stream each for transmitting and receiving audio data.

So there are two related issues here I think:

1. If we are passing a mutable ref or moving the entire peripheral into the `Transfer` constructor for SAI Channel A then I can no longer create a second `Transfer` for SAI Channel B.
1. Because the data register for each SAI channel is a sub-register we cannot pass it into the `peripheral_target_address!` macro.

Would it make sense to modify `Transfer` to just accept the address of the data register rather than the whole peripheral?

i.e. something like:

```
let mut dma1_str0: Transfer<_, _, MemoryToPeripheral, _> = Transfer::init(
    dma1_stream.0,
    sai1.target_address(),
    tx_buffer,
    None,
    dma1_config,
);

let mut dma1_str1: Transfer<_, _, PeripheralToMemory, _> = Transfer::init(
    dma1_stream.1,
    sai1.target_address(),
    rx_buffer,
    None,
    dma1_config,
);
```

And also modify the `peripheral_target_address!` macro to allow us to specify a sub-register using something like:

```
peripheral_target_address!(
    (pac::SAI1, cha.dr, u32, M2P, DMAReq::SAI1A_DMA),
    (pac::SAI1, chb.dr, u32, P2M, DMAReq::SAI1B_DMA),
);
```

mattico:
@antoinevg 

My expectation is that it would work like `Serial.split()` but returning some `Channel<>` struct that's just used for DMA.

I'm not familiar with the SAI peripheral but for example if you need to ensure the peripheral settings aren't changed during the DMA transfer, do it like `Serial.split()` and consume the serial port while the DMA is happening, then you can have a function that consumes the `Channel`s and gives you access to the peripheral again.

If you don't need to ensure that peripheral settings aren't changed during the DMA transfer you can just have the peripheral store an `Option<Channel>` that you `take()` so you can't get multiple copies of the channel.

If you need to ensure that only some peripheral settings aren't changed during the DMA transfer you can do something like `Serial.split()` but also return a version of the peripheral with some typestate so that certain functions aren't implemented in that state (see `spi::Enabled` and `spi::Disabled`).

mattico:
I've had some DMA code using the serial peripheral: https://github.com/mattico/stm32h7xx-hal/tree/add-dma-example
I'm going to rebase it on this and try to modify the HAL serial to support DMA.

mattico:
@richardeoin Re: Do the compiler fences also need matching synchronisation barriers for the Cortex-M7?

https://community.st.com/s/article/FAQ-DMA-is-not-working-on-STM32H7-devices has good information about this if you haven't seen it.

From what I remember, I couldn't get DMA to work with the serial port without [disabling the dcache on the DMA buffers](https://github.com/mattico/stm32h7xx-hal/blob/c7c36fde8855d8fe43357be5c37c07f37c782ff3/examples/dma_uart_tx.rs#L214-L230). No combination of `dsb()` `compiler_fence()` or even not enabling the global dcache (!) worked. (so I was probably doing something wrong...)

adamgreig:
With DMA in a cached region you'd have to use the cache control operations to specifically invalidate (on DMA writes) or clean (on DMA reads) the memory in use for DMA.

antoinevg:
@mattico 

> My expectation is that it would work like `Serial.split()` but returning some `Channel<>` struct that's just used for DMA.
> \<schnip/\>

Thank you for taking the time to answer.

I think I understood about 1% of what you were saying there, I really meant it when I said I'm new to embedded-hal development! 🤣

Could you show what you mean in a few lines of hypothetical code? Alternatively, is there a tutorial or documentation I could read that can help bridge the gaps between the working SAI/DMA implementation code I posted earlier in this thread, my two questions and your reply?

I could probably piece things together on my own from the sources but time is expensive and it would be cool if I could put the little I have towards contributing functionality rather than reverse engineering!

richardeoin:
Hi @antoinevg @mattico ! Those examples are great, thanks for sharing them. Also nice that you're trying to update them to work with this PR. 

Indeed DMA with the SAI isn't well supported without several changes, so right now you're running a bit ahead of where this PR is @antoinevg 

> Are we assuming that a given peripheral will only ever have a single DMA stream associated with it?

Yes, the current instances of `peripheral_target_address!(` assume that a given peripheral only has one associated DMA stream (at once). That's a result of us representing ownership to the compiler in blocks of 'peripheral', which is the approach most current PACs take (but not the only possible approach).

Indeed the approach taken by the serial HAL would be a good option. Here's the function where the serial HAL (which owns the serial PAC) is [split](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/serial.rs#L549) into two [ZSTs](https://runrust.miraheze.org/wiki/Zero-sized_type). Here we're using ZSTs to represent ownership, whilst not consuming resources at runtime.

You'd then implement the DMA like so:
```
peripheral_target_address!(
    (Rx<pac::USART1>, rdr, u8, P2M, DMAReq::USART1_RX_DMA),
    (Tx<pac::USART1>, tdr, u8, M2P, DMAReq::USART1_TX_DMA),
);
```

Potentially this approach would work for the SAI channels too.

richardeoin:
@mattico The link about memory layout and dcache is useful, although it's a separate issue (I think). This PR doesn't consider the dcache at all yet, it just assumes it's off.

richardeoin:
@antoinevg I added an [extra option](https://github.com/stm32-rs/stm32h7xx-hal/pull/153/files#diff-2b1a1e3a5b0cfbc820a67da5f1fb3dcb8830f5f317ef2557a140adb5165cf6e3R50) to the `peripheral_target_address!` macro, so the proposal from your comment above should compile now if you add it!


```
peripheral_target_address!(
    (pac::SAI1, cha.dr, u32, M2P, DMAReq::SAI1A_DMA),
    (pac::SAI1, chb.dr, u32, P2M, DMAReq::SAI1B_DMA),
);
```

If you have a neater way of modifying the macro that's also great, I didn't spend too much time thinking about it.

Of course it doesn't solve the problem with actually initialising both streams at the same time, since they both want a mutable reference to pac::SAI1

antoinevg:
> @antoinevg I added an [extra option](https://github.com/stm32-rs/stm32h7xx-hal/pull/153/files#diff-2b1a1e3a5b0cfbc820a67da5f1fb3dcb8830f5f317ef2557a140adb5165cf6e3R50) to the `peripheral_target_address!` macro, so the proposal from your comment above should compile now if you add it!

Awesome, working nicely thank you!

> Of course it doesn't solve the problem with actually initialising both streams at the same time, since they both want a mutable reference to pac::SAI1

We also need access to pac::SAI1 after DMA has started up.

The SAI peripheral is a bit weird in that it needs DMA to be up & running before setting you can set its `dmaen` and `en` bits.

I see that `transfer.start(|sai1| {})` supplies a reference to the peripheral's register block so I could just twiddle the bits directly. 

That said, wouldn't it be better to rather have the hal peripheral wrapper passed so we can call e.g. `enable_dma()` and `fifo_is_empty()` methods?

richardeoin:
> I see that transfer.start(|sai1| {}) supplies a reference to the peripheral's register block

Yup, this closure allows access to whatever memory the Transfer has exclusive mutable access to. That that would be the place to set the `dmaen` and `dma` bits. I [experimented](https://github.com/stm32-rs/stm32h7xx-hal/pull/153/files#diff-d531c3a6fad2af5abe1c6987a5968864fe5b257e6e1a60f0861ce1c9b55baaa5R482) with implementing TargetAddress for both the PAC and HAL types for I2C. It seems to work well, so I think that can be done for other peripherals too.

> It looks like the SAI peripheral needs the DMA to be fully configured before it can be successfully configured

Ah, you mean that the DMA must be configured before the call to `SAI1.i2s_ch_a(...)`? That does make it very difficult to give the transfer exclusive ownership to anything.

If you're willing to use `unsafe`, then working around this is quite easy. The most obvious escape hatch that already exists is the steal() method.

```rust 
let mut sai1_a_for_dma = unsafe { pac::Peripherals::steal().SAI1 };
let mut sai1_b_for_dma = unsafe { pac::Peripherals::steal().SAI1 };

let mut transfer_ch0: Transfer<_, _, PeripheralToMemory, _> = Transfer::init(
        streams.0,
        &mut sai1_a_for_dma,
        ...);
let mut transfer_ch1: Transfer<_, _, MemoryToPeripheral, _> = Transfer::init(
        streams.1,
        &mut sai1_b_for_dma,
        ...);
```

The usefulness of this construction is an argument that TargetAddress should be implemented directly on the PAC/HAL struct itself, rather than a [mutable reference](https://github.com/stm32-rs/stm32h7xx-hal/pull/153/files#diff-2b1a1e3a5b0cfbc820a67da5f1fb3dcb8830f5f317ef2557a140adb5165cf6e3R16) to them. Then you could write:

```rust
let mut transfer_ch0: Transfer<_, _, PeripheralToMemory, _> = Transfer::init(
        streams.0,
        unsafe { pac::Peripherals::steal().SAI1 },
        ...);
```

mtthw-meyer:
FWIW when I was working on SAI I was thinking the SAI would be handed off to the DMA.

antoinevg:
Thank you for the suggestions @richardeoin, they helped a lot.

We've now got a first iteration of the `sai_dma_passthru` example implemented against `embedded_dma`: 

https://github.com/antoinevg/stm32h7xx-hal/commit/e01b7e8cd22a88730105e9005b29b72ee5190e0c

There are still a few things which are not ideal, but hopefully this gets us a step closer.

**Notes:**

1. I had to add a type-alias for the other DMA Request Multiplexer for SAI3/4 in [src/dma/dma.rs:755](https://github.com/antoinevg/stm32h7xx-hal/blob/e01b7e8cd22a88730105e9005b29b72ee5190e0c/src/dma/dma.rs#L757)
1. It would be nice if we could also implement `TargetAddress` for `u16`'s  
1. It doesn't look like `dma::Transfer` has support for configuring the circular buffer so I still needed to reach for the escape hatch there.
1. I couldn't figure out the type signature for `dma::Transfer` 🤯 so I also needed to reach for the hatch in the interrupt handler. This is also blocking me from doing an implementation using `rtic`.

richardeoin:
Great! The example looks reasonable already, and it's very useful to have to play around with.

Some quick responses to your notes:

1. You almost certainly should not do that. DMA1/2 do not have any request lines from SAI4, those lines go to the BDMA instead. So SAI4 transfers (unlike SAI1/2/3) should be defined in bdma.rs instead.
2. Yes, I think for some peripherals it would be useful to implement more than one of u32,u16,u8. At the

richardeoin:
Great! The example looks reasonable already, and it's very useful to have to play around with.

Some quick responses to your notes:

1. You almost certainly should not do that. DMA1/2 do not have any request lines from SAI4, those lines go to the BDMA instead. So SAI4 transfers (unlike SAI1/2/3) should be defined in bdma.rs instead.
2. Yes, I think for some peripherals it would be useful to implement more than one of u32,u16,u8. At the moment it's not possible because `MemSize` is an associated type, not a generic type. It wouldn't be too complicated to add, but it would add a 5th(!) parameter to the type signature of `dma::Transfer`.
3. Indeed there's no support for the circular buffer, I haven't looked at that at all.
4. The type signature is something like `dma::Transfer<dma::dma::Stream1<DMA1>, &mut stm32::SAI1, dma::PeripheralToMemory, &mut [u32; 128]>`. To use it with RTIC you'll need to specific the static lifetime for two references. That shouldn't be too hard for the buffer (`&'static mut [u32; 128]`) but it's more complicated with the peripheral.  It's another good reason that TargetAddress should be implemented directly on the PAC/HAL struct itself, rather than a mutable reference. So I'll probably change that soon.

richardeoin:
Rebased onto master

antoinevg:
Thank you again for the feedback @richardeoin!

I've pushed a second iteration of the example:

https://github.com/stm32-rs/stm32h7xx-hal/compare/dma...antoinevg:sai_dma_passthru_example_v0.2?expand=1

1. I fixed the target address macro entries for SAI4 - *\*whoops\**
1. Updated my code to take advantage of the fact that TargetAddress is now directly on the PAC/HAL struct. 👍 

I also:

1. Added support for enabling circular buffering to `DmaConfig` and the `DoubleBufferedStream` trait.
1. Added support for the half transfer interrupt to the `Stream` trait.

I'm not sure if this fits in with how you want to do things, but I thought it would be interesting to see what it would look like and just went with it.

richardeoin:
Hi @antoinevg, your changes look good! If you're happy with the example, would you be able to open a PR to add those changes to the dma branch? (You should see a prompt when you click on the link you posted)

ryan-summers:
We're looking to heavily leverage DMA for https://github.com/quartiq/stabilizer and our use case looks like the following:
* DMA stream 0 configured to trigger off a periodic timer in circular mode - DMA transaction writes data (data is always 0) into SPI TXFIFO to trigger a SPI transaction (SPI MOSI is disconnected, so the TX data is dont-care)
    - This provides us with a periodic ADC read over SPI entirely managed by hardware - we need all available CPU cycles for data processing
* DMA stream 1 is configured to pump the SPI RX data into a double-buffered, circular transfer. Interrupt generated when each half of the double-buffer transfer completes
* Stream 2-3 are utilized for an identical purpose on a different SPI instance

There's also more streams necessary for generating output writes to a DAC, but I'll leave them out of this post for now.

As such, the existing implementation _almost_ fits our needs - the main issue I see is that the DMA stream owns the SPI peripheral, so it's not currently possible to use DMA for both RX and TX, which is our use case. It seems like we could borrow the peripheral only for the purposes of getting the target address, and then perhaps add some marker into the peripheral to indicate that the TX/RX path is handled by DMA, which would then disable those functions?

It seems like the peripheral doesn't necessary need to be consumed, but is just consumed for safety reasons, which limits the capabilities of the HAL. Open to discussion on this point

mtthw-meyer:
Let me know when it's ready and I'll test the SAI DMA on the daisy board.

richardeoin:
@mtthw-meyer Great! The dma branch now includes the `sai_dma_passthru` example that @antoinevg has been working on (and I think it's written to run on the daisy already)

ryan-summers (changes requested):
I've been using the branch for my own code and I've been using the following workaround for the ownership issues:

Instead of giving the actual peripheral to the DMA transfer, I create a ZST that implements the necessary traits for DMA. I then give DMA the ZST instead of the peripheral if I need to keep ownership, so there's plenty of work-arounds to the existing approach of peripheral ownership. I don't think it's much of a deal-breaker.

However, that being said, The 3 use cases I've seen (mine, one by James Munns, and the I2S example previously added here), have _all_ needed to implement some sort of work-around to the DMA transfer owning the peripheral, which indicates that the peripheral ownership semantics aren't flexible enough for the general use-case. 

Given the observations above, I would advocate for removing the peripheral ownership implementation. However, if we decide to continue forward, there are workarounds, it just requires some moderate boiler-plate on the user side.

src/dma/dma.rs
```diff
@@ -0,0 +1,841 @@
...
+        u8,
+        M2P,
+        DMAReq::SPI2_TX_DMA
+    ),
+    (pac::SPI3, rxdr, u8, P2M, DMAReq::SPI3_RX_DMA),
+    (pac::SPI3, txdr, u8, M2P, DMAReq::SPI3_TX_DMA),
+    (pac::SPI4, rxdr, u8, P2M, DMAReq::SPI4_RX_DMA),
+    (pac::SPI4, txdr, u8, M2P, DMAReq::SPI4_TX_DMA),
+    (pac::SPI5, rxdr, u8, P2M, DMAReq::SPI5_RX_DMA),
+    (pac::SPI5, txdr, u8, M2P, DMAReq::SPI5_TX_DMA)
```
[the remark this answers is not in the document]
ryan-summers: Follow-up: I'm preparing a PR to address this.

src/dma/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+impl<STREAM, PERIPHERAL, DIR, BUF> Drop
+    for Transfer<STREAM, PERIPHERAL, DIR, BUF>
+where
+    STREAM: Stream,
+    PERIPHERAL: TargetAddress<DIR>,
+    DIR: Direction,
+    BUF: WriteBuffer<Word = <PERIPHERAL as TargetAddress<DIR>>::MemSize>
+        + 'static,
+{
+    fn drop(&mut self) {
```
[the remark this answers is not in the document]
ryan-summers: An example is using a circular transfer to write a single word into the SPI TXFIFO to trigger a transaction. The transfer may be triggered by a timer overflow, which results in a periodic SPI read. The actual transfer and TX contents are don't care, but the transfer should continue on infinitely and there's no value of keeping the `Transfer` around since it is never modified.

src/dma/mod.rs
```diff
@@ -0,0 +1,799 @@
...
+    }
+}
+
+/// DMA Transfer.
+pub struct Transfer<STREAM, PERIPHERAL, DIR, BUF>
+where
+    STREAM: Stream,
+    PERIPHERAL: TargetAddress<DIR>,
+    DIR: Direction,
+    BUF: WriteBuffer<Word = <PERIPHERAL as TargetAddress<DIR>>::MemSize>
```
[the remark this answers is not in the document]
ryan-summers: I believe the implication here is that we can't provide a constant buffer in a MemoryToPeripheral transfer since we have to provide a mutable ref. This is not desirable if we are doing e.g. a circular transfer of some constant code-word

src/dma/dma.rs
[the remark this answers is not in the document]
ryan-summers: https://github.com/stm32-rs/stm32h7xx-hal/pull/164

ryan-summers:
I'm back to add some context after further developing with DMA (if curious, my implementation is here: https://github.com/quartiq/stabilizer/pull/165).

After further review, I think the ownership of the peripheral is the correct approach.

The STM32H7 DMA implementation should only have a single DMA request line connected to a signle DMA stream or there's contention on acknowledging the request from multiple streams. This can result in weird hangup/stalled peripherals when one request is provided to multiple streams. Using the ownership API as implemented prevents this because the user will not be able to provide a single peripheral (and thus a single request) to more than one stream.

src/dma/mod.rs
```diff
@@ -0,0 +1,798 @@
...
+        cortex_m::asm::dmb();
+
+        // "Preceding reads and writes cannot be moved past subsequent writes"
+        compiler_fence(Ordering::Release);
+
+        unsafe {
+            self.stream.enable();
+        }
+
+        Ok((old_buf.unwrap(), CurrentBuffer::FirstBuffer))
```
ryan-summers, line 576: In cases where the user is stopping a DMA transfer before it completes, this doesn't provide a means of detecting how many elements in `old_buf` are valid (in the case of MemoryToMemory or PeripheralToMemory). Perhaps we should also return the number of data actually transferred here? Alternative would be the number remaining, but they're essentially the same thing.

richardeoin:
> Using the ownership API as implemented prevents this because the user will not be able to provide a single peripheral (and thus a single request) to more than one stream.

That's useful to know, I hadn't realised the limitation of one-stream-per-request-line existed.

For those peripherals that have two request lines (one Rx, one Tx) I think the end solution is to implement `split()` for those peripherals. But for now the ownership API needs to be worked-around with `unsafe`/`steal` in those cases, in order to setup a stream for each request line.

ryan-summers:
The snippet is in RM0433 17.4.4:
image

The behavior I saw when doing this was that one stream would report as ENABLED, NDTR = 0, and trying to clear the ENABLE bit was unsuccessful (channel permanently held on with 0-data size). A processor reset was required to clear up the fault. After switching to using two separate DMA requests, the problem disappeared.

src/dma/mod.rs
ryan-summers: Resolved in https://github.com/stm32-rs/stm32h7xx-hal/pull/173

richardeoin:
@ryan-summers Could you take a look at the merge above? I had to re-work much of #173 to merge #171, in the end it was quite simple but I hope it still works (On hardware I only checked that `remainder` was zero after non-cancelled transfers).

ryan-summers:
I gave it a review and did some hardware testing with the `dma` branch after the merge and everything is looking good from my end

src/dma/mod.rs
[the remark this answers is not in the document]
richardeoin: I've proposed one way of addressing this in https://github.com/stm32-rs/stm32h7xx-hal/pull/178

richardeoin:
Breaking change: Added additional generic type on `Transfer`. See #178

src/dma/mod.rs
```diff
@@ -0,0 +1,726 @@
...
+            #[inline(always)]
+            pub fn get_half_transfer_flag(&self) -> bool {
+                STREAM::get_half_transfer_flag()
+            }
+        }
+    };
+}
+
+db_transfer_def!(DBTransfer, init, StaticWriteBuffer, write_buffer, mut;);
+db_transfer_def!(ConstDBTransfer, init_const, StaticReadBuffer, read_buffer;
```
ryan-summers, line 621: This actually precludes single transfers from using the `next_transfer()` API, which is still highly desirable even for single-buffer-mode, since you may want to continue using the same transfer with a new source/dest buffer (such is the case where asynchronous data collection from an external course)

ryan-summers: Note that we can currently work around this by falsely claiming those types of transfers are `DBTransfer` or `ConstDBTransfer`

richardeoin:

src/dma/mod.rs
richardeoin: Is precluding single transfers from using `next_transfer()` highly desirable? Or not?

I think this my be a naming issue, because here "DB Transfer" means any transfer (including single buffered ones) happening on a DMA that supports double buffering. So that would be any transfer on DMA1, DMA2 or a BDMA. I see how this naming could be confusing, and would be solved by having a better collective description for transfers on (DMA1 or DMA2 or BDMAs).

richardeoin: @ryan-summers is this comment still open? Other than this comment, I think we can merge the dma branch

ryan-summers: As pointed out, this actually works just fine. I was just confused by naming conventions.

ryan-summers (approved):
LGTM! We can always implement future changes if we want to make some things cleaner (like supporting full duplex DMA), but this is a wonderful start

PR #156, opened 2020-10-27
An example for EXTI interrupts

leprechaun:
Some of the folks on the chat helped me out getting some hello world type stuff working this evening. Might as well open a PR with an extra example, in case this could help someone else out.

Thanks peeps

### DIFF
--- a/examples/digital_read.rs
+++ b/examples/digital_read.rs
@@ -1,0 +1,40 @@
+#![no_main]
+#![no_std]
+
+use stm32h7xx_hal::hal::digital::v2::InputPin;
+
+use cortex_m_rt::entry;
+use stm32h7xx_hal::{pac, prelude::*};
+
+#[macro_use]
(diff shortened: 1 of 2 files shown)


richardeoin:
Thanks for this, it's awesome that people are starting with the crate and contributing! The PR is great, but there's just a couple of best-practice points:

Could you fix the clippy warning? (see the github actions view).

```
examples/exti_interrupt.rs#L94
unnecessary parentheses around block return value
```

Some of the other simple examples to have a `#![deny(warnings)]` at the top. That helps the CI catch bit-rot early, especially if dependencies changes. Does adding that work here?

leprechaun:
Thanks for the heads up, I hadn't noticed the warnings generated in the check. Fixed all of my own with `deny_warnings` on.

richardeoin:
Nice!

bors r+

PR #159, opened 2020-10-28
Updating SPI CS management

### LINKED ISSUE #71: SPI operates in "endless transaction" mode
When a SPI transaction occurs, the HAL operates the SPI in "endless transaction" mode, which means that the SPI peripheral is not able to detect when a transaction completes.

The SPI_CR2.TSIZE register value needs to be updated with the size of the transaction in bytes before a transaction is initiated.

### LINKED ISSUE #158: SPI CS management does not work
It looks like I made a conversion mistake for managed-CS SPI transactions, and the config for managed CS doesn't function as intended.

Proper register configurations are show here:
https://github.com/quartiq/stm32h7xx-hal/blob/ff00e938f2b226211c178f26c092f36462c44404/src/spi.rs#L470-L473

It looks like we're missing SSOE bit:
https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/spi.rs#L491

ryan-summers:
This PR fixes #158 by correcting the usage of the SPI management bits.
This PR fixes #71 by working around "endless transaction mode" by allowing more control of the way CS is managed.

This PR also exposes methods to allow SPI to be used in tx-only and rx-only configurations. It also allows for CS to automatically be idled when no data is available in the FIFO for transmission.

### DIFF
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -199,12 +199,23 @@ impl Config {
     }
 
     /// CS pin is automatically managed by the SPI peripheral.
+    ///
+    /// # Note
+    /// SPI is configured in "endless transaction" mode, which means that the SPI CSn pin will
+    /// assert when the first data is sent and will not de-assert.
+    ///
+    /// If CSn should be de-asserted between each data transfer, use `suspend_when_inactive()` as
+    /// well.
     pub fn manage_cs(mut self) -> Self {
         self.managed_cs = true;
         self
     }
 
     /// Suspend a transaction automatically if data is not available in the FIFO.
+    ///
+    /// # Note
+    /// This will de-assert CSn when no data is available for transmission and hardware is managing
+    /// the CSn pin.
     pub fn suspend_when_inactive(mut self) -> Self {
         self.suspend_when_inactive = true;
         self
--- a/src/spi.rs
+++ b/src/spi.rs
@@ -484,7 +484,7 @@ macro_rules! spi {
                         spi.cr1.write(|w| w.ssi().slave_not_selected());
 
                         // Calculate the CS->transaction cycle delay bits.
-                        let cycle_delay: u8 = {
+                        let (start_cycle_delay, interdata_cycle_delay) = {
                             let mut delay: u32 = (config.cs_delay * spi_freq as f32) as u32;
 
                             // If the cs-delay is specified as non-zero, add 1 to the delay cycles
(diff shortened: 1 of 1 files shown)


richardeoin:
Looks good! I tried this out with the spi_dma example on the dma branch.

I somewhat expected that with `manage_cs()` the !CS pin would go high again after the transaction, but I also need `suspend_when_inactive()` for that, since the SPI HAL doesn't set CR1.CSUSP or use TSIZE so that SR.EOT is set automatically. It is worth adding a note about this to the `manage_cs` documentation string? Other than that this PR can be merged.

Potentially this is an area where the SPI HAL could be improved. Adding a `suspend()` method to `Spi<_, Enabled, _>` that sets CR1.CSUSP would be a useful companion to `manage_cs()` when not using the DMA, so that user code can end the transaction manually.

ryan-summers:
> I somewhat expected that with `manage_cs()` the !CS pin would go high again after the transaction, but I also need `suspend_when_inactive()` for that, since the SPI HAL doesn't set CR1.CSUSP or use TSIZE so that SR.EOT is set automatically. It is worth adding a note about this to the `manage_cs` documentation string? Other than that this PR can be merged.

This is an implication of operating SPI in "endless transaction mode" (e.g. TSIZE = 0). The SPI CS pin is held low for the duration of an entire transaction, which means that it is held low indefinitely when in endless transaction mode. I'll add documentation to describe what behavior should be expected.

Note that using TSIZE has some weird implications to the SPI FIFO. Notable, once TSIZE "data" have been written to the SPI FIFO, the TXTF is set, which causes all further TX FIFO writes to be ignored:

```
Any write to the TxFIFO is ignored when there is no sufficient room to store at least a single data frame (TXP event is
not respected), when TXTF is set or when the SPI is disabled
```

And the TXTF must be manually cleared by software, so we'd have to continually check + clear the flag in `send()`, which seems really redundant and not-that-useful.


> Potentially this is an area where the SPI HAL could be improved. Adding a `suspend()` method to `Spi<_, Enabled, _>` that sets CR1.CSUSP would be a useful companion to `manage_cs()` when not using the DMA, so that user code can end the transaction manually.

I think it's a good idea as well, although I think it's somewhat unintuitive to "suspend" the transaction - I think a user would just assume that the transaction is over when they are done sending data. "suspend" requires some understanding as to how we are configuring the underlying peripheral. It likely makes more sense to just automatically suspend the transaction after "sending" all of the data, although with how the `embedded-hal` implements `transfer`, it's not entirely clear how this would be best accomplished.

richardeoin (approved):

richardeoin:
Indeed the `embedded-hal` implementation of `Transfer` doesn't give us any opportunity to do anything when the transfer completes, which is a pain. So I think it does need to be a separate method.

Anyhow everything in this PR is good, thanks!

bors r+

PR #161, opened 2020-10-28
Updating QSPI to sample on falling edges

### LINKED ISSUE #160: QSPI reads unstable
Due to a likely silicon bug, the QSPI peripheral (reference https://github.com/quartiq/stabilizer/issues/101), QSPI reads may be incorrect when reading at the rising clock edge.

To work around this, the QSPI peripheral can be updated to read on the falling clock edge instead by setting the SSHIFT bit:
image

ryan-summers:
This PR fixes #160 by updating the QSPI to read on clock falling edges instead of rising edges. This is a workaround to signal contention discovered on the QSPI bus. 

Refer to https://github.com/quartiq/stabilizer/issues/101 for evidence of signal contention that prompted this change

### DIFF
--- a/src/qspi.rs
+++ b/src/qspi.rs
@@ -341,9 +341,13 @@ impl Qspi {
             _ => panic!("Invalid QSPI frequency requested"),
         };
 
-        // Write the prescaler
-        regs.cr
-            .write(|w| unsafe { w.prescaler().bits(divisor as u8) });
+        // Write the prescaler and the SSHIFT bit. Note that SSHIFT is required because it appears
+        // that the QSPI may have signal contention issues when reading. SSHIFT forces the read to
+        // occur on the falling edge instead of the rising edge. Refer to
+        // https://github.com/quartiq/stabilizer/issues/101 for more information
+        regs.cr.write(|w| unsafe {
+            w.prescaler().bits(divisor as u8).sshift().set_bit()
+        });
 
         match bank {
             Bank::One => regs.cr.modify(|_, w| w.fsel().clear_bit()),


irwineffect:
Should we make setting the SSHIFT bit configurable rather than hardcoded?

ryan-summers:
I don't think this should be configurable since reading doesn't work with out.

It's not listed in the [STM32H7 errata](https://www.st.com/resource/en/errata_sheet/dm00368411-stm32h742xig-and-stm32h743xig-device-limitations-stmicroelectronics.pdf), but the captured device behavior does not align with what the datasheet indicates:

image

[The datasheet](https://www.st.com/resource/en/reference_manual/dm00314099-stm32h742-stm32h743-753-and-stm32h750-value-line-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf) indicates that I/O changeover occurs on the falling clock edge, but image shows that it appears to actually change over on the rising edge.

richardeoin:
I checked that this isn't an issue when dummy cycles are used (they probably would be for a memory), which is possibly why ST haven't caught it/don't care. The H7 releases the IO line half a cycle into a 1 cycle dummy phase.

I've just updated some docs and then this is good

richardeoin:
Ah I can't push onto this branch, so I actually made a PR to it instead https://github.com/quartiq/stm32h7xx-hal/pull/1 

Feel free to cherry-pick / copy the commit if that's easier.

ryan-summers:
Thanks for the background investigation - I've merged your changes, so they should now be present here

PR #164, opened 2020-11-03
Updating DMA macro for SPI types

ryan-summers:
This PR updates the DMA macro for SPI types. When SPI is configured for various transaction sizes, the `TargetAddress` needs to have a different associated `MemSize` so that the SPI peripheral can be used by DMA.


richardeoin:
Ok, agreed that specialised macros for each peripheral are better than repeating all the variants each time (especially when there's 5 of them!)

PR #167, opened 2020-11-08
[DMA] Adding DMA+SPI+RTIC example

ryan-summers:
This PR adds in an `example/spi-dma-rtic.rs` example that demonstrates how to use RTIC, DMA transfers, and SPi together.
I also did a merge of `master` to gain access to the spi transmit-only functionality for the example.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -186,3 +186,7 @@ required-features = ["rm0433"]
 [[example]]
 name = "spi_dma"
 required-features = ["rm0433"]
+
+[[example]]
+name = "spi-dma-rtic"
+required-features = ["rm0433"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -189,4 +189,4 @@ required-features = ["rm0433"]
 
 [[example]]
 name = "spi-dma-rtic"
-required-features = ["rm0433"]
+required-features = ["rm0433","rt"]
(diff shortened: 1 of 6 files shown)


richardeoin:
Looks good, thanks! I'll try to keep the `dma` branch re-based onto master to avoid the noise in future, but it's OK for this.

I usually write `use stm32h7xx_hal::{dma, spi, gpio, ..};` to make the type signatures less verbose, but actually `hal::` is probably better for an educational example like this.

For me `cs` goes high about 40µs (~16 bytes) before the end of the transmission - I guess that's the SPI FIFO in action. Perhaps worth adding a comment about that.

Clippy has one reasonably valid point to clean up the example. After that this can be merged!

```
error: this import is redundant
  --> examples/spi-dma-rtic.rs:11:1
   |
11 | use cortex_m;
   | ^^^^^^^^^^^^^ help: remove it entirely
   |
note: the lint level is defined here
  --> examples/spi-dma-rtic.rs:5:9
   |
5  | #![deny(warnings)]
   |         ^^^^^^^^
   = note: `#[deny(clippy::single_component_path_imports)]` implied by `#[deny(warnings)]`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_component_path_imports
```

ryan-summers:
@richardeoin You're right about the SPI FIFO. When I was testing this, I only had 1 oscilloscope probe not hooked up to something else, so I didn't notice this. I have fixed it to busy wait for the FIFO transmission to complete. Also fixed the clippy comment :)

richardeoin (approved):

PR #168, opened 2020-11-10
Reduce smoltcp dependency features to a minimum

richardeoin:
Since [*Dependencies are resolved with the union of all features enabled on them*](https://doc.rust-lang.org/cargo/reference/resolver.html#features), it's good practice to remove unnecessary dependencies.

More features are added by the dev-dependency so that the examples build


hargoniX (approved):
If it compiles it works xD.

PR #170, opened 2020-11-15
Add DelayFromCountDownTimer for when SYST is unavailable

### LINKED ISSUE #169: Alternative Delay implementation
If I want to use scheduling in `rtic` the init function does not have access to the `SYST` timer as described here (end paragraph of first section): [https://rtic.rs/0.5/book/en/by-example/timer-queue.html](https://rtic.rs/0.5/book/en/by-example/timer-queue.html).

This means that there's no way to construct the current delay implementation which requires the `SYST` timer.

[This comment](https://github.com/rtic-rs/cortex-m-rtic/issues/117#issuecomment-692751736) on the issue over at cortex-m-rtic suggest that HAL crates provide alternate Delay implementations to support this use case, potentially using `cortex_m::asm::delay`.

I tested this a bit locally but a delay based on `cortex_m::asm::delay` seemed pretty unstable and delayed for less time than I had calculated. Later I found this issue seem to explain the problem: [https://github.com/rust-embedded/cortex-m/issues/236](https://github.com/rust-embedded/cortex-m/issues/236)
The gist of it being that the current implementation is only accurate on M3/M4.

As far as I understand though, `Delay` in this crate is only really used for startup configuration where one has to wait _at least_ some amount of time, so it could still be useful to have an implementation based on `cortex_m::asm::delay` if it's augmented to account for the inaccuracy of the current implementation, and properly documented to inform the user.

If this seems reasonable I could provide a PR for adding this.

glindstedt:
Fixes #169

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -14,7 +14,6 @@ use crate::stm32::{ADC1, ADC2};
 #[cfg(not(feature = "rm0455"))]
 use crate::stm32::{ADC3, ADC3_COMMON};
 
-use crate::delay::Delay;
 use crate::gpio::gpioa::{PA0, PA1, PA2, PA3, PA4, PA5, PA6, PA7};
 use crate::gpio::gpiob::{PB0, PB1};
 use crate::gpio::gpioc::{PC0, PC1, PC2, PC3, PC4, PC5};
@@ -304,7 +303,7 @@ pub trait AdcExt<ADC>: Sized {
 
     fn adc(
         self,
-        delay: &mut Delay,
+        delay: &mut impl DelayUs<u8>,
         prec: Self::Rec,
         clocks: &CoreClocks,
     ) -> Adc<ADC, Disabled>;
@@ -344,7 +343,7 @@ fn check_clock(prec: &impl AdcClkSelGetter, clocks: &CoreClocks) -> Hertz {
 pub fn adc12(
     adc1: ADC1,
     adc2: ADC2,
-    delay: &mut Delay,
+    delay: &mut impl DelayUs<u8>,
     prec: rec::Adc12,
     clocks: &CoreClocks,
 ) -> (Adc<ADC1, Disabled>, Adc<ADC2, Disabled>) {
@@ -403,7 +402,7 @@ macro_rules! adc_hal {
                 type Rec = rec::$Rec;
 
 	            fn adc(self,
-                       delay: &mut Delay,
+                       delay: &mut impl DelayUs<u8>,
                        prec: rec::$Rec,
                        clocks: &CoreClocks) -> Adc<$ADC, Disabled>
 	            {
@@ -416,7 +415,7 @@ macro_rules! adc_hal {
                 ///
                 /// Sets all configurable parameters to one-shot defaults,
                 /// performs a boot-time calibration.
-                pub fn $adcX(adc: $ADC, delay: &mut Delay,
+                pub fn $adcX(adc: $ADC, delay: &mut impl DelayUs<u8>,
                              prec: rec::$Rec, clocks: &CoreClocks
                 ) -> Self {
                     // Consume ADC register block, produce Self with default
@@ -455,7 +454,7 @@ macro_rules! adc_hal {
                 /// Disables Deeppowerdown-mode and enables voltage regulator
                 ///
                 /// Note: After power-up, a [`calibration`](#method.calibrate) shall be run
-                pub fn power_up(&mut self, delay: &mut Delay) {
+                pub fn power_up(&mut self, delay: &mut impl DelayUs<u8>) {
                     // Refer to RM0433 Rev 6 - Chapter 24.4.6
                     self.rb.cr.modify(|_, w|
                         w.deeppwd().clear_bit()
--- a/src/delay.rs
+++ b/src/delay.rs
@@ -1,4 +1,38 @@
-//! Delays
+//! Delay providers
+//!
+//! There are currently two delay providers. In general you should prefer to use
+//! [Delay](Delay), however if you do not have access to `SYST` you can use
+//! [DelayFromCountDownTimer](DelayFromCountDownTimer) with any timer that
+//! implements the [CountDown](embedded_hal::timer::CountDown) trait. This can be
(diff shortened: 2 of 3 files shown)


richardeoin:
Thanks! Would you consider adding some documentation and/or an example for this? I think it's quite difficult to discover `DelayFromTimer` exists, either adding some documentation at the top of `delay.rs` or an example would really help. We don't currently have an example for the RTIC scheduler, this could be a great excuse to add that.

This abstraction breaks down for durations longer than a second (think about what happens for `delay.delay_ms(2_000)`). I don't think that's stopping us merging this, but it would be an improvement to fix it. Since I wrote that snippet we now have the [`set_timeout`](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/timer.rs#L336) method on the timers, if that method was part of a trait it could be used here. Something to experiment with if you're interested!

glindstedt:
Ah right... I didn't look too closely at the implementation. I'll have a look at `set_timeout` and see where that takes me. I'll definitely add some documentation, good point.

glindstedt:
@richardeoin I pushed a new commit and would like some quick feedback (still WIP):
I extracted some common timer configuration methods into a trait `TimerConf` (maybe `TimerControl` is better, Conf doesn't really make sense with some methods like `pause()`), which allows for a delay implementation using the `set_timeout` method. I tested with a delay of 5s and seems to work as expected.

Should I try to grab all the methods currently defined on Timer and put it in the new trait? Or should I prioritize to see if it can be generalized over `LpTimer` as well? I figure the trait could be more useful then.

Does it make sense to keep around an implementation based on `CountDown`? If the new trait can't be used for an `LpTimer` it might still be useful.

Also I spotted now that `clear_irq()` and `clear_uif_bit()` methods are identical, I can fix that.

richardeoin:
Ok thanks for this. I understand the approach of putting enough functionality in the `TimerConf` trait implementation and then making `DelayFromTimer` generic over `T: TimerConf`.

My first comment would be that if `TimerConf` is part of the external API, it should abstract as much hardware-specific detail as possible. So I would go in the direction of prioritizing functionality and making it generic over `LpTimer` as well. However `LpTimer` currently doesn't have any methods for delays > 1s (there's no `set_timeout` method) so you would need to add that.

Another thing to note is that `CountDown` comes from the embedded-hal crate, and therefore is important for writing device drivers that are portable between HAL crates. I realise that in practice the `DelayMs`/`DelayUs` traits are in much more commonly used, but I think we should keep implementing as much of embedded-hal as possible - including `CountDown`.

Would another approach be to write `DelayFromCountDownTimer` so that it loops over the `block!` calls if t > 1s? That would allow it to use the existing `CountDown<Time = Hertz>` interface.

glindstedt:
Ok great idea about looping over `block!`, got it working for both `Timer` and `LpTimer`. There was one hitch though, for some configurations the maximum timeout for `LpTimer` is <1s, so I couldn't just loop on a `1.hz()` timeout. In my case where I was testing it the clock used was 120Mhz (PCLK1 I think), and I've set the looping timeout to work with this. I'm not sure what the maximum clock configuration is though as the manual is quite dense, do you happen to know?

It would be great if the `CountDownTimer` trait had some way to signal the maximum supported value for the `start()` method.

This implementation solves the original issue, so I could remove the other implementation as this is more generic now since it supports `LpTimer` as well.

There might be some value to gain from a shared trait, I can see if I can clean that up a bit and what be made generic over `LpTimer`.

richardeoin:
Cool, nice that you got it to work! The arithmetic looks good to me, could you just add an if statement to protect against division by zero when `time_left` = 0?

120MHz is the maximum PCLK for most parts (the HAL has asserts [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/rcc/mod.rs#L693)), but it is possible to clock timers at up to 480MHz depending on the part. In the case of the LPTIMs this is through the kernel clock multiplexers (eg. RM0399 Fig. 65). So maybe you can just increase that divisor to 60?

> It would be great if the CountDownTimer trait had some way to signal the maximum supported value for the start() method.

Traits can have associated constants, so that would be an obvious solution. You could propose adding it to `CountDown` in [embedded-hal](https://github.com/rust-embedded/embedded-hal/issues?q=CountDown). It would be a non-breaking change to embedded-hal if the trait sets a default value.

> This implementation solves the original issue, so I could remove the other implementation as this is more generic now since it supports LpTimer as well.

Sure. If there's a shared trait that's useful, that would be great to have. Otherwise `DelayFromCountDownTimer` is also useful on its own.

glindstedt:
I think extracting into a generic trait might warrant more careful consideration, especially given the Enabled/Disabled semantic of `LpTimer`, so I backtracked on that. I did leave some changes where the `CountDown` impl dogfoods some of the public methods on `Timer` though.

I have one more question though since I'm already touching these things: The pre-existing methods [clear_uif_bit()](https://github.com/stm32-rs/stm32h7xx-hal/pull/170/files#diff-abbd7706dff206239bd03de70e864a2893146421b4877638f17149848491fcd0R395) and [clear_irq()](https://github.com/stm32-rs/stm32h7xx-hal/pull/170/files#diff-abbd7706dff206239bd03de70e864a2893146421b4877638f17149848491fcd0R445) on `Timer` are identical. Should I remove one of them to avoid confusion? `LpTimer` also has a `clear_irq()` method but uses a different bit. I guess `clear_irq()` would make sense for a generic trait over both types, but for now maybe `clear_uif_bit()` is more transparent.

richardeoin:
Yes, I think it would be clearer if one of them was removed. It doesn't terribly matter which, but I would have a small preference for keeping `clear_irq`. `UIF` is the only interrupt flag that is common among all timers afaict.

glindstedt:
Alright, I removed the `clear_uif_bit()` method. I squashed the commits into two, one with the `DelayFromCountDown` implementation and the other with the Timer refactorings. If you think it looks ok I'd be ready for merging.

glindstedt:
Hm, there's still the question of documentation for purpose of discoverability, I can see if I can add a short example at the top of `delay.rs`

richardeoin (approved):

richardeoin:
Ok! Thanks for the example too

bors r+

PR #171, opened 2020-12-01
DMA buffer swap logic

jordens:
This PR (into the `dma` branch) rewrites the `next_transfer()` and `next_transfer_with()` implementations.

* `CurrentBuffer` enums renamed to be more accurate and peripheral-matching.
* Added a few stream methods to expose the dbm/ct bits in joint semantics (also faster).
* Using `usize` for the addresses instead of `u32`.
* The non-double buffer and the double buffer code paths are unified.
* It poisons the inactive buffer address to convert silent corruption into DMA/bus errors (@adamgreig 's idea).
* With that, I'd consider `next_transfer_with()` `safe`. It's not DMA overrun safe in general but it is safe from any concurrent buffer access by user and DMA since the buffer the user has access to is always out of reach of the DMA (by poisoning or implicit disable). Certainly safe in the sense that there is nothing that the user can do to make it safer or less safe.
* Overrun errors due to due to DMA hitting the poison or writing to the active address will not go unnoticed (user needs to clear, handle, and restart).
* Errors due to `next_transfer{,_with}` being invoked late (DMA wrap around) can still go unnoticed. I don't think there is anything we can do about it since the peripheral will happily wrap around unattended (AFAWK). But they now are guaranteed to not lead to corruption within a buffer.
* `next_transfer()` just calls `next_transfer_with()` with triple buffering (one active, one inactive, and one user).
* `fence()` added.

Pending:

* [x] This is untested on hardware.
* [x] `BDMA` is not done.
* [ ] Returning a/the buffer on error might make sense.

@thalesfragoso @richardeoin @ryan-summers comments appreciated.

### DIFF
--- a/src/dma/mod.rs
+++ b/src/dma/mod.rs
@@ -490,7 +490,7 @@ where
                 // between reading the CT bit and poisoning the inactive address, this
                 // write will fail and lead to a transfer error (TEIF) and disable
                 // the stream.
-                // If DMA wins the race by the time we write the new valid addressi
+                // If DMA wins the race by the time we write the new valid address
                 // (below), it gets a bus error and errors/stops.
                 unsafe {
                     self.stream.set_memory_address(inactive, 0xffff_ffffusize);
--- a/src/dma/mod.rs
+++ b/src/dma/mod.rs
@@ -32,7 +32,7 @@ mod macros;
 #[cfg(not(feature = "rm0455"))] // Remove when fixed upstream
 pub mod dma; // DMA1 and DMA2
 
-// pub mod bdma;
+pub mod bdma;
 
 pub mod traits;
 use traits::{
--- a/src/dma/mod.rs
+++ b/src/dma/mod.rs
@@ -177,19 +177,19 @@ impl From<u8> for FifoLevel {
 #[derive(Debug, Clone, Copy, PartialEq)]
 pub enum CurrentBuffer {
     /// The first buffer (m0ar) is in use.
-    FirstBuffer,
+    Buffer0,
     /// The second buffer (m1ar) is in use.
-    DoubleBuffer,
+    Buffer1,
 }
 
 impl Not for CurrentBuffer {
     type Output = CurrentBuffer;
 
     fn not(self) -> Self::Output {
-        if self == CurrentBuffer::FirstBuffer {
-            CurrentBuffer::DoubleBuffer
+        if self == CurrentBuffer::Buffer0 {
+            CurrentBuffer::Buffer1
         } else {
-            CurrentBuffer::FirstBuffer
+            CurrentBuffer::Buffer0
         }
     }
 }
@@ -363,7 +363,7 @@ where
         //
         // Must be a valid memory address
         unsafe {
-            stream.set_memory_address(buf_ptr as u32);
+            stream.set_memory_address(CurrentBuffer::Buffer0, buf_ptr as u32);
         }
 
         let is_mem2mem = DIR::direction() == DmaDirection::MemoryToMemory;
@@ -395,7 +395,7 @@ where
                     // Double buffer is the source in mem2mem mode
                     stream.set_peripheral_address(db_ptr as u32);
                 } else {
-                    stream.set_memory_double_buffer_address(db_ptr as u32);
+                    stream.set_memory_address(CurrentBuffer::Buffer1, db_ptr as u32);
                 }
             }
             Some(db_len)
@@ -498,13 +498,13 @@ where
                 return Err(DMAError::SmallBuffer(new_buf));
             }
 
-            if STREAM::current_buffer() == CurrentBuffer::DoubleBuffer {
+            if STREAM::current_buffer() == CurrentBuffer::Buffer1 {
                 unsafe {
-                    self.stream.set_memory_address(new_buf_ptr as u32);
+                    self.stream.set_memory_address(CurrentBuffer::Buffer0, new_buf_ptr as u32);
                 }
                 // Check if an overrun occurred, the buffer address won't be
                 // updated in that case
-                if self.stream.get_memory_address() != new_buf_ptr as u32 {
+                if self.stream.get_memory_address(CurrentBuffer::Buffer0) != new_buf_ptr as u32 {
                     self.stream.clear_transfer_complete_interrupt();
                     return Err(DMAError::Overrun(new_buf));
                 }
@@ -516,15 +516,15 @@ where
                 let old_buf = self.buf.replace(new_buf);
 
                 // We always have a buffer, so unwrap can't fail
-                return Ok((old_buf.unwrap(), CurrentBuffer::FirstBuffer));
+                return Ok((old_buf.unwrap(), CurrentBuffer::Buffer0));
             } else {
                 unsafe {
                     self.stream
-                        .set_memory_double_buffer_address(new_buf_ptr as u32);
+                        .set_memory_address(CurrentBuffer::Buffer1, new_buf_ptr as u32);
                 }
                 // Check if an overrun occurred, the buffer address won't be
                 // updated in that case
-                if self.stream.get_memory_double_buffer_address()
+                if self.stream.get_memory_address(CurrentBuffer::Buffer1)
                     != new_buf_ptr as u32
                 {
                     self.stream.clear_transfer_complete_interrupt();
@@ -538,7 +538,7 @@ where
                 let old_buf = self.double_buf.replace(new_buf);
 
                 // double buffering, unwrap can never fail
-                return Ok((old_buf.unwrap(), CurrentBuffer::DoubleBuffer));
+                return Ok((old_buf.unwrap(), CurrentBuffer::Buffer1));
             }
         }
         self.stream.disable();
@@ -552,7 +552,7 @@ where
         let buf_len = unsafe {
             let (buf_ptr, buf_len) = new_buf.write_buffer();
 
-            self.stream.set_memory_address(buf_ptr as u32);
+            self.stream.set_memory_address(CurrentBuffer::Buffer0, buf_ptr as u32);
             buf_len
         };
         self.stream.set_number_of_transfers(buf_len as u16);
@@ -573,7 +573,7 @@ where
             self.stream.enable();
         }
 
-        Ok((old_buf.unwrap(), CurrentBuffer::FirstBuffer))
+        Ok((old_buf.unwrap(), CurrentBuffer::Buffer0))
     }
 
     /// Stops the stream and returns the underlying resources.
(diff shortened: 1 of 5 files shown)


thalesfragoso:
I haven't looked at all the PR yet, but one thing you could change is to use `fence` instead of `compiler_fence` + `dmb`, that way you avoid a function call (`cortex_m::asm::dmb` is never inlined on stable).

jordens:
The cortex-m assembly is being inlined even on stable for a while now.

thalesfragoso:
> The cortex-m assembly is being inlined even on stable for a while now.

I don't think so, as far as I know, inlined assembly calls in stable was only made possible recently in cortex-m 0.7.0 and only when using `linker-plugin-lto`. As of today, I don't think there is a single PAC/HAL using cortex-m 0.7, which means that people can't really use 0.7 together with those crates, also, the use of `linker-plugin-lto` isn't widespread. (See https://github.com/rust-embedded/cortex-m/pull/259 and https://github.com/rust-embedded/cortex-m/issues/139)

And btw, this is the exactly use case of `fence`, it will insert a `DMB` instruction.

src/dma/mod.rs
```diff
@@ -465,115 +448,128 @@ where
...
-
-                let old_buf = self.buf.replace(new_buf);
-
-                // We always have a buffer, so unwrap can't fail
-                return Ok((old_buf.unwrap(), CurrentBuffer::FirstBuffer));
-            } else {
+                // Poison the peripheral's inactive memory address to get a memory
+                // error instead of potentially silent corruption if the DMA peripheral
+                // wins the race to the inactive buffer.
+                // NOTE(safety): Memory corruption will still occur in the "incactive"
```
ryan-summers, line 494: nit: s/incactive/inactive

src/dma/mod.rs
```diff
@@ -465,115 +448,128 @@ where
...
-                        .set_memory_double_buffer_address(new_buf_ptr as u32);
-                }
-                // Check if an overrun occurred, the buffer address won't be
-                // updated in that case
-                if self.stream.get_memory_double_buffer_address()
-                    != new_buf_ptr as u32
-                {
-                    self.stream.clear_transfer_complete_interrupt();
-                    return Err(DMAError::Overrun(new_buf));
+                    self.stream.set_memory_address(inactive, 0xffff_ffffusize);
```
ryan-summers, line 498: It occurs to me that there's still a race condition here. If the stream swaps over directly after we check the active stream and before we update the address, the address write will fail (because the inactive buffer has become active).

A potential fix for this is to treat it like reading a two register timer - that is, we detect the inactive buffer and then attempt to "acquire" it by overwriting the buffer address. Then, we can check to see if the poisoned address sticks (write will fail if the buffer becomes active). We repeat that process until the poisoned address is accepted. Once we've managed to acquire one of the two buffers (by poisoning the associated address), we can continue with execution with a guarantee that there was no race condition.

I think it's dangerous to leave a race condition like this present in the HAL code otherwise.

src/dma/mod.rs
```diff
@@ -465,115 +448,128 @@ where
...
-        }
-        self.stream.disable();
-        self.stream.clear_transfer_complete_interrupt();
+        };
 
-        // "No re-ordering of reads and writes across this point is allowed"
+        // Protect the instruction sequence of preceding DMA disable/inactivity
+        // verification/poisoning and subsequent (old completed) buffer content
+        // access.
         compiler_fence(Ordering::SeqCst);
```
ryan-summers, line 507: As @thalesfragoso has noted, it seems like just [`core::sync::atomic::fence(Ordering::SeqCst)`](https://doc.rust-lang.org/core/sync/atomic/fn.fence.html) is appropriate here to reduce the need to invoke two separate fences. This is true below as well wherever we have these fences

src/dma/mod.rs
```diff
@@ -465,115 +448,128 @@ where
...
+        if !double {
+            self.stream.set_number_of_transfers(buf_len as u16);
+        } else if buf_len != usize::from(self.transfer_length) {
+            // We can't change the transfer length while double buffering
+            return Err(DMAError::SmallBuffer);
+        }
 
         unsafe {
-            self.stream.enable();
+            self.stream.set_memory_address(inactive, buf_ptr as usize);
```
ryan-summers, line 540: Would you mind adding a comment here that we can ignore failures to write this address in double buffer mode specifically because of the address poisoning that we invoked earlier? That is, _if_ this buffer has already become active and a write would fail, the DMA has gone into an error state and the stream has disabled anyways, so this write is guaranteed to pass.

Adding this note will likely help us in the future determine why this is acceptable (and why it may not be if we refactor)

ryan-summers (approved):
In general, I think these are nice improvements and increase maintainability. There's still a race condition that I would like to resolve though. Comments inline

ryan-summers (changes requested):
Inadvertently approved when I requested changes

src/dma/mod.rs
jordens: It's already commented and addressed. The poison write will lead to TEIF and disable the stream. There is no (silent, leading to corruption of an active buffer) race.
Even the case where DMA does several buffer swaps in the time we get to writing the poison (a few instructions) is the same as the case where we just handle the TCIF late.

src/dma/mod.rs
jordens: It's not like the write failing or passing in that case matters. In fact it always passes from the CPU perspective.

jordens:
> I don't think so, as far as I know, inlined assembly calls in stable was only made possible recently in cortex-m 0.7.0 and only when using `linker-plugin-lto`.

Ack. It felt to me like it has been there for much longer. Changed to `fence()`.

src/dma/mod.rs
ryan-summers: There's technically still a case here where the DMA interrupt occurs some multiple of 2 times between getting the inactive buffer and setting the memory address. This might occur if an ISR happens between these two functions.

The result would be that a buffer would be overwritten (potentially multiple times) before the swap occurs.

jordens: That's what I just said in my the last sentence. But it's not a separate problem from the general case (which should be now described and documented in abundant detail).
Or am I misunderstanding you?

ryan-summers: Yeah, I agree with you here. Closing this as resolved - I don't think we can justifiably worry about that at this level.

src/dma/mod.rs
```diff
@@ -41,28 +41,12 @@ use traits::{
 };
 
 /// Errors.
-#[derive(PartialEq)]
-pub enum DMAError<T> {
+#[derive(PartialEq, Debug, Copy, Clone)]
+pub enum DMAError {
     /// DMA not ready to change buffers.
-    NotReady(T),
+    NotReady,
```
ryan-summers, line 47: Instead of adding a `NotReady`, can we use an `nb::Result::WouldBlock`? I think it's essentially being used in the same way - a user needs to wait for the transfer to finish first.

src/dma/mod.rs
```diff
@@ -465,133 +441,159 @@ where
...
+        };
 
-                // "Subsequent reads and writes cannot be moved ahead of
-                // preceding reads"
-                compiler_fence(Ordering::Acquire);
+        // Protect the instruction sequence of preceding DMA disable/inactivity
+        // verification/poisoning and subsequent (old completed) buffer content
+        // access.
+        // Cortex-M7: Also protect the corresponding data access sequence.
+        // NOTE: The data cache also needs to be flushed (if enabled).
```
ryan-summers, line 506: It seems to me like this is something we should be doing for all users, since the DMA memory is guaranteed to be in a cache-able memory location. We should do an MVA invalidate for all addresses in the buffer if the d-cache is enabled, as the comment here indicates.

Note that there seems to be some issue here because `cortex-m` requires ownership of the SCB for `invalidate_dcache_by_address()` and cache lines must be 32-byte aligned, which means that DMA buffers must be cache-aligned when the d-cache is enabled as well.

Now that I've written all of this, it may be worth breaking out into a separate PR to add support for d-cache-enabled implementations.

https://docs.rs/cortex-m/0.7.0/cortex_m/peripheral/struct.SCB.html#method.invalidate_dcache_by_address

ryan-summers: It looks like that would also restrict us to using buffers that are a length that is a multiple of the cache size.

ryan-summers: Note that there's a lot of other cache operations that we aren't supporting (such as cleaning the active buffer to commit it to memory before starting a DMA write), so this isn't necessarily the only place where cache operation support needs to be added. Let's leave this for the future though and add a disclaimer that DMA does not support the D-cache being enabled.

ryan-summers (commented):
I think one addition to an `nb::Result` would be appropriate - other than that, we should likely test this on hardware, and pending that, it looks good to me otherwise.

src/dma/mod.rs
jordens: I think that's useful. But this error should also carry the unused buffer (otherwise that's lost). The alternative would be to have another closure that gets the unused buffer in case of error, making `transfer_next_with()` somewhat like `Result::map_or_else()`.

ryan-summers: That's a good point, but this PR doesn't support returning the unused buffer on failure anyways

jordens: That regression is on the TODO list above.

src/dma/traits.rs
```diff
@@ -102,19 +102,17 @@ pub trait Stream: Sealed {
 /// Trait for Double-Buffered DMA streams
 pub trait DoubleBufferedStream: Stream + Sealed {
     /// Set the peripheral address (par) for the DMA stream.
-    unsafe fn set_peripheral_address(&mut self, value: u32);
+    unsafe fn set_peripheral_address(&mut self, value: usize);
 
     /// Set the memory address (m0ar) for the DMA stream.
```
richardeoin, line 107: Technically this comment is out-of-date now.

"Set the memory address pointer for the DMA stream. Can be either m0ar or m1ar"

src/dma/traits.rs
```diff
@@ -102,19 +102,17 @@ pub trait Stream: Sealed {
...
-    unsafe fn set_memory_address(&mut self, value: u32);
+    unsafe fn set_memory_address(
+        &mut self,
+        buffer: CurrentBuffer,
+        value: usize,
+    );
 
     /// Get the memory address (m0ar) for the DMA stream.
```
richardeoin, line 114: Ditto

"Get the memory address pointer for the DMA stream. Can be either m0ar or m1ar"

richardeoin (commented):
Looks good to me, with the caveat that I haven't tried it on hardware either. Thanks for the improvements `u32` -> `usize` and `fence`.

I'd say this can be merged into the `dma` branch as soon as you're happy with it, since I see the whole `dma` branch as experimental/unstable.

ryan-summers:
I have tested this on hardware and verified that it functions as intended.

richardeoin:
I've also successfully tested this on hardware. Let's continue the discussion on https://github.com/stm32-rs/stm32h7xx-hal/pull/153 if there's more to add

Issue #148, opened 2020-10-05
Implement tick_timer() for lptimers


richardeoin:
Ref https://github.com/stm32-rs/stm32h7xx-hal/pull/114

The `tick_timer` method is not implemented for LPTIM[1-5]

*Is there some limitation of the LPTIMs that makes it difficult?*

The LPTIMs require a specific initialisation sequence (for example, CFGR must be set before enabling the timer, ARR must be set after), so it's not difficult but rather a little tedious to check against the RM that it's correct.

noppej:
@richardeoin 
I needed the `tick_timer` method, so I have a push available with this code.   There are a couple of differences in how LPTIM'ers work relative to the other TIM'ers, but nothing super complex. 

My code includes a couple of (proposed) fixes such as 
- fn counter() sometimes gave incomprehensible results, so I changed the  
`self.tim.cnt.read()bits()`  
to  
`self.tim.cnt.read().cnt().bits().into()`

richardeoin:
Hi @noppej That's great, proposed improvements and fixes are very welcome! Normally we do this by Pull Requests, so if you make a Pull Request with your code I would review it.

noppej:
Hi @richardeoin ... I will gladly create a pull request. Before I do, I'd like to get your view on something, so that I can fix that at the same time. 

Currently, for regular (non low power) timers, the `fn set_tick_freq<T>(&mut self, frequency: T)`, sets a reload value in the ARR register. This function, is also called from `fn tick_timer<T>(self, frequency: T, prec: Self::Rec, clocks: &CoreClocks)` .  
Is it just me, or should the tick_timer set the frequency (PSC) ONLY, and NOT set the ARR?

richardeoin:
I think `set_tick_freq` does set the ARR to 0xFFFF for 16-bit timers, and 0xFFFF_FFFF for 32-bit timers. That's those two lines https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/timer.rs#L376-L377

The documentation on the `tick_timer` method in its trait definition says that it "Counts from 0 to the counter's maximum value". That sentence could be made more precise (does it include the maximum value or not?), but setting ARR to the maximum value seem reasonable based on that.

PR #172, opened 2020-12-06
Implement tick_timer() for LPTIM'ers

noppej:
Proposed implementation for issue #148

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 * pwr: Add backup domain voltage regulator control [#303][303]
 * MSRV increased to 1.56.1
+* timer: Implemented tick_timer method for LP Timers
+* timer: `counter()` method on LP Timers returns u16
+* timer: Fix bug in `counter()` TIM1/8/12/13/14 where return value would also
+  include UIF bit
 
 ## [v0.11.0] 2021-12-18
 
--- a/examples/tick_lptimer.rs
+++ b/examples/tick_lptimer.rs
@@ -1,0 +1,94 @@
+//! Demonstrates a microsecond-scale 64-bit timestamp counter using the LPTIM2
+//! timer
+
+#![no_main]
+#![no_std]
+
+use core::{
+    cell::RefCell,
+    sync::atomic::{AtomicU32, Ordering},
+};
+
+use cortex_m::{asm, interrupt::Mutex};
+use cortex_m_rt::entry;
+use log::info;
+
+use pac::interrupt;
+use stm32h7xx_hal::{pac, prelude::*, timer};
+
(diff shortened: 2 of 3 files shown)


noppej:
@richardeoin ... As discussed, can you please review and comment?  Tx :)

richardeoin:
Thanks for fixing the counter register read! That was a real bug for TIM1/8/12/13/14 (but I think not the other timers, or the LPTIMs), as the UIF bit was being read alongside the counter value. In any case explicitly specifying the field is better.

For your implementation of LPTIM `tick_timer`, I don't think it does what I expect `tick_timer` to do. `LpTimer::$timx` calls `start` and that calls `priv_set_freq`, which configures the timer to *overflow* at the given frequency, not count up at that frequency. 

I think it needs to be more like
```rust
let ticks = clk / frequency;
assert!(ticks < 128);

// Calculate prescaler
let (prescale, prescale_div) = match ticks {
.
.
```
(note there's no divide by `1 << 16`)

But otherwise accessing the ARR register to "fix" it is OK! I think this code does work for the special case where the frequency > `lptim_ker_ck` (clk), because in that case `ticks == 0` therefore `ticks == ticks / (1 << 16)`.

Perhaps you just need to multiply the frequency by `(1 << 16)`?

noppej:
@richardeoin You make some good points above. In my original implementation, I felt a bit 'burdened' by the fact that `LpTimer::$timx` does things that needs to be 'undone' for `tick_timer` but I will try to clean it up as much as possible. In the meantime, I have another related question from looking at the `tick_timer` implementation for the regular TIMs. 

The code currently ( [timer.rs#L272-L273](https://github.com/stm32-rs/stm32h7xx-hal/blob/9426c3fa6d10e847bd739843322f07ef31d21071/src/timer.rs#L272-L273) ) enables a UEV Event on timer overflow.  The documentation doesn't say anything about enabling events. My question to you ... should I enable that event for LPTIM implementation also, or ???

ryan-summers:
In my humble opinion, I think the timer implementation needs to be re-thought-out. Right now, it's hard to use timers provided by the HAL for precision control applications because it allows either an unknown tick value or an unknown tick frequency. I think it would also be desirable to add a timer API to specify tick rate + count so that it's easy to specify the timer configuration precisely when users need to, but that's likely something that I should break into a separate issue/PR.

noppej:
@ryan-summers ... I agree. I've resorted to register level code to achieve my use cases. It does seem to be an API that is shared across multiple stm32 hal's and your suggestion will likely result in breaking changes. I am new to this community, so not sure where to start the conversation. Any suggestions?

richardeoin:

PR #173, opened 2020-12-07
[DMA] Next transfer returns number of remaining data

ryan-summers:
This PR updates the DMA `next_transfer` to return the number of data that was not transferred. This is necessary for applications that are cancelled prematurely (e.g. in scatter-gather implementations where we gather as many data as possible in a fixed amount of time).

### DIFF
--- a/src/dma/mod.rs
+++ b/src/dma/mod.rs
@@ -481,7 +481,7 @@ where
     pub fn next_transfer(
         &mut self,
         mut new_buf: BUF,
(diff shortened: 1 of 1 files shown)


richardeoin (approved):

src/dma/mod.rs
```diff
@@ -481,7 +481,7 @@ where
     pub fn next_transfer(
```
ryan-summers, line 481: Docs need to be updated to indicate what the third returned value represents.

PR #175, opened 2021-01-02
Added advanced PWM functionality

mlamoore:
* TIMx.pwm_advanced builder pattern for configuring advanced options
* Original TIMx.pwm() code still works with advanced features disabled
* Added pwm_advanced example in examples directory
* Supports complementary PWM (on supporting timers)
* Supports left aligned, right aligned, or center aligned PWM (on supporting timers)
* Supports setting PWM frequency or prescaler/period
* Supports deadtime (on supporting timers)
* Supports break input (on supporting timers)
* Supports active-high or active-low PWM


richardeoin:
This is great! I tried the example out on a NUCLEO-H7A3ZI-Q and it works perfectly as far as I can see (compiled with `cargo build --example pwm_advanced --features stm32h7a3,rt,example-smps`)

I have two small bits of feedback after a (brief) look over the code, but neither of them are big problems.

* Often when using Builder methods on hardware, the mutator methods only change fields in some configuration struct. Only when the `finalize` method is called is the configuration applied to hardware (this is what we do for SPI, and in the QSPI PR I just submitted). I wouldn't say setting registers as you go along is "wrong", but leaving the configuration until the end can be more robust against the fields being supplied in an unexpected order.

It's an extreme case, but I would've expected the following to work:
```rust
    let pwm_builder = dp
        .TIM1
        .pwm_advanced(pins, ccdr.peripheral.TIM1, &ccdr.clocks)
        .center_aligned(); // Most hardware is centre-aligned

    // Left aligned output on this hardware variant
    #[cfg(feature = "special-hw-feature-flag")]
    let pwm_builder = pwm_builder.left_aligned();

    let (mut t1control, (t1c1, t1c2, t1c3, t1c4)) = pwm_builder.finalize();
```

That case can easily be fixed without moving everything into `finalize`, but it's an example of the edge-cases that are tricky to cover.

* In the RCC module at least, we've tried to stick to the rule that the "real" frequency would be less than or equal to the value specified. The truncation from a simple division does the opposite. I'm not sure what the right choice would usually be for motor control (probably a `base_freq` much higher than the PWM frequency, so the error is minimal), but it's worth mentioning the rounding error in a comment.

src/pwm.rs
```diff
@@ -644,45 +831,286 @@ macro_rules! tim_hal {
...
+                            .assume_init()
+                    }
+                }
+
+                /// Set the PWM frequency; will overwrite the previous prescaler and period
+                pub fn frequency<T: Into<Hertz>>(self, freq: T) -> Self {
+                    let tim = unsafe { &*$TIMX::ptr() };
+
+                    let freq: Hertz = freq.into();
+                    let reload: u32 = self.base_freq.0 / freq.0; // u32
```
richardeoin, line 899: For example: 1000 / 92 = 10; -> 100Hz frequency

You can make the it less than or equal to the frequency specified with this

```let reload: u32 = (self.base_freq.0 + freq.0 - 1) / freq.0; // u32```

Then: 1091 / 92 = 11; -> 90.9Hz frequency

Comment (mlamoore) [reply src/pwm.rs:899]: Is there a clean way to use a builder pattern with typestate programming while doing all the configuration in the finalize method?

I would like it if I could run some conditional code (if or match) on the type of some of the type states; if we're left-aligned, use one bit pattern, otherwise if we're center-aligned, use a different bit pattern and adjust the period to match. But I'd rather not define a trait for each parameter of the struct and implement those traits separately for each value of each type parameter.

I was going to say there isn't much inter-dependence between the sub-parameters, but I just realized that .frequency().center_aligned() will act differently than .center_aligned().frequency(), so at least the alignment really should be handled in the finalize method. Is there a better way to do it than just adding a new trait for alignment that sets those bits properly?

For rounding, usually the error will be a fraction of a percent, but if you're doing a high frequency digital power topology and using a low base clock to reduce power consumption, I would want it to round to nearest instead of up or down:

```
let reload: u32 = (self.base_freq.0 + freq.0 >> 1) / freq.0; // u32
```

If you don't have any examples of running code differently inside a function based on type parameters, I'll just add a new trait for alignment to handle that at finalize time similar to the PwmPinEnable trait I added to handle enabling/disabling channels properly based on the complementary state. I feel like period and prescaler can still be handled the way they are now, at least until const generics is stabilized.

Comment (richardeoin) [reply src/pwm.rs:899]: >Is there a clean way to use a builder pattern with typestate programming while doing all the configuration in the finalize method?

Hmm I don't think I've seen typestates used on a builder struct before, so I don't have any examples to compare against. Some use of typestates is quite nice here (for example `FaultEnabled`/`FaultDisabled`), but my first suggestion would still be to save the alignment and frequency values as members of `PwmBuilder` and then apply them at finalize time.

> I would want it to round to nearest instead of up or down

Sounds good to me! I would just mention that in a doc comment also.

**Comment by richardeoin** [comment]

**Comment by mlamoore** [comment]
I think this PR is ready to go; please give it a review!

**Comment by richardeoin** [comment]
Looks really great, I will try it out hardware again soon

File: examples/pwm_advanced.rs [hunk 2]
```diff
@@ -0,0 +1,411 @@
...
+
+    // Constrain and Freeze power
+    info!("Setup PWR...                  ");
+    let pwr = dp.PWR.constrain();
+    let pwrcfg = example_power!(pwr).freeze();
+
+    // Constrain and Freeze clock
+    info!("Setup RCC...                  ");
+    let rcc = dp.RCC.constrain();
+    let ccdr = rcc.sys_ck(8.mhz()).freeze(pwrcfg, &dp.SYSCFG);
```
richardeoin, line 74: I would specify `.hclk(4.mhz())` directly here. Without this the example breaks on my NUCLEO-H7A3ZI-Q because for this family the HAL defaults to a [sys_clk/hclk ratio of 1 rather than 2](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/rcc/mod.rs#L710-L714) and thus ends up with a `timx_ker_ck` of 8MHz rather than the 4MHz you were expecting. That in turn breaks the deadtime on TIM1 (too many ticks) and TIM2 where the period is specified.

You could make the example even clearer with an explicit assert on the clocks you need. Something like:
```rust
assert_eq!(ccdr.clocks.timx_ker_ck().0, 4_000_000);
```
That causes a panic earlier if someone (for example) specifies pclks which push the hclk/timx_ker_clk ratio to something other than 1 (see RM0433 Rev 7 Table 56).

richardeoin:
I tried this on my NUCLEO-H7A3ZI-Q again and it works well apart from the one comment above

richardeoin (approved):

mlamoore:
Is there anything else you'd like to see before this gets merged?

richardeoin:
Nope, everything else worked when I tested it 🎉  

Thanks for the contribution!

bors r+

PR #176, opened 2021-01-04
[qspi] Add hardware configuration structure

richardeoin:
* Bus width (but the `configure_mode` method can also be used)
* Address size (8/16/24/32-bit)
* Number of dummy cycles
* Sample on rising or falling edge

The configuration structure replaces the frequency argument in the constructors, but backwards compatibility is maintained with an Into<Config> implementation.


richardeoin:

PR #177, opened 2021-01-07
Add initial support for MDMA

richardeoin:
TODO:

- [x] Support for memory-memory transfers
- [x] Example of memory-memory transfers - [examples/mdma.rs](https://github.com/richardeoin/stm32h7xx-hal/blob/dma-mdma1/examples/mdma.rs)
- [x] Support for memory-peripheral transfers
- [ ] Example of memory-peripheral transfers
- [ ] Support for configurable AXI burst sizes
- [ ] Support for different source/destination types for memory-memory transfers

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -190,3 +190,7 @@ required-features = ["rm0433"]
 [[example]]
 name = "spi-dma-rtic"
 required-features = ["rm0433","rt"]
+
+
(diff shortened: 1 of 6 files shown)


richardeoin:
This isn't yet part of #153, since it requires the nightly PAC (and the CI fails because of the nightly PAC combined with our `rtc` feature).

PR #178, opened 2021-01-07
Implement DMA for constant source buffers

richardeoin:
The lack of support for constant source buffers was first raised by @ryan-summers here https://github.com/stm32-rs/stm32h7xx-hal/pull/153#discussion_r515865869

This solution adds a 5th generic type parameter to `Transfer`, and uses this to implement `Transfer` for two different type constraints on `BUF`.

The additional generic type parameter will also be useful for MDMA support, which will also need its own versions of `init`, `next_transfer`, `next_transfer_with` and so on.

The downside is the additional complexity of another type parameter.

### DIFF
--- a/examples/dma.rs
+++ b/examples/dma.rs
@@ -81,13 +81,14 @@ fn main() -> ! {
         .peripheral_increment(true) // source mem
         .fifo_enable(true);
 
-    let mut transfer: Transfer<_, _, MemoryToMemory<u32>, _> = Transfer::init(
-        streams.4,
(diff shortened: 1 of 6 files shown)


richardeoin:
Overall I think the additional type parameter is worth it for the extra flexibility it provides. Merging into `dma` now, but still very open to receiving comments there

ryan-summers:
I don't think the type parameter is too much to ask. I'll give the const-buffer changes some test usage. Thanks for getting this together!

PR #183, opened 2021-01-30
Updating smoltcp on DMA branch

ryan-summers:
This PR merges `master` back into `dma` so that we can get the smoltcp updates into `dma` as well.

As an aside, it seems to me like we should probably merge `dma` in to master - I'm fairly sure a lot of people are essentially treating `dma` _as_ `master` to get access to DMA transfers right now. Curious to hear thoughts here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-## [0.8.0] 2020-10-26
+## [v0.8.0] 2020-10-26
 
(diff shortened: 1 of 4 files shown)


richardeoin:
I think there's a reasonable change of breaking changes still, and also changes in `embedded-dma`. But indeed if projects are using the dma implementation (and I see stabilizer is), then it might as well be in master. So let's finish the current review thread and merge the merge the `dma` branch.

richardeoin:
That was unexpected bors. Actually I'd rather you didn't delete master

PR #185, opened 2021-02-13
Update stm32h7

mattico:
stm32h7 should have a release soon. Preparing for that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,11 @@
 
 ## [Unreleased]
 
-* Updates `cortex-m` to v0.7.1. `cortex-m` v0.6.5+ [are forward compatible with v0.7.0+][cm6-changelog], so
-  running `cargo update` should be sufficient to avoid dependency type version problems.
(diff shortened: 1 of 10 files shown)


mattico:
Hmm, some DAC changes as well.

mattico:
see https://github.com/stm32-rs/stm32-rs/pull/499

richardeoin:
Thanks for this! I'll finish off the DAC changes, in this case adding new type parameters in the right place can get a bit tricky..

richardeoin:
Hmm, when building the RTIC examples we end up with both cortex-m 0.6 (from RTIC) and cortex-m 0.7 (everything else) in the tree, and the compiler sees types from each version as distinct.

Potentially this can be resolved by tweaking the examples

mattico:
The newer releases of cortex-m 0.6 re-export the types from cortex-m 0.7, so you should be able to just call `cargo update`.

mattico:
I see now that stm32 0.6 is not fully forward compatible: https://github.com/rust-embedded/cortex-m/issues/333

richardeoin:
Looks good, I added a little more explanation in the changelog.

bors r+

PR #187, opened 2021-02-14
Add new/new_unchecked methods for USB structures

richardeoin:
Breaking change on the USB API, since the structure no longer contains instances of the pin types


mattico:
Yes, this looks absolutely fine. Being able to take `impl Into<Usb1UlpiDirPin>` makes that a bit nicer as well.

mattico (approved):

PR #189, opened 2021-02-15
USB: Implement instructions_per_us

mattico:
A minor improvement on #187 if https://github.com/stm32-rs/synopsys-usb-otg/pull/16 is merged/released.


richardeoin:
Nice, this looks great. I'll wait and see what happens upstream before picking this or #187

mattico:
We could merge #187 for 0.8.0, maybe including changing the clock argument to be `&CoreClocks` so we're ready for this when upstream is.

richardeoin:
Let's do that over on #187

mattico:
`force_reset` is being implemented in a way that doesn't require this change anymore.

PR #191, opened 2021-02-16
Add barriers to interrupt clearing functions

### LINKED ISSUE #190: Some interrupts fire twice depending on how you clear the interrupt flag
I've noticed this happening in EXTI interrupt handlers run through RTIC. It seems to be a bus synchronization thing: http://false.ekta.is/2012/06/interrupt-service-routines-double-firing-on-stm32/ Probably need a `dsb()` somewhere?

mattico:
If you clear an interrupt flag and quickly return from the interrupt, the interrupt will often immediately execute again. There are two possible causes for this:

1. The peripheral bus that executes the interrupt clear operation is clocked slower than the CPU clock, so there are a few instructions that are executed after clearing the IRQ flag but before the write takes effect.

2. The instruction pipeline may be executing future instructions that haven't 'noticed' the changed interrupt flag.

Adding the appropriate barrier instructions ensures that the side effects of the register write occur before the program continues executing.

See ARM Application Note 321.

Fixes #190

### DIFF
--- a/src/dma/bdma.rs
+++ b/src/dma/bdma.rs
@@ -246,10 +246,11 @@ impl<I: Instance> StreamsTuple<I> {
 //
 // The implementation does the heavy lifting of mapping to the right fields on
 // the stream
+// (Stream0, 0, ifcr, ctcif1, chtif1, cteif1, cgif1, isr, tcif1, htif1),
 macro_rules! bdma_stream {
     ($(($name:ident, $number:expr,
         $ifcr:ident, $tcif:ident, $htif:ident, $teif:ident, $gif:ident,
-        $isr:ident, $tcisr:ident, $htisr:ident)
+        $isr:ident, $tcisr:ident, $htisr:ident, $teisr:ident, $gisr:ident)
     ),+$(,)*) => {
         $(
             impl<I: Instance> Stream for $name<I> {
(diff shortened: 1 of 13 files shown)


richardeoin:
This is indeed an issue, and I expect using the crate with both the icache and dcache enabled will uncover problems of this nature. However so far in this crate we've been quite careful to only add memory barriers where needed (for instance in the DMA HAL) and not add unnecessary ISB/DSBs.

> The instruction pipeline may be executing future instructions that haven't 'noticed' the changed interrupt flag.

I particularly disagree with this. The CM-7 core does not re-order data transfers within peripherals, and DAI0321A says as much: 
image

Futhermore the decision to trigger an interrupt is not made by future instructions or their data, but by the NVIC based on IRQ lines. Section 4.9 has a very helpful diagram

image

I think the first of the causes you gave does get to the heart of the problem though

> so there are a few instructions that are executed after clearing the IRQ flag but before the write takes effect.

Indeed it seems that it is possible to finish executing the ISR before the IRQ line has gone low, particularly when those final instructions are hot in the icache. However that points to a different solution, which is to poll the peripheral register until the flag (and thus the IRQ line) has gone low. There's no need for memory barriers when doing this, because the reads from peripheral memory will always be in-order.

It's possible that one extra read is needed after confirming the flag is de-asserted. This would account for the delay through the clock synchroniser. 

There could be something I've missed here, but I'm reasonably convinced that polling for the IRQ to be de-asserted is a better approach.

mattico:
Appreciate the thorough explanation. I was reading the wrong parts of the app note, it seems. Also my initial test polling the IRQ didn't work because I didn't add it to both functions in gpio.rs ...

I'll try changing to a polling model.

richardeoin:
This implementation looks good, thanks for covering so many peripherals!

bors r+

PR #192, opened 2021-02-19
Added interrupt for wwdg

diondokter:
Tested it and it works. It follows the same interrupt pattern and names as the RTC.

It took me a while to test because the PAC/SVD has an error it where it says there are only 2 prescaler bits. I was wondering why the timeout came so quickly... :P
I'm fixing that here: https://github.com/stm32-rs/stm32-rs/pull/502

### DIFF
--- a/src/watchdog.rs
+++ b/src/watchdog.rs
@@ -13,6 +13,16 @@ use crate::stm32::WWDG1 as WWDG;
 #[cfg(all(feature = "rm0399", feature = "cm4"))]
 use crate::stm32::WWDG2 as WWDG;
 
+/// Event enum for [SystemWindowWatchdog]
+pub enum Event {
+    /// Early wakeup interrupt. This will generate an interrupt when the watchdog would otherwise reset.
+    /// This interrupt can then be used for data logging or even recovery.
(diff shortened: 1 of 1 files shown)


src/watchdog.rs
```diff
@@ -82,6 +129,9 @@ impl WatchdogEnable for SystemWindowWatchdog {
         self.wwdg.cfr.modify(|_, w| w.wdgtb().bits(wdgtb));
```
mattico, line 129:
```suggestion
        // TODO: stm32h7 0.14.0 WDGTB is 3 bits
        const WDGTB_MASK: u32 = 0b111 << 11;
        self.wwdg.cfr.modify(|_, w| unsafe { w.bits((r.bits() & !WDGTB_MASK) | (wdgtb << 11) });
```

mattico: Might as well fix that issue until the PAC is updated, it could be a while.

diondokter: I've now fixed it in the PAC on master. You sure we want this hack in the HAL for the time being?

richardeoin:
Looks great, thanks!

bors r+

PR #195, opened 2021-02-26
Use register reads to delay peripheral interrupt clearing

mattico:
If the interrupt pending flag is set quickly multiple times it could be set after the pending flag is cleared but before the clear could be read by the busy wait loop, causing it to loop forever waiting for the pending flag to be cleared again.

### DIFF
--- a/src/dma/bdma.rs
+++ b/src/dma/bdma.rs
@@ -283,13 +283,8 @@ macro_rules! bdma_stream {
                                     .$teif().set_bit() //Clear transfer error interrupt flag
                                     .$gif().set_bit() //Clear global interrupt flag
                     );
-                    while {
-                        let r = dma.$isr.read();
-                        r.$tcisr().bit_is_set() ||
-                        r.$htisr().bit_is_set() ||
-                        r.$teisr().bit_is_set() ||
-                        r.$gisr().bit_is_set()
-                    } {}
+                    let _ = dma.$isr.read();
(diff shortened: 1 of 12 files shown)


richardeoin:
Ok, I see this is a problematic edge-case. Perhaps the original issue (#190) can still be fixed by converting

```rust
while xxx.isr.read().yyy().bit_is_set() {}
```

to  

```rust
let _ = xxx.isr.read(); // At least one cycle of peripheral clock
let _ = xxx.isr.read(); // At least one cycle of peripheral clock
```

The idea is that this still gives the peripheral enough cycles to process the Interrupt Clear write before the ISR exits.

src/rtc.rs
```diff
@@ -685,27 +685,27 @@ impl Rtc {
...
             Event::Wakeup => {
                 self.reg.cr.modify(|_, w| w.wutie().clear_bit());
-                exti.unlisten(ExtiEvent::RTC_WAKEUP);
+                exti.listen(ExtiEvent::RTC_WAKEUP);
                 exti.rtsr1.modify(|_, w| w.tr19().disabled());
             }
             Event::Timestamp => {
                 self.reg.cr.modify(|_, w| w.tsie().clear_bit());
-                exti.unlisten(ExtiEvent::RTC_OTHER);
+                exti.listen(ExtiEvent::RTC_OTHER);
```
mattico, line 708: Remember to not revert this fix

mattico:
Okay, that seems to work.

PR #196, opened 2021-02-28
Prepare for v0.9.0

richardeoin:


hargoniX:
Do we want to wait on #199 before the release still?

richardeoin:
Sure, please keep reviewing it and I'll check again in a few days 👍

richardeoin:
Rebased onto master and added the `crc` module to the build on docs.rs

PR #198, opened 2021-03-07
Ethernet: Remove extraneous interrupt-free regions

### LINKED ISSUE #197: Ethernet DMA ring management: Gratuitous interrupt-free regions?
The Ethernet DMA driver places the memory barrier and descriptor tail buffer writes into an interrupt-free region, e.g.:

https://github.com/stm32-rs/stm32h7xx-hal/blob/4ca09fa94af0fed8084633c142767cf6515ed8de/src/ethernet/eth.rs#L322-L331

We didn't have this in the quartiq/stabilizer driver, and I can't think of a good reason for it – what's the story here?

dnadlinger:
There is no gain from disabling interrupts between the memory barriers
and writing to the DMA engine descriptor ring pointers - even if an
interrupt is executed between the barrier and the atomic store to the
pointer, it doesn't make the descriptors any less valid.

Interrupts are also irrelevant for transmit/receive ring initialisation,
which needs to run wiht the DMA engine disabled anyway.

GitHub: Fixes #197.

### DIFF
--- a/src/ethernet/eth.rs
+++ b/src/ethernet/eth.rs
@@ -124,18 +124,17 @@ impl TDesRing {
         }
         self.tdidx = 0;
 
-        cortex_m::interrupt::free(|_cs| unsafe {
+        // Initialise pointers in the DMA engine. (There will be a memory barrier later
+        // before the DMA engine is enabled.)
+        unsafe {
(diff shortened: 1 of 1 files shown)


richardeoin:
Thanks!

bors r+

PR #199, opened 2021-03-07
Add support for CRC unit

AlyoshaVasilieva:
This PR adds support for the hardware CRC unit. In addition to the hardware features, this adds automatic output XOR with a user-configured value; many CRC algorithms use an all-1 XOR and a few use a bit pattern XOR.

Tested on an STM32H750VBT6 and an STM32H743VIT6.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,7 @@ fmc = ["stm32-fmc"]
 sdmmc = ["sdio-host"]
 ethernet = ["smoltcp"]
 rtc = ["chrono"]
+crc = []
 rt = ["stm32h7/rt"]
 usb_hs = ["synopsys-usb-otg", "synopsys-usb-otg/hs"]
 stm32h742 = ["stm32h7/stm32h743", "device-selected", "rm0433"]
@@ -184,4 +185,4 @@ required-features = ["rm0433","rt"]
 
 [[example]]
 name = "crc"
-required-features = ["rt"]
+required-features = ["crc", "rt"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -181,3 +181,7 @@ required-features = ["rm0433"]
 [[example]]
 name = "spi-dma-rtic"
 required-features = ["rm0433","rt"]
+
+[[example]]
+name = "crc"
+required-features = ["rt"]
--- a/examples/crc.rs
+++ b/examples/crc.rs
@@ -1,0 +1,76 @@
+//! Example of using the CRC peripheral with various configurations
+//!
+//! A good resource of CRC algorithms, to assist in determining their exact settings,
+//! is https://reveng.sourceforge.io/crc-catalogue/
+
+#![no_main]
+#![no_std]
+
+use cortex_m_rt::entry;
+use log::info;
+
+use stm32h7xx_hal::crc::{Config, Crc, Polynomial};
+use stm32h7xx_hal::{pac, prelude::*};
+
+#[macro_use]
+mod utilities;
+
+/// Standard input data for testing CRCs
+const CHECK_DATA: &[u8] = b"123456789";
+
+#[entry]
+fn main() -> ! {
+    utilities::logger::init();
+    let dp = pac::Peripherals::take().unwrap();
+
+    // Constrain and Freeze power
+    info!("Setup PWR...                  ");
+    let pwr = dp.PWR.constrain();
+    let pwrcfg = example_power!(pwr).freeze();
+
+    // Constrain and Freeze clock
+    info!("Setup RCC...                  ");
+    let rcc = dp.RCC.constrain();
+
+    let ccdr = rcc.sys_ck(200.mhz()).freeze(pwrcfg, &dp.SYSCFG);
+
+    info!("");
+    info!("stm32h7xx-hal example - CRC");
+    info!("");
+
+    let mut crc = Crc::enable(dp.CRC, ccdr.peripheral.CRC);
+    // feed standard check data and ensure result is expected
+    // the default config is CRC32 MPEG2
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -181,3 +181,7 @@ required-features = ["rm0433"]
 [[example]]
 name = "spi-dma-rtic"
 required-features = ["rm0433","rt"]
```
hargoniX, line 183: Since CRC isn't standartized via embedded_hal yet it would be nice if we could feature gate it so only people who actually want to use it have to compile the code and the docs don't get too cluttered.

examples/crc.rs
```diff
@@ -0,0 +1,76 @@
...
+    info!("Setup RCC...                  ");
+    let rcc = dp.RCC.constrain();
+
+    let ccdr = rcc.sys_ck(200.mhz()).freeze(pwrcfg, &dp.SYSCFG);
+
+    info!("");
+    info!("stm32h7xx-hal example - CRC");
+    info!("");
+
+    let mut crc = Crc::enable(dp.CRC, ccdr.peripheral.CRC);
```
hargoniX, line 41: Usually we'd expect this to be done via extension traits on the PAC peripheral since this is already how every other peripheral we implemented does it (you can just check any example and the *Ext traits if you do not know what I'm referring to) so just a little inconsistency to the rest of our API.

src/crc.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    /// Read the CRC without resetting the unit.
+    #[inline]
+    pub fn read_crc(&self) -> u32 {
+        self.reg.dr().read().dr().bits() ^ self.output_xor
+    }
+
+    /// Write the independent data register. The IDR can be used as
+    /// temporary storage. It is not cleared on CRC hash reset.
+    pub fn set_idr(&mut self, value: u32) {
```
hargoniX, line 93: Having an example of usage for this would be nice if you can come up with some.

src/crc.rs
```diff
@@ -0,0 +1,356 @@
...
+        self.reg.idr.write(|w| w.idr().bits(value));
+    }
+
+    /// Get the current value of the independent data register.
+    pub fn get_idr(&self) -> u32 {
```
hargoniX, line 98: Having an example of usage for this would be nice if you can come up with some.

hargoniX (changes requested):
Just a few inconsistencies with how we usually do things in our API, apart from that very nice job, thank you for your contribution.

src/crc.rs
AlyoshaVasilieva: I don't know any specific use case beyond storing a u32. In many other chips it has 24 reserved bits, so it's possible it was used internally by the unit and just exposed because earlier chips had much less RAM.

It's possible to use it to store an ongoing CRC calculation* but it's not specific to the IDR (anywhere works) and mildly complicated to the point where I'd probably want to add assistance to the CRC impl itself, rather than have users do most of it. Some kind of `read_crc_state` method, eg:

```rust
pub fn read_crc_state(&self) -> u32 {
    let state = self.read_crc_no_xor();
    if self.reg.cr.read().rev_out().is_reversed() {
        state.reverse_bits()
    } else {
        state
    }
}
```

\* For example an incomplete CRC calculation could have its DR read and stored, then unit could be reset and used to calculate a different CRC for presumably less data, then config restored with old DR as INIT. I think the CRC unit supports DMA in some way, where this might be more useful. I've never used DMA on any MCU so I wouldn't know for sure.

Rough code example:

```rust
fn idr_example(crc: &mut Crc) {
    let config = Config::new().reflect(true).output_xor(0xFFFF_FFFF);
    crc.set_config(&config);
    crc.update(b"12345");
    crc.set_idr(crc.read_crc_state()); // store ongoing CRC

    crc.set_config(&Config::new()); // perform unrelated CRC
    assert_eq!(crc.update_and_read(CHECK_DATA), 0x0376_E6E7);

    // retrieve ongoing CRC from IDR, set it as the initial value
    crc.set_config(&config.initial_value(crc.get_idr()));
    assert_eq!(crc.update_and_read(b"6789"), 0xCBF4_3926); // finish CRC, check result
}
```

If this seems potentially useful I can add it. The `read_crc_state` method may be useful enough on its own without involving the IDR.

Cargo.toml
AlyoshaVasilieva: Added feature gate.

examples/crc.rs
AlyoshaVasilieva: Added extension trait, method is now `dp.CRC.crc(ccdr.peripheral.CRC)`. Wasn't sure if I should name it `crc`, `enable`, or `constrain` but I think this otherwise matches other peripherals.

src/crc.rs
hargoniX: Hmmm the reference manual doesn't tell us more than
```
The CRC_IDR register can be used to hold a temporary value related to CRC calculation.
```
either. And not even the CRC application note (AN4187), which does go into a bit of detail regarding DMA, mentions it. So its probably only some zombie register that nobody really needs I guess.

In this case, could you add a comment that explains to the user that this register doesn't play any role in the actual calculation of the CRC so people who accidentally stumble over it don't happen to waste their time on trying to figure out what it's good for?

mlamoore:
I've seen it used to suspend and resume a CRC calculation when a higher
priority calculation comes through, but that only allowed one level of
reentrancy, while a local variable in the CRC calculation function allows
infinite levels of reentrancy.

In my opinion, I'd rather just leave it out altogether, no need to have the
HAL expose every register on the chip, especially less useful ones like
this.

-Morgan LaMoore

hargoniX (approved):

mlamoore:
Yeah, I have no problem with that.

PR #200, opened 2021-03-08
SPI: Support 32-bit frames

mattico:


richardeoin:
LTGM, thanks for both PRs!

bors r+

PR #208, opened 2021-03-29
Better calculations for set_timeout_ticks

diondokter:
Ran into an issue where my devices would crash randomly.
I've dug around and found that when the set_timeout_ticks function gets a ticks value of an exact multiple of 2^16, it will overflow.

Also when `ticks > u32::MAX - u16::MAX`, the `+1` would overflow. I think it was tried to do it in 32 bit, but the parentheses were wrong.

This solves those issues. All psc and arr values remain the same with this change, except that there should be no panics anymore and all ticks values are valid.

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -350,7 +350,7 @@ macro_rules! hal {
                 }
 
                 fn set_timeout_ticks(&mut self, ticks: u32) {
-                    let psc = u16((ticks - 1) / (1 << 16)).unwrap();
+                    let psc = u16(ticks / (1 << 16)).unwrap();
                     self.tim.psc.write(|w| w.psc().bits(psc));
 
                     let arr = u16(ticks / u32(psc + 1)).unwrap();
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -349,7 +349,29 @@ macro_rules! hal {
                     self.set_timeout_ticks(ticks.max(1));
                 }
 
+                /// Sets the timer's prescaler and auto reload register so that the timer will reach
+                /// the ARR after `ticks` amount of timer clock ticks.
+                ///
+                /// ```
+                /// // Set auto reload register to 50000 and prescaler to divide by 2.
+                /// timer.set_timeout_ticks(100000);
+                /// ```
+                ///
+                /// This function will round down if the prescaler is used to extend the range:
+                /// ```
+                /// // Set auto reload register to 50000 and prescaler to divide by 2.
+                /// timer.set_timeout_ticks(100001);
+                /// ```
                 fn set_timeout_ticks(&mut self, ticks: u32) {
+                    // We want to have `ticks` amount of timer ticks before it reloads.
+                    // But `ticks` may have a higher value than what the timer can hold directly.
+                    // So we'll use the prescaler to extend the range.
+
+                    // To know how many times we would overflow with a prescaler of 1, we divide `ticks` by 2^16 (the max amount of ticks per overflow).
+                    // If the result is e.g. 3, then we need to increase our range by 4 times to fit all the ticks.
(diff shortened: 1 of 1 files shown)


diondokter:
Update: I've used this branch now for a while and I can confirm for sure that the issue is fixed

src/timer.rs
```diff
@@ -350,10 +350,10 @@ macro_rules! hal {
                 }
 
                 fn set_timeout_ticks(&mut self, ticks: u32) {
-                    let psc = u16((ticks - 1) / (1 << 16)).unwrap();
+                    let psc = u16(ticks / (1 << 16)).unwrap();
                     self.tim.psc.write(|w| w.psc().bits(psc));
 
-                    let arr = u16(ticks / u32(psc + 1)).unwrap();
```
ryan-summers, line 356: I'm fairly sure the only required change here is `u16(ticks / psc as u32 + 1).unwrap()`. I don't see how removing the `ticks - 1` changes anything here?

diondokter: Ok, let's do the calculations manually. I agree that this code is non-obvious and I could add some comments.
```
ticks = 65536
psc = (ticks - 1) / 65536 = 65535 / 65536 = 0
arr = 65536 / (psc + 1) = 65536 / (0 + 1) = 65536 / 1 = 65536 = overflow
```

With the -1 removed:
```
ticks = 65536
psc = ticks / 65536 = 65536 / 65536 = 1
arr = 65536 / (psc + 1) = 65536 / (1 + 1) = 65536 / 2 = 32768
```
The result is correct because  `32768 * (psc + 1) = ticks` is true.

src/timer.rs
```diff
@@ -349,11 +349,33 @@ macro_rules! hal {
...
+                    // To know how many times we would overflow with a prescaler of 1, we divide `ticks` by 2^16 (the max amount of ticks per overflow).
+                    // If the result is e.g. 3, then we need to increase our range by 4 times to fit all the ticks.
+                    // We can increase the range enough by setting the prescaler to 3 (which will divide the clock freq by 4).
+                    // Because every tick is now 4x as long, we need to divide `ticks` by 4 to keep the same timeout.
+
+                    let psc = u16(ticks / (1 << 16)).unwrap();
                     self.tim.psc.write(|w| w.psc().bits(psc));
 
-                    let arr = u16(ticks / u32(psc + 1)).unwrap();
+                    let arr = u16(ticks / (u32(psc) + 1)).unwrap();
```
ryan-summers, line 378: Thanks for the explanation - I see your issue now. Would you mind adding comments to why these unwraps are safe? E.g:
```
// Note(unwrap): If ticks > u16::max, PSC +1 is guaranteed to be  divisor such that ticks / divisor will fit within a u16.
```

ryan-summers:
Some final doc updates and this should be okay to merge. It may be worth adding in some tests of these calculations as well so we don't get them wrong in the future

diondokter:
I extracted it to a separate function so I could add some tests.

_Btw, I couldn't get the test to compile on windows so I had to ssh into my server. Don't know if that's known or even cared about._

richardeoin:
Nice, thanks @diondokter for fixing the underlying bug and breaking it out into a nice testable function, and @ryan-summers for reviewing!

I have never tried rust on windows, but if you find an easy fix to your problem it would be great to have of course

bors r+

diondokter:
Oh no! I just spotted a spelling mistake!
Abort! :P

richardeoin:
I think your wish was granted by a random error in github actions.. 🙃

diondokter:
Haha, I fixed it :P

richardeoin:
Great 😃  

It seems github is hitting its error quite often, but let's try once more

bors r+

richardeoin:
^ I'll retry when github actions is fixed

richardeoin:
bors retry

PR #209, opened 2021-03-31
Fixed daisy examples

mtthw-meyer:
Fixed and tested both of these on my daisy board.

sai-i2s-passthru was panicking due to calling the core wrong.
sai_dma_passthru needed it's jumpstart.


examples/sai-i2s-passthru.rs
```diff
@@ -1,19 +1,16 @@
 // This demo code runs on the Electro Smith Daisy Seed board
 // https://www.electro-smith.com/daisy
+#![allow(unused_macros)]
 #![deny(warnings)]
-#![deny(unsafe_code)]
```
richardeoin, line 4: Please leave this lint enabled, although I appreciate it has to be temporarily removed for debugging

examples/sai_dma_passthru.rs
```diff
@@ -2,23 +2,24 @@
 // https://www.electro-smith.com/daisy
 
 #![allow(unused_macros)]
-#![deny(warnings)]
```
richardeoin, line 5: Same for this lint

examples/sai-i2s-passthru.rs
```diff
@@ -39,7 +40,8 @@ const APP: () = {
     }
 
     #[init]
-    fn init(ctx: init::Context) -> init::LateResources {
+    fn init(mut ctx: init::Context) -> init::LateResources {
+        utilities::logger::init();
```
richardeoin, line 44: I guess this is what causes the panic in `device::CorePeripherals::take()` further down, but in any case it looks good now!

examples/sai_dma_passthru.rs
```diff
@@ -60,11 +77,26 @@ fn main() -> ! {
...
+        .use_hse(HSE_CLOCK_MHZ)
+        .sys_ck(CLOCK_RATE_HZ)
+        .pclk1(PCLK_HZ) // DMA clock
+        // PLL1
+        .pll1_strategy(rcc::PllConfigStrategy::Iterative)
+        .pll1_p_ck(PLL1_P_HZ)
+        .pll1_q_ck(PLL1_Q_HZ)
+        .pll1_r_ck(PLL1_R_HZ)
+        // PLL2
+        .pll2_p_ck(PLL2_P_HZ) // Default adc_ker_ck_input
```
richardeoin, line 89: Is there a reason for specifying all these PLL outputs here? It would be simpler to leave them out if they're not used for this example

examples/sai_dma_passthru.rs
```diff
@@ -45,6 +46,22 @@ const PLL3_P_HZ: Hertz = Hertz(AUDIO_SAMPLE_HZ.0 * 257);
 static mut TX_BUFFER: [u32; DMA_BUFFER_LENGTH] = [0; DMA_BUFFER_LENGTH];
 #[link_section = ".sram3"]
 static mut RX_BUFFER: [u32; DMA_BUFFER_LENGTH] = [0; DMA_BUFFER_LENGTH];
+pub const CLOCK_RATE_HZ: Hertz = Hertz(400_000_000_u32);
+
+const HSE_CLOCK_MHZ: MegaHertz = MegaHertz(16);
+
+// PCLKx
+const PCLK_HZ: Hertz = Hertz(CLOCK_RATE_HZ.0 / 4);
+// 49_152_344
```
richardeoin, line 55: Not sure what this number is doing in the comment here, perhaps remove it to avoid confusion

richardeoin (commented):
Thanks for taking the time to clean up and improve these! I left a few inline comments

examples/sai_dma_passthru.rs
mtthw-meyer: Sure, I can remove those. I put them in for debugging some stuff but they shouldn't be needed.

examples/sai_dma_passthru.rs
mtthw-meyer: Sure, I can remove those. I put them in for debugging some stuff but they shouldn't be needed.

examples/sai-i2s-passthru.rs
mtthw-meyer: The log-itm code relies on an unsafe block. Unsafe cannot be denied if the utility logger is used the checks choke on it.

https://github.com/stm32-rs/stm32h7xx-hal/blob/a625d2b56a675dd2990056751301bb136338a86c/examples/utilities/logger.rs#L20

richardeoin: Hmm yes good point. Let's leave it as-is for now then

richardeoin:
Looks great!

bors r+

PR #211, opened 2021-04-08
RTC Improvements

mattico:
Some small tweaks to the RTC initialization that I found while debugging, and added a note to LSE CSS configuration about an errata.


richardeoin:
Looks good, thanks for the tweaks!

bors r+

PR #216, opened 2021-04-23
Extended SPI with more hardware CS features

diondokter:
The hardware CS now comes in 4 specific modes that are setup in the config:

- Disabled
- EndlessTransaction
- WordTransaction
- FrameTransaction

New here is the frame transaction where `TSIZE` is used to send frames of a specific length.
There's now also `end_transaction` that is useful for endless and mandatory for frame transactions.

The blocking implementation now checks if we're in frame mode and will send frames when we are.

I don't have a nice dev board, so all testing was with an oscilloscope with the probe manually held to the pins. So I couldn't check everything as good as I'd like, but it should all work from what I could see.

### DIFF
--- a/examples/spi_hardware_cs.rs
+++ b/examples/spi_hardware_cs.rs
@@ -60,17 +60,17 @@ fn main() -> ! {
         (sck, miso, mosi, hcs),
         // Create a config with the hardware chip select given
         spi::Config::new(spi::MODE_0)
-                // Put 1 us idle time between every word sent. (the max is 15 spi peripheral ticks)
-                .inter_word_delay(0.000001) 
-                // Specify that we use the hardware cs
-                .hardware_cs(spi::HardwareCS {
-                    // See the docs of the HardwareCSMode to see what the different modes do
-                    mode: spi::HardwareCSMode::EndlessTransaction,
-                    // Put 1 us between the CS being asserted and the first clock
-                    assertion_delay: 0.000001,
-                    // Our CS should be high when not active and low when asserted
-                    polarity: spi::Polarity::IdleHigh,
-                }),
+            // Put 1 us idle time between every word sent. (the max is 15 spi peripheral ticks)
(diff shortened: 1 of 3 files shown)


hargoniX:
Is there a good reason as to why you aren't consuming the hardware CS pin as well just like we consume the other pins so users don't end up messing around with the pins while we are using them?

diondokter:
This wasn't done before either with the existing hardware CS support. Doesn't mean that wouldn't be a good thing to do though...
Should I look into adding that?

hargoniX:
I think that's a good idea yeah...and since you're speaking of some testing code, maybe you could come up with some sort of minimal example for examples/ as well? I see spi, spi-dma, and spi-dma-rtic there but no concrete example for hardware CS yet so that'd definitely be cool as well! (Speaking from experience, the H7 HAL is known among the other stm32-rs HALs for having examples for most of its features)

diondokter:
Sure! I'll look into taking ownership of the CS pin and I will add an example.

diondokter:
Ok, both have been added now

hargoniX (approved):
I guess if we really wanted to, we could move the assert_eq! regarding the presence of a HW cs pin into the type system by also moving a bit of the SPI config into the type system and only allowing SPI config types with compatible Pin trait implementations but that would just increase our types and complexity some more for basically no benefit so lets keep it like this!

diondokter:
> HW cs pin into the type system

I had considered that. But this way people can keep their old pin config and don't need to add a NoHCS pin.

Glad this got in! 😁
Now the chip I talk to doesn't bug out anymore at random

hargoniX:
In the end its
- If you provide a Hardware CS pin and try to use it you should presumably be clever enough to also provide a proper config for it
- -An assert that happens during initialisation of the peripheral,so if a user makes a mistake he should notice that preeeetty quickly and not when they're already deploying it into production.

PR #221, opened 2021-05-10
Updating smoltcp dependency features

### LINKED ISSUE #220: Smoltcp feature flags invalid for smoltcp version 0.7.1
Smoltcp release 0.7.1 will generate a compiler error with the feature flags that `stm32h7xx-hal` provides, as `smoltcp` appears to have revamped how feature flags are required between 0.7.0 and 0.7.1

This means that 0.7.1 is not backward compatible with 0.7.0, so if Cargo decides to use 0.7.1 instead of 0.7.0 (which it is allowed to do via the config options here), smoltcp breaks the build.

ryan-summers:
This PR fixes #220 by using a valid feature set for the `smoltcp` dependency. The existing feature set was invalid, but worked because of a Cargo bug that caused the `dev-dependencies` feature to unify with the normal features.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ embedded-display-controller = { version = "^0.1.0", optional = true }
 [dependencies.smoltcp]
 version = "0.7.0"
 default-features = false
-features = ["ethernet", "proto-ipv4"]
+features = ["ethernet", "proto-ipv4", "socket-raw"]
 optional = true
 
 [dependencies.chrono]


richardeoin:
Good catch, thanks.

I wonder how we could detect this in CI? It seems that `cargo build --features stm32h743v,rt,ethernet` isn't enough to surface this error, but `cargo build --features stm32h743v,rt,ethernet -Z avoid-dev-deps` would do the trick (on nightly of course).

PR #225, opened 2021-05-23
Avoid creating intermediate references to packed fields

richardeoin:
Such an intermediate reference is UB, even if it is never dereferenced. The previous code that created addresses / raw pointers via intermediate references will become a hard error in a future version of Rust. Use [core::ptr::addr_of](https://doc.rust-lang.org/core/ptr/macro.addr_of.html) instead.

This PR increases the MSRV to 1.51.0


richardeoin:
bors retry

PR #226, opened 2021-05-24
Fast double-buffered DMA

jordens:
When handling multiple double-buffered DMA streams the current safe-and-universal API becomes slower than necessary.
This adds `next_dbm_transfer_with()`, a closure-based API to the inactive buffer, without address poisoning, compiler fences, data barriers, or buffer swapping. For some applications this reduces the overhead significantly (speedup of 5).

### DIFF
--- a/src/dma/bdma.rs
+++ b/src/dma/bdma.rs
@@ -288,11 +288,18 @@ macro_rules! bdma_stream {
                 }
 
                 #[inline(always)]
-                fn clear_transfer_complete_interrupt(&mut self) {
+                fn clear_transfer_complete_flag(&mut self) {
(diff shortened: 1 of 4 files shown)


richardeoin:
It's really interesting that such a dramatic speedup is possible. I guess that's with release mode / optimisations enabled in both cases?

The PR looks good to me.

jordens:
Yes. That's always in `release` mode. The gain is mostly in cancelling the buffer `Option` swapping operations, several address writes through the slower bus matrices, various checks of DMA state that depend on whether it's double buffer or not, and the redundant `fence` in the case of multi-DMA handlers.

richardeoin (approved):

richardeoin:
Thanks!

bors r+

Issue #165, opened 2020-11-03
No QSPI support for memories


richardeoin:
Some of the features that would be required to communicate with memories over QSPI (instructions, dummy cycles..) are not supported

x37v:
I'm wanting to interface with 25LP064 QSPI flash memory and so I'm hoping my efforts can help this HAL.
I'm wondering if you all have any advice for the API for this?
Would it make sense to add `instruction: Option<u8>` to read and write and `mode: Option<u8>` to read (I haven't seen mode in write)? 
I'm assuming I can update the `imode` value based on the value set in `admode` ?

richardeoin:
I've started implementing some new methods for interfacing with OCTOSPI flash memories, and these are probably applicable to QSPI also. I'll make a PR with the progress so far..

richardeoin:
In the example in that PR there's an instruction with no address and no data phases ([line 80](https://github.com/stm32-rs/stm32h7xx-hal/pull/230/files#diff-37e1cf6c7c3da03fb936b858738b81830c681cced7575695592991fb5186d7ceR80)), as well as an instruction with both address and data phases ([line 84](https://github.com/stm32-rs/stm32h7xx-hal/pull/230/files#diff-37e1cf6c7c3da03fb936b858738b81830c681cced7575695592991fb5186d7ceR84)).

It should be close to working for QSPI memories also, but I've not tried it. If there's only small changes needed to work with the 25LP064, that would be very useful to know.

x37v:
> In the example in that PR there's an instruction with no address and no data phases ([line 80](https://github.com/stm32-rs/stm32h7xx-hal/pull/230/files#diff-37e1cf6c7c3da03fb936b858738b81830c681cced7575695592991fb5186d7ceR80)), as well as an instruction with both address and data phases ([line 84](https://github.com/stm32-rs/stm32h7xx-hal/pull/230/files#diff-37e1cf6c7c3da03fb936b858738b81830c681cced7575695592991fb5186d7ceR84)).
> 
> It should be close to working for QSPI memories also, but I've not tried it. If there's only small changes needed to work with the 25LP064, that would be very useful to know.

Great @richardeoin I'll give this a go soon.

Longer term, I'm guessing the original read/write would need to be altered to correctly do the `setup_extended`.. if they're not removed entirely?

PR #230, opened 2021-06-22
Initial OCTOSPI and QUADSPI memories support

richardeoin:
`write_extended` and `read_extended` methods allow more complicated transactions, such as those needed to read/write from flash memories.

Rename the `quadspi` flag to `xspi`

Ref #165

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
           - stm32h7b0
           - stm32h735
     env:                        # Peripheral Feature flags
-      FLAGS: rt,quadspi,sdmmc,fmc,usb_hs,rtc,ethernet,ltdc,crc
+      FLAGS: rt,xspi,sdmmc,fmc,usb_hs,rtc,ethernet,ltdc,crc
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -19,7 +19,7 @@ jobs:
           - log-semihost
           - log-rtt
     env:                        # Peripheral Feature flags
-      FLAGS: rt,quadspi,sdmmc,fmc,usb_hs,rtc,ethernet,ltdc,crc
+      FLAGS: rt,xspi,sdmmc,fmc,usb_hs,rtc,ethernet,ltdc,crc
 
     steps:
       - uses: actions/checkout@v2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@
 * spi: Add more hardware CS features #216
 * timers: Better calculations for `set_timeout_ticks` #208
 
+* **Breaking**: Rename the `quadspi` flag to `xspi`
+
 ## [v0.9.0] 2021-03-12
 
 * Updates `cortex-m` to v0.7.1. `cortex-m` v0.6.5+ [are forward compatible with
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -148,6 +148,10 @@ name = "qspi"
 required-features = ["xspi", "rm0433"]
 
 [[example]]
+name = "qspi_mdma"
+required-features = ["xspi", "rm0433"]
+
+[[example]]
 name = "octospi"
 required-features = ["xspi", "rm0468"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -148,8 +148,8 @@ name = "qspi"
 required-features = ["xspi", "rm0433"]
 
 [[example]]
-name = "qspi_mdma"
-required-features = ["quadspi", "rm0433"]
+name = "octospi"
+required-features = ["xspi", "rm0468"]
 
 [[example]]
 name = "sdmmc"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,7 +86,7 @@ cm4 = []
 cm7 = []
 smps = []
 ltdc = ["embedded-display-controller"]
-quadspi = []
+xspi = []
 fmc = ["stm32-fmc"]
 sdmmc = ["sdio-host"]
 sdmmc-fatfs = ["embedded-sdmmc", "sdmmc"]
@@ -145,7 +145,7 @@ required-features = ["fmc", "rm0399"]
 
 [[example]]
 name = "qspi"
-required-features = ["quadspi", "rm0433"]
+required-features = ["xspi", "rm0433"]
 
 [[example]]
 name = "qspi_mdma"
--- a/examples/octospi.rs
+++ b/examples/octospi.rs
@@ -1,0 +1,98 @@
+//! OCTOSPI peripheral example in indirect mode
+//!
+//! Tested on a STM32H735G-DK with a Macronix MX25LM51245GXDI00
+
+#![deny(warnings)]
+#![no_main]
+#![no_std]
+
+#[macro_use]
+mod utilities;
+
+use cortex_m_rt::entry;
+use stm32h7xx_hal::rcc::rec::{OctospiClkSel, OctospiClkSelGetter};
+use stm32h7xx_hal::{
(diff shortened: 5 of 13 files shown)


x37v:
@richardeoin I haven't had any luck yet...
I'm not experienced with qspi so maybe I'm just missing some configuration that I need.. should the chip be handling `ncs` itself if I set the correct alternate function?
I'm not seeing an rcc ahb3enr enable, but maybe that happens behind the scenes somehow?

Anyway, I'm trying to read the product information (in 1 line mode) just for some sanity checking.. i only get zeros?

```rust
        let _ncs = pg6.into_alternate_af10().set_speed(Speed::VeryHigh); //QUADSPI_BK1_NCS

        let sck = pf10.into_alternate_af9().set_speed(Speed::VeryHigh);
        let io0 = pf8.into_alternate_af10().set_speed(Speed::VeryHigh);
        let io1 = pf9.into_alternate_af10().set_speed(Speed::VeryHigh);
        let io2 = pf7.into_alternate_af9().set_speed(Speed::VeryHigh);
        let io3 = pf6.into_alternate_af9().set_speed(Speed::VeryHigh);

        let config = Config::new(50.mhz()).mode(QspiMode::OneBit);
        let mut qspi = regs.bank1((sck, io0, io1, io2, io3), config, &clocks, prec);

        let mut info: [u8; 3] = [0xFF; 3];
        while qspi.is_busy().is_err() {}
        qspi.read_extended(
            QspiWord::U8(0x9F),
            QspiWord::None,
            QspiWord::None,
            0,
            &mut info,
        )
        .unwrap();
        while qspi.is_busy().is_err() {}
```

I'm not sure about the pin speeds.. if I don't set them though I get stuck in the read on `while self.rb.sr.read().tcf().bit_is_clear() {}` in the HAL..

richardeoin:
Ok, just to confirm which QSPI memory part (and/or board) are you using to test this? The chip should be handling the `ncs` here iirc, but the HAL documentation about this is indeed lacking. In the H7 HAL the AHB enable happens by passing the `prec` variable, which the HAL then calls `enable()` on.
Could you try the JEDEC ID read with the old API? Whilst that wouldn't work for most instructions, the old API should work for the JEDEC ID read ([this is in the example](https://github.com/stm32-rs/stm32h7xx-hal/pull/230/files#diff-37e1cf6c7c3da03fb936b858738b81830c681cced7575695592991fb5186d7ceR71)). I would also suggest trying a lower clock speed first - maybe 1MHz. Once that's working then you can try higher clock speeds.

x37v:
> Ok, just to confirm which QSPI memory part (and/or board) are you using to test this? The chip should be handling the `ncs` here iirc, but the HAL documentation about this is indeed lacking. In the H7 HAL the AHB enable happens by passing the `prec` variable, which the HAL then calls `enable()` on.

I'm using the `IS25LP064A` 8M flash part on the Daisy Seed board that also has a `stm32h750` on it.

> Could you try the JEDEC ID read with the old API? Whilst that wouldn't work for most instructions, the old API should work for the JEDEC ID read ([this is in the example](https://github.com/stm32-rs/stm32h7xx-hal/pull/230/files#diff-37e1cf6c7c3da03fb936b858738b81830c681cced7575695592991fb5186d7ceR71)). I would also suggest trying a lower clock speed first - maybe 1MHz. Once that's working then you can try higher clock speeds.

I haven't tried the old API for the JEDEC ID but I'll give that a go. I did use a lower clock speed for a bit but not as low as 1MHz so I'll try that and report back.

Thanks @richardeoin !

x37v:
> > Ok, just to confirm which QSPI memory part (and/or board) are you using to test this? The chip should be handling the `ncs` here iirc, but the HAL documentation about this is indeed lacking. In the H7 HAL the AHB enable happens by passing the `prec` variable, which the HAL then calls `enable()` on.
> 
> I'm using the `IS25LP064A` 8M flash part on the Daisy Seed board that also has a `stm32h750` on it.
> 
> > Could you try the JEDEC ID read with the old API? Whilst that wouldn't work for most instructions, the old API should work for the JEDEC ID read ([this is in the example](https://github.com/stm32-rs/stm32h7xx-hal/pull/230/files#diff-37e1cf6c7c3da03fb936b858738b81830c681cced7575695592991fb5186d7ceR71)). I would also suggest trying a lower clock speed first - maybe 1MHz. Once that's working then you can try higher clock speeds.
> 
> I haven't tried the old API for the JEDEC ID but I'll give that a go. I did use a lower clock speed for a bit but not as low as 1MHz so I'll try that and report back.
> 
> Thanks @richardeoin !

I set the clock to 1Mhz and successfully read the JEDEC ID with the old `read` (still on your branch).

x37v:
In case anyone else is watching this, I got the quadspi working on the h750 with some alterations to @richardeoin 's branch. The changes are now merged in.

@richardeoin any thoughts on what is next? Would it make sense to replace the `read` and `write` with the `read_extended` etc? Does it seem like this work could get into this HAL?

Considering the instruction on some of the chips (h750 at least) is always 1 byte, would it make sense to have a feature gated `Option<u8>` for that

richardeoin:
Great that you got this working with the Quadspi peripheral as well!

I would say that the `read` and `write` functions can remain mostly unchanged, since they provide a simple API for the majority of non-memory ICs. The `read_extended` and `write_extended` methods add extra functionality that it would be useful to have in the HAL too imo.

You're right that for the Quadspi peripheral the instruction is always a `u8` (for the Octospi peripheral it can be larger). If there's a nice way to simply the API using feature gates then that would be nice to see!

x37v:
> Great that you got this working with the Quadspi peripheral as well!
> 
> I would say that the `read` and `write` functions can remain mostly unchanged, since they provide a simple API for the majority of non-memory ICs. The `read_extended` and `write_extended` methods add extra functionality that it would be useful to have in the HAL too imo.

would it make sense to refactor the `read` and `write` to simply call the `*_extended` methods so that there is less maintenance to consider?

> 
> You're right that for the Quadspi peripheral the instruction is always a `u8` (for the Octospi peripheral it can be larger). If there's a nice way to simply the API using feature gates then that would be nice to see!

I'll look into that, I was thinking that it could just be a feature gated `pub type XSPIAddr = ..` or something..

richardeoin:
> would it make sense to refactor the read and write to simply call the *_extended methods so that there is less maintenance to consider?

The read and write methods already call the `begin_read` and `begin_write` methods, so I don't think there's much to be achieved by refactoring them to call something else. If anything, the common code that actually transfers the data could be refactored into a common method.

Something like: 
```rust
fn transfer_read_bytes(dest: &mut [u8]) -> Result<(), XspiError> {
    // Wait for the transaction to complete
    while self.rb.sr.read().tcf().bit_is_clear() {}

    // Check for underflow on the FIFO.
    if (self.rb.sr.read().flevel().bits() as usize) < dest.len() {
        return Err(XspiError::Underflow);
    }

    // Read data from the FIFO in a byte-wise manner.
    unsafe {
        for location in dest {
            *location =
                ptr::read_volatile(&self.rb.dr as *const _ as *const u8);
        }
    }

    // Wait for the peripheral to indicate it is no longer busy.
    while self.is_busy().is_err() {}
}
```

phoracek:
Thank you for working on this, this is awesome. Reading the ID worked on my DaisySeed as well. @x37v do you by any chance have a read/write example for the flash on DaisySeed? I'm trying to translate [Xilinx driver](https://github.com/Xilinx/embeddedsw/blob/master/XilinxProcessorIPLib/drivers/qspips/examples/xqspips_flash_intr_example.c) to rust, but I struggle. In any case, thanks!

richardeoin:
Without #235 it's not possible to build or test the OCTOSPI support in this PR. For that reason it's blocked for now, but if it really becomes an issue then we can look at splitting up this PR (or just accepting some code with no CI coverage on master for a bit)

richardeoin:
Rebased onto master

PR #231, opened 2021-06-25
Remove explicit invalidate_icache from examples

adamgreig:
The cortex-m [enable_icache](https://github.com/rust-embedded/cortex-m/blob/master/src/peripheral/scb.rs#L328) method already invalidates before enabling (and is guaranteed to do so in the docs too).


richardeoin:
Thanks!

bors r+

PR #233, opened 2021-07-20
v0.10.0

richardeoin:


hargoniX:
I'm guessing #172 would be the only PR that is left for this release? But it seems like the author has abandoned it quite some time ago so we might as well go for it.

ryan-summers:
> It would be a nice thing if we would finally fix all these clippy lints from our CI though, they always clutter the diffs a ton.

 +1 from me. We should elevate clippy warnings to errors and force them to be fixed through CI in the future. If you really don't want a check, you can reason why and add a `#[cfg()]` instead to disable the check at that point.

Other than that, the release sounds good to me. I've been working off `master` for quite some time without issue.

ryan-summers (approved):

hargoniX:
If we don't want to tackle this in this PR I'd approve as well...do we?

mlamoore:
That definitely sounds like a much bigger separate PR to me...

hargoniX (approved):

richardeoin:
I definitely agree with keeping the clippy warnings under control! #224 removed a few, and #225 will tackle some more. In fact, I only left #225 unmerged so that we could have one last release on the old MSRV, otherwise it's ready for review.

Anyhow I'll make the release now, thanks all!

PR #234, opened 2021-07-22
Removing various clippy lints

ryan-summers:
This PR removes a variety of clippy lints.

Notable changes:
* Time conversion now implements more traits (preferring From over Into)
* I2S constructor now uses a combination type to take less arguments
* Lint warnings are now elevated to failure in CI

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -19,4 +19,4 @@ jobs:
       - uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --examples --target thumbv7em-none-eabihf --features=rt,stm32h743v
+          args: --examples --target thumbv7em-none-eabihf --features=rt,stm32h743v -- -D warnings
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
 ## [Unreleased]
+* Fixed clippy lints:
+   * Added safety docs for some DMA functions
+   * Implemented additional conversion utilities for `time`
+   * Changed I2S constructors to take less arguments
 
 ## [v0.10.0] 2021-07-xx
 
--- a/examples/sai-i2s-passthru.rs
+++ b/examples/sai-i2s-passthru.rs
@@ -16,8 +16,8 @@ pub use stm32h7xx_hal::hal::digital::v2::OutputPin;
 use stm32h7xx_hal::prelude::*;
 use stm32h7xx_hal::rcc::rec::Sai1ClkSel;
 use stm32h7xx_hal::sai::{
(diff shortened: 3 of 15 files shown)


richardeoin:
It seems that the clippy in the CI needs `-A clippy::nursery -A clippy::pedantic -A clippy::restriction` to switch off the `allow` level warnings that we don't want. I'd suggest adding that to the invocation.

ryan-summers:
> It seems that the clippy in the CI needs `-A clippy::nursery -A clippy::pedantic -A clippy::restriction` to switch off the `allow` level warnings that we don't want. I'd suggest adding that to the invocation.

I'm not actually convinced we want all of these flags - I think the default clippy lints are likely a good balance for now, but I'm open to feedback here. Sorry for the delay answering here, I read about this a while back and then forgot to report back

richardeoin:
I agree that default clippy lints are a good balance. The flags I suggested set those defaults, as described in the [table in the clippy docs](https://github.com/rust-lang/rust-clippy/#clippy). 

If we use the CI to deny lints from the `nursery`, `pedantic` and `restriction` groups, then a contributor trying to test their branch locally will have to turn on these lints explicitly. In particular I think the CI should not fail on `restriction` lints, because the clippy docs advise against blanket use of these lints:
> These should be used very selectively, if at all.

ryan-summers:
The deault clippy-lint level of those categories is `allow`, so they are off by-default - I was confused by your comments because of this. I thought you wanted these categories enabled:
image
image

richardeoin:
I'm still confused by the behaviour of the CI, but let's continue anyhow. I think we're both in agreement that lints in the `clippy::pedantic` group should not block merging.

bors r+

richardeoin:
I tried a PR that raises a `clippy::pedantic` lint here https://github.com/stm32-rs/stm32h7xx-hal/pull/239 (in fact it just reverts your latest commit d87a63a).

The CI didn't reject it, which is the expected outcome 🎉 . So it appears there's no issue here. Thanks for all the fixes!

richardeoin:
Ok, I've found the explaination for this. The `inconsistent_struct_constructor` was a `clippy::style` lint in Rust 1.53.0, so got flagged as such in the CI run on 22nd July. By the time I came to review this PR on 2nd August, Rust 1.54.0 had been released and this lint was re-categorised as `clippy::pedantic`. Thus I thought the github CI was flagging `clippy::pedantic` lints and needed some extra flags to prevent that.

It's also why you probably didn't spot that one locally (I guess using something other than latest stable 😄 ), and explains that d87a63a wasn't really nessesary (you could have just said bors try or something).

ryan-summers:
Ah, that is indeed confusing! That lint wasn't there when I originally
wrote up the PR a few weeks back (or so I thought?). Hopefully all is
resolved though :) I'm probably using 1.53 or so locally, although I don't
recall exactly

PR #236, opened 2021-07-24
Migrate to Rust 2021 as part of the public testing

hargoniX:
Migration to Rust 2021 as part of the public testing.

We did actually catch one ICE bug with this \o/: https://github.com/rust-lang/rust/issues/87426


richardeoin:
bors try

richardeoin:
Updating the MSRV to 1.56.1 for #236 #301 #304

richardeoin:
bors retry

PR #242, opened 2021-09-02
Fix inverted UART parity configuration

### LINKED ISSUE #241: USART parity configuration is inverted
In `serial.rs`:
```rust
                            .variant(match config.parity {
                                Parity::ParityOdd => PS::EVEN,
                                _ => PS::ODD,
                            })
```
This causes the incorrect behavior in practice.

starthal:
Fixes #241


richardeoin:
Thank you for the fix!

bors r+

PR #244, opened 2021-09-03
Adding implementation of InputPin trait for Alternate function GPIO pins

sephamorr:
This PR implements the InputPin trait for Alternate-mode GPIO pins, allowing them to be read. This is just copying the implementation from Output and Input pins. There are no side effects to this atomic read, so this is safe to add, regardless of the alternate use of the pin.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -267,6 +267,20 @@ macro_rules! gpio {
                 }
             }
 
+            impl<MODE> InputPin for $PXx<Alternate<MODE>> {
+                type Error = Never;
+
+                fn is_high(&self) -> Result<bool, Never> {
+                    self.is_low().map(|v| !v)
+                }
+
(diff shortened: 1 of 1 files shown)


hargoniX:
Does this actually work out for all AF's and all pins like the trait implementation claims? I'd imagine there is a reason this wasnt added in the first place.

sephamorr:
Per the reference manual (RM0433), section 11.3.12 (Alternate function configuration), "a read access to the input data register gets the I/O state". This is the same language used in 11.3.9 (Input configuration).
image

sephamorr:
Made a change; the trait needed to be implemented for the pin-erased and non-pin-erased variants. I hadn't tested both earlier.

richardeoin:
Looks good! Thanks for adding this and testing it @sephamorr 

bors r+

PR #247, opened 2021-09-05
Make ethernet::DesRing generic over the number of buffers

richardeoin:
This allows users to decide how many transmit and receive buffers they wish to have the the `ethernet::DesRing`. Typically this allows a trade-off in performance vs memory usage.

The number of buffers need to be annotated on both `ethernet::DesRing` and `ethernet::EthernetDMA`. Requiring it on both is a slight usability downside. However this is small compared to the previous solution, which was making a local copy of the HAL in order to edit a constant value in `src/ethernet/eth.rs`.


richardeoin:
Also needed to update example `ethernet-rtic-stm32h735g-dk`

bors r+

PR #248, opened 2021-09-06
Make more CRC methods const

AlyoshaVasilieva:
The only one that has any significant effect is Config's `reflect` since it's now usable when you're making a const Config.


richardeoin:
Thanks, looks good!

bors r+

PR #249, opened 2021-09-07
Ethernet: Disable RA in filtering, fix MAC address setting

adamgreig:
The MACA0LR/HR registers hold the address the MAC can use for filtering received packets, but they're only loaded into the MAC core when LR is written to. Currently the ethernet driver writes to LR first and then HR, so the MAC core doesn't get the bits from HR, which breaks filtering. I believe that's why the RA (receive-all) bit has been set, which bypasses the filtering.

This PR fixes writing the MAC address and disables the no-longer-required RA bit.

I've tested this with an H745 nucleo board and the ethernet-rtic-stm32h747i-disco example, changing just the GPIO setup to match the Nucleo.

See also https://github.com/embassy-rs/embassy/pull/389.

### DIFF
--- a/src/ethernet/eth.rs
+++ b/src/ethernet/eth.rs
@@ -478,7 +478,14 @@ pub unsafe fn new_unchecked<'a>(
                 .dcrcc()
                 .clear_bit()
         });
-        // Set the MAC address
+        // Set the MAC address.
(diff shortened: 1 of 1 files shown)


richardeoin:
LGTM, thanks for testing it out also! Agreed this write order for LR/HR is correct and that RA off is a good default configuration

bors r+

PR #250, opened 2021-09-08
Allow parallel conversion on multiple ADCs

phoracek:
This patch splits the current `convert()` method of ADC in two public methods: `start_conversion()` and `read_sample()`. With this, it is now possible to start conversion on multiple ADCs at the same time and then collect results.

There is one issue I'm not sure how to solve - I can call `start_conversion` on one channel and then `read_sample` on another and it would successfully read something. I suppose it would be better to protect user by exploding in case two channels are mixed. Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 * MSRV increased to 1.51.0
 * **Breaking**: Simplified API for reading device signature
   values. `VAL::get().read()` becomes `VAL::read()`
+* adc: Allow parallel execution of multiple ADCs through `start_conversion()`
 
 ## [v0.10.0] 2021-07-xx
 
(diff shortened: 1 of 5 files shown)


richardeoin:
This is a nice improvement for the ADC support, thanks!

There is definitely potential for confusion when supplying `pin` as an argument to both methods. One simple solution would be to store the current channel as an `Option<u8>` in the [Adc struct](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/adc.rs#L69-L72). Then there's no extra arguments needed for `read_sample`.

phoracek:
@richardeoin right, I have updated that.

richardeoin:
Looks good, thanks!

bors r+

PR #253, opened 2021-09-24
Mark change to I2S constructor as breaking

phoracek:
This change has broken all existing calls to the constructor.

https://github.com/stm32-rs/stm32h7xx-hal/commit/6768a6f4ba9b2abf477851996679819256945593

Signed-off-by: Petr Horacek <hrck@protonmail.com>


richardeoin:
Thanks!

bors r+

PR #256, opened 2021-10-06
Implement IoPin

olback:
Implements the IoPin trait from embedded-hal.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -771,11 +771,11 @@ macro_rules! gpio {
                 for $PXi<Output<OpenDrain>>
                 {
                     type Error = Never;
-                    fn into_input_pin(self) -> Result<Self, Self::Error> {
+                    fn into_input_pin(self) -> Result<Self, Never> {
                         Ok(self)
                     }
-                    fn into_output_pin(mut self, state: PinState) -> Result<Self, Self::Error> {
-                        self.set_state(state)?;
+                    fn into_output_pin(mut self, state: PinState) -> Result<Self, Never> {
+                        self.set_state(state).unwrap(); // Infallible
                         Ok(self)
                     }
                 }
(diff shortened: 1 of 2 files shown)


src/gpio.rs
```diff
@@ -761,6 +761,46 @@ macro_rules! gpio {
...
 
+                impl IoPin<Self, Self>
+                for $PXi<Output<OpenDrain>>
+                {
+                    type Error = Never;
+                    fn into_input_pin(self) -> Result<Self, Self::Error> {
+                        Ok(self)
+                    }
+                    fn into_output_pin(mut self, state: PinState) -> Result<Self, Self::Error> {
+                        self.set_state(state)?;
```
eldruin, line 772: I see here the problem that if setting the state fails, the pin is dropped.

richardeoin (commented):
I'd not seen the `IoPin` trait yet, but it's certainly interesting.

This PR should also bump the `embedded-hal` version in `Cargo.toml` to [0.2.6](https://github.com/rust-embedded/embedded-hal/blob/v0.2.x/CHANGELOG.md), since 0.2.{4,5} would no longer be supported.

olback:
> This PR should also bump the embedded-hal...

Should the other outdated dependencies also be updated?

src/gpio.rs
richardeoin: The `set_low` and `set_high` methods implemented here are infallible (use the `Never` type as an error), so [`set_state`](https://docs.rs/embedded-hal/0.2.6/embedded_hal/digital/v2/trait.OutputPin.html#method.set_state)  would also be infallible. 

Maybe we should implement our own infallible `set_state` (as [stm32f4xx-hal does](https://github.com/stm32-rs/stm32f4xx-hal/blob/f80106fa9e41c2180cd694d3b17b97198c7d4d99/src/gpio.rs#L431)) and use that?

richardeoin:
No, the issue is that with this change the crate would no longer compile with `embedded-hal` 0.2.4. Therefore it's necessary to increase the minimum patch version (3rd number) so Cargo explicitly knows it can't use 0.2.4 but has to download 0.2.6 (or 0.2.7, 0.2.8 ...). For other dependencies that have a new major or minor version, it's good to update them but probably in a separate PR. 

If there's a good solution to avoid ever returning `Err` from the `into_output_pin` methods, then I think we can get this merged

olback:
Would that really be necessary though? `set_state` never fails, it's just the trait impl that requires a `Result`.

richardeoin:
Hmm, yes there is the `type Error = Never;` in the implementation. I would still replace `Self::Error` with `Never` for consistency with every other implementation in this file, and even replace the `?` with an `.unwrap()` to make completely clear that this never fails.

olback:
Somewhat unrelated, is there a reason for using `enum Never {}` over `core::convert::Infallible`?

richardeoin:
Looks good!

My best guess is that `enum Never` was the solution before `convert::Infalliable` was stabilised, and there was an expectation that `convert::Infalliable` would soon be [replaced with the actual `!` never type](https://doc.rust-lang.org/core/convert/enum.Infallible.html#future-compatibility). However stabilisation of `!` seems to have hit some roadblocks, and `enum Never` works ok, so it's not been changed.

PR #257, opened 2021-10-07
Support cortex-m-rt 0.7.0

mattico:
Tested working with both cortex-m-rt 0.6.15 and 0.7.0.

Also: couldn't cortex-m-rt be optional, enabled by the `rt` feature? We shouldn't disable it now, as some might be relying on it, but for the next release, perhaps.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,6 @@ targets = ["thumbv7em-none-eabihf"]
 embedded-hal = "0.2.4"
 embedded-dma = "0.1.2"
 cortex-m = "^0.7.1"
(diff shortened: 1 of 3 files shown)


richardeoin:
This seems reasonable, but your comment prompted me to look a bit closer. Actually it seems this crate shouldn't depend on `cortex-m-rt` at all. If the `rt` feature is specified it will be included in the build tree by the PAC, otherwise not. And we never need to use it from the HAL.

It is needed here as a dev-dependency to build the examples, and your version specifier would be good for that. I'd say we can move it to a dev-dependency on master already?

PR #260, opened 2021-10-08
add spi traits

andrewgazelka:
# Background

SPIs didn't have proper traits, so it was hard/impossible to be generic overall SPis.

# Description

- Added `SpiEnabledExt` for enabled SPIs
- Added `SpiDisabledExt` for disabled SPIs
- Added `SpiAllExt` which `SpiEnabledExt` and `SpiDisabledExt` must implement.

# Questions
- These changes are breaking as people previously using the Spi will need to `use Spi{Enabled,Disabled,All}Ext`. We could also make the traits delegate to the previously existing implementations, but this might not be as clean in terms of hal code imo—and the version is `< 1.x.y` anyway. Thoughts?

### DIFF
--- a/src/dma/macros.rs
+++ b/src/dma/macros.rs
@@ -125,7 +125,7 @@ macro_rules! peripheral_target_instance {
             unsafe impl TargetAddress<M2P> for spi::Spi<$peripheral, spi::Disabled, $size> {
                 #[inline(always)]
                 fn address(&self) -> usize {
-                    use spi::SpiAllExt;
+                    use spi::SpiAll;
                     &self.inner().$txreg as *const _ as usize
                 }
 
@@ -137,7 +137,7 @@ macro_rules! peripheral_target_instance {
             unsafe impl TargetAddress<P2M> for spi::Spi<$peripheral, spi::Disabled, $size> {
                 #[inline(always)]
                 fn address(&self) -> usize {
-                    use spi::SpiAllExt;
+                    use spi::SpiAll;
                     &self.inner().$rxreg as *const _ as usize
                 }
 
--- a/src/dma/macros.rs
+++ b/src/dma/macros.rs
@@ -125,6 +125,7 @@ macro_rules! peripheral_target_instance {
             unsafe impl TargetAddress<M2P> for spi::Spi<$peripheral, spi::Disabled, $size> {
                 #[inline(always)]
                 fn address(&self) -> usize {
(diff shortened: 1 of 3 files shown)


richardeoin:
Hmm we've not used this level of abstraction before, since usually it's enough to either use the embedded-hal traits (`FullDuplex` etc) or use a concrete type in the application code.

The Extension `Ext` traits are usually implemented for the PAC types and them imported in the prelude, which then allows calling `.spi()` or `.spi_unchecked()` on the PAC types. Compared to this, using `Ext` in the names of these new traits is a bit confusing, since these are actually implemented on the HAL types.

I do see the reason for implementing broader traits than embedded-hal provides though, and a breaking change that requires users to import these traits is ok.

andrewgazelka:
> Hmm we've not used this level of abstraction before, since usually it's enough to either use the embedded-hal traits (`FullDuplex` etc) or use a concrete type in the application code.

Most of the time `FullDuplex` has been enough. However, I was recently in a situation where we were using the SPI's DMA, which requires enabling and disabling the SPI. These traits allowed me to do this over any SPI type. However, currently, DMA still requires modifying SPI registers, so I wasn't able to fully utilize the traits. 
 
> The Extension `Ext` traits are usually implemented for the PAC types and them imported in the prelude, which then allows calling `.spi()` or `.spi_unchecked()` on the PAC types. Compared to this, using `Ext` in the names of these new traits is a bit confusing, since these are actually implemented on the HAL types.

This makes sense, I will rename it. Do you think calling it `Spi{Disabled,Enabled,All}` makes the most sense?

> I do see the reason for implementing broader traits than embedded-hal provides though, and a breaking change that requires users to import these traits is ok.

I could also try adding these traits to `embedded-hal` if you think this makes more sense.

mlamoore:
I thought Ext traits like SpiExt were a general naming convention for
traits used to add features/functionality to an external type. So we would
name traits Ext when adding functions to u32, PAC types, embedded-hal
traits, or anything not in our crate, but to add methods to our own types
internal to stm32h7xx-hal we would use more descriptive names.
Spi{Disabled,Enabled,All} sounds good to me.

andrewgazelka:
I don't get any issues when I run cross locally on `v1.56.1`, but for some reason the CI thinks there is a circular dependency.

richardeoin:
Thanks, this looks good now. Let's try it

bors r+

PR #262, opened 2021-10-12
Add glue to make embedded-sdmmc work nicely

olback:
It would have been nice to implement `embedded:BlockDevice` directly on `Sdmmc<$SDMMCX>` but that is not possible since `BlockDevice` takes `&self` and the methods that already exist take `&mut self`.


richardeoin:
Thanks, looks good to me! 

bors r+

PR #268, opened 2021-10-30
Add support for UART9 and USART10

richardeoin:


richardeoin:
Closed in favour of #370

PR #269, opened 2021-10-31
Utilise the SPI FIFO in the Transfer and Write implementations

richardeoin:


robamu:
Thanks for this feature. I actually implemented a HAL for an SPI peripheral with a FIFO as well and this helped a lot.

PR #272, opened 2021-11-01
Example README update

robamu:
I think it makes more sense to give users instructions on how to  build the example applications in their own binary crate. This is how some other HAL crates did it. This PR restructures the `README.md` file a bit and provides complete instructions on how to create a new binary crate and build the blinky example in it.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,6 +1,66 @@
 Examples
 ======
 
+## Parts with a SMPS
+
+For these parts, the program needs to know more about the power supply
+scheme in order to successfully transition from Run* mode to Run mode. For
+an explaination of Run* mode, see RM0433 Rev 7 Section 6.6.1 "System/D3
+domain modes".
+
+For your own code, see the
+[documentation](https://docs.rs/stm32h7xx-hal/latest/stm32h7xx_hal/pwr/index.html#smps)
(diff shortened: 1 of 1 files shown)


richardeoin:
Instructions for creating a binary crate are useful to have, especially if this is the first Rust HAL crate that someone is using. It would be nice to make the README shorter not longer, but that's somewhat unavoidable. Are there any steps that can be removed? Maybe the bit about the examples folder? I see some other guides [recommend copying to main.rs](https://github.com/stm32-rs/stm32f1xx-hal#setting-up-your-project) instead - which might be better.

I would also not recommend mentioning 'openocd' since that project a common source of bugs/issues. Of course you can use it, but it's better to direct new users to probe-rs or pyocd perhaps.

robamu:
yeah, copying to examples would be possible as well, I just need to test whether copying the `utilities`  folder to `src` works in a similar way because this is the dependency of most examples.

I can remove the `openocd` part or make it more generic

robamu:
I simplified the README a bit and removed a few steps.

richardeoin:
I re-wrote this until I was happy with it. I also removed the list of version numbers since these will only become inaccurate over time.

I think it's a useful guide that will help new users avoid frustrating mistakes (like forgetting memory.x). Thanks for the contribution @robamu !

PR #273, opened 2021-11-01
update MSRV to 1.52.0

andrewgazelka:
- There is a compiler bug in 1.51.0 that blocks #260. This is fixed in
  1.52.0


mattico (commented):
We could add a badge to the README as well:

```markdown
![Minimum rustc version](https://img.shields.io/badge/rustc-1.52.0+-yellow.svg)
```

andrewgazelka:
> We could add a badge to the README as well:
> 
> ```gfm
> Minimum rustc version
> ```
> 
> Minimum rustc version

I added this. I left it without a link as I don't know what URL we would have (or if one is even needed).

richardeoin:
It seems it might be possible to solve `error[E0391]: cycle detected when computing the supertraits of...` by [being more specific about the associated type](https://stackoverflow.com/questions/61060677/how-to-solve-cycle-detected-when-computing-the-supertrait). But the MSRV bump is reasonable so let's do it anyhow :)

bors r+

PR #277, opened 2021-11-11
FDCAN support using the fdcan crate

richardeoin:


yorickdewid:
Why has this not been merged yet? The crate works and the tests pass

richardeoin:
It's waiting for the stm32-rs/fdcan crate to be released on crates.io. I tagged it on the next milestone so it won't be forgotten!

yorickdewid:
Ah. Thanks. Just let me know if I can help out

richardeoin:
Rebased onto master and use stm32-rs/fdcan from crates.io

PR #279, opened 2021-11-25
Set ulpi pin speed to VeryHigh

olback:
I don't know if there are any downsides to automatically setting the pin speed to VeryHigh apart from power consumption.


richardeoin:
Thanks, LGTM! Using an unnecessarily high pin speed (more likely to be called "slew rate" in EE terms) can also cause signal integrity and EMC problems, but it depends on the PCB design also. I suspect "VeryHigh" will be the right choice for most people, and anyone wanting something different can use the `new_unchecked` method. Could you add a note to the doc comment with the suggestion to use `new_unchecked` if someone wants a different pin speed for some reason?

PR #282, opened 2021-11-26
improve spi traits

andrewgazelka:
- rename to SPI traits to possibly better names
  -  `SpiAll` -> `HalSpi`
  - `SpiEnabled` -> `HalEnabledSpi`
  - `SpiDisabled` -> `HalDisabledSpi`
 
- `reset` no longer needs to take ownership of the given SPI peripheral

### DIFF
--- a/src/dma/macros.rs
+++ b/src/dma/macros.rs
@@ -125,7 +125,7 @@ macro_rules! peripheral_target_instance {
             unsafe impl TargetAddress<M2P> for spi::Spi<$peripheral, spi::Disabled, $size> {
                 #[inline(always)]
                 fn address(&self) -> usize {
-                    use spi::SpiAll;
+                    use spi::HalSpi;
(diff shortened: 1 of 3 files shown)


richardeoin:
The rename seems fine, thanks! I agree that the reset method should use a `&mut self` receiver, but your implementation seems a bit extreme. My suggestion would be instead:
* Move the implementation from the trait to the ordinary API, so that the reset method is available even when not using the traits
* Add a private method like `fn internal_disable(&mut self)` that contains the register writes, and use that method in both the `pub fn disable(self)` and `pub fn reset(&mut self)` methods. If you like, private methods can be implemented for both typestates (inside `impl<EN> Spi<$SPIX, EN, $TY> { ... }`) whilst keeping the external typestate API.

richardeoin (approved):
Looks good now, thanks!

PR #284, opened 2021-12-02
Impl rand_core::RngCore & support all integers in Rng

olback:


src/rng.rs
```diff
@@ -131,8 +131,104 @@ macro_rules! rng_core {
...
+macro_rules! rng_core_large {
+    ($($type:ty),+) => {
+        $(
+            impl RngCore<$type> for Rng {
+                fn gen(&mut self) -> Result<$type, ErrorKind> {
+                    const WORDS: usize = mem::size_of::<$type>() / mem::size_of::<u32>();
+                    let mut res: $type = 0;
+
+                    for i in 0..WORDS {
+                        res |= (self.value()? as $type) << (i * (mem::size_of::<u32>() * 8))
```
olback, line 144: https://github.com/rust-lang/rust/pull/76492

src/rng.rs
```diff
@@ -131,8 +131,104 @@ macro_rules! rng_core {
...
+                    }
+
+                    Ok(res)
+                }
+
+                fn fill(&mut self, dest: &mut [$type]) -> Result<(), ErrorKind> {
+                    let len = dest.len() * (mem::size_of::<$type>() / mem::size_of::<u32>());
+                    let ptr = dest.as_mut_ptr() as *mut u32;
+                    let slice_u32 = unsafe { core::slice::from_raw_parts_mut(ptr, len) };
```
mattico, line 153: Is alignment an issue?

src/rng.rs
olback: Nvm, requires rust 1.53.0 and bumping the MSRV just for this seems unnecessary.

src/rng.rs
```diff
@@ -131,8 +131,104 @@ macro_rules! rng_core {
...
+        self.fill(dest).unwrap()
+    }
+
+    /// Try to fill a slice with random data. Return an error if RNG fails.
+    fn try_fill_bytes(
+        &mut self,
+        dest: &mut [u8],
+    ) -> Result<(), rand_core::Error> {
+        self.fill(dest).map_err(|e| {
+            core::num::NonZeroU32::new(e as u32)
```
AlyoshaVasilieva, line 225: `rand_core::Error` custom code [should be above `0xC000_0000`](https://docs.rs/rand_core/0.6.3/rand_core/struct.Error.html#associatedconstant.CUSTOM_START)

src/rng.rs
olback:
```
mem::align_of::<u32>() = 4
mem::align_of::<u64>() = 8
mem::align_of::<u128>() = 8
```
I don't think so.  u64/u128 is 8-byte aligned which would be valid for 4-byte aligned values as well. https://doc.rust-lang.org/reference/type-layout.html#size-and-alignment

src/rng.rs
olback: Yup, I totally missed this. Thank you!

src/rng.rs
mattico: Oh right, I hadn't seen the distinction between `rng_core` and `rng_core_large`.

src/rng.rs
```diff
@@ -15,8 +15,8 @@ use crate::time::Hertz;
 
 #[derive(Debug)]
 pub enum ErrorKind {
-    ClockError,
-    SeedError,
+    ClockError = 1,
+    SeedError = 2,
```
richardeoin, line 19: I think technically this is a breaking change, so should go in the CHANGELOG

richardeoin (approved):
Looks good to me! Thanks @olback, and also @AlyoshaVasilieva @mattico for the reviews 🎉

src/rng.rs
olback: Better to keep breaking changes to a minimum. The only reason I did this was because I wanted to be explicit about what `ErrorEnum as u32` would result in. Changing this to 
```
pub enum ErrorKind {
    ClockError = 0,
    SeedError = 1,
}
```
wouldn't be a breaking change but still being explicit.

richardeoin: Ok, great!

richardeoin (approved):
bors r+

PR #287, opened 2021-12-04
Fix the rt feature for stm32h7 0.14.0

newAM:
The stm32h7 crate includes 'rt' as a default feature for 0.14.0: https://docs.rs/crate/stm32h7/0.14.0/source/Cargo.toml


richardeoin:
Upstream change here: https://github.com/stm32-rs/stm32-rs/pull/582

richardeoin:
I suggest we follow the same suggestion at make `rt` a default feature here also. All our CI builds include this feature, so a bit strange for it not be default really

newAM:
> I suggest we follow the same suggestion and make `rt` a default feature here also. All our CI builds include this feature, so a bit strange for it not to be default really

Done! :+1:

richardeoin (approved):
Thanks!

bors r+

PR #289, opened 2021-12-04
Improve error diagnostics when kernel clocks are stopped

richardeoin:
* More helpful panic messages that name the consumer and the stopped clock
* Expose more of these methods in the public API
* Standardise around `kernel_clk`/`kernel_clk_unwrap` naming (excluding timers)


richardeoin:
bors ping

richardeoin:
rebased onto master

PR #295, opened 2021-12-13
Disable ethernet MAC MMC counter interrupts

### LINKED ISSUE #275: Ethernet interrupts can cause application hanging
# Summary
There is a by-default-on interrupt in the Ethernet MAC (MMC) that occurs after a specific number of packets have been sent. If the ethernet ISR is enabled when this interrupt occurs, the ethernet ISR will hang, as the status flag is never cleared.

# Background
The Ethernet MAC MMC contains a number of counters that count MAC-related statistics like the number of ethernet frames sent. When these counters reach 1/2 full (2^31 packets) and full (2^32 packets), an interrupt flag is set.

There is a mask register to disable these counter interrupts, but the mask is reset-0, which means these interrupts are **not** masked by default (thus, they are enabled).

The failure mode here is that an application can appear to hang up after a number of days of heavy ethernet traffic.

Refer to https://github.com/quartiq/stabilizer/issues/450 for further information.

richardeoin:
Patch from https://github.com/quartiq/stabilizer/pull/462

Tested that the expected registers are indeed set on a STM32H747I-DISCO

> MMC_RX_INTERRUPT_MASK:               0x0C020060  MMC Rx interrupt mask register
> MMC_TX_INTERRUPT_MASK:               0x0C20C000  MMC Tx interrupt mask register

Closes #275

### DIFF
--- a/src/ethernet/eth.rs
+++ b/src/ethernet/eth.rs
@@ -428,6 +428,12 @@ pub fn new<'a, const TD: usize, const RD: usize>(
 /// [StationManagement](super::StationManagement) trait. This can be used to
 /// communicate with the external PHY.
 ///
+/// All the documented interrupts in the `MMC_TX_INTERRUPT_MASK` and
+/// `MMC_RX_INTERRUPT_MASK` registers are masked, since these cause unexpected
+/// interrupts after a number of days of heavy ethernet traffic. If these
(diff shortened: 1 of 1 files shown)


jordens (approved):
Thanks a lot!
Checked that this matches the fix what we analyzed and tested extensively.

PR #297, opened 2021-12-15
serial: document parity and word length configuration, set word length based on parity

### LINKED ISSUE #296: Serial parity bits don't work correctly
See this test for example: https://github.com/mattico/stm32h7xx-hal/blob/serial-loopback-test/examples/serial-loopback.rs
Both even and odd parity fail. It also appears that changing the number of stop bits doesn't do anything, but I haven't looked closely.

Tested on a custom board and on an ST NUCLEO-H743ZI2. Tested on USART1 and USART3.

Here's a sigrok pulseview capture of this running (none, odd, then even), rename the .zip to .sr.
[parity_fail.zip](https://github.com/stm32-rs/stm32h7xx-hal/files/7715628/parity_fail.zip)

image

Going to continue investigating, but it's time for dinner.

mattico:
The first commit documents how serial data word length and parity are related.

The second commit removes manual control over serial data word length but configures it automatically based on the parity. This seems reasonable to do since we only support reading/writing 8-bit data.

closes #296

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
   constants to specify how many transmit / receive buffers to include in
   `ethernet::DesRing`. To replicate the previous behaviour, use `DesRing<4, 4>`
 * spi: Utilise FIFO in `Transfer` and `Write` implementations
+* **Breaking**: manual control over serial data length removed, now set automatically based on parity
 
 ## [v0.10.0] 2021-07-xx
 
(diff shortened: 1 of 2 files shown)


richardeoin:
I think the behaviour you're seeing agrees with Table 399 in RM0433 so this was definitely a bug. Agreed with the fix also, it's simple and losing support for 9-bit words is fine.

bors r+

PR #298, opened 2021-12-15
Fix docs.rs quadspi feature

mattico:


richardeoin:
Nicely spotted, thanks!

bors r+

Issue #291, opened 2021-12-09
More memory.x examples for device families


TomDeRybel:
Hello,
I think it would be useful to have a couple complete and documented memory.x examples in the hal crate (or the stm32-rs crate?), ideally one for each of the major device families. As a newbie, that would help.

I've gone ahead and wrote one for the STM32H735 family, attached.
[memory_x.txt](https://github.com/stm32-rs/stm32h7xx-hal/files/7685173/memory_x.txt)

While reading RM0468, it became clear that many of the memory sizes (aside from FLASH) in the example are very different. What is confusing is that the corresponding files generated by CubeMX again show completely different sizes vs what the reference manual suggests. All of this leaves me rather stumped, and I could not find much in the way of documentation on how to do this properly.

Hence the attempt and proposal. I am not sure about the alignments, or the sizes, or anything else. I've tested this with some code (not making use of any of defined regions yet), and it compiles & runs.

Any feed-back is welcome. I'll improve the file accordingly and, if there is interest, will submit it as a PR.

richardeoin:
The example memory.x file lives in this crate, so this is the right place to ask! The Reference Manual should be the source of truth for the memory layout, so I wouldn't pay too much attention to CubeMX unless you find additional evidence it matches with reality. 

I think the memory map for the H735 sub-family diverges a bit more from the other three, so I'd definitely support a separate memory.x example with as much documentation as possible. At a quick glance, the attached file looks good!

For consistency it might make sense split up all 4 sub-families. You'd end up with something like:
`memory_742_743_753_750.x`
`memory_745_755_747_757.x`
`memory_7A3_7B3_7B0.x`
`memory_725_735_730.x`

It is possible to use a build.rs script to automatically select the right one when building the examples (see [stm32f7xx_hal](https://github.com/stm32-rs/stm32f7xx-hal) or [stm32l0xx_hal](https://github.com/stm32-rs/stm32l0xx-hal)). Probably that's a good idea, since it means the correct memory layout is being used for CI.

TomDeRybel:
Thank you for the positive reply. I'll look into this later this week and will try to write those four sub-family files to submit them as a combined PR sometime next week.

The automatic build script approach does look elegant, but is above my skill level. (To complicate it even further, there are some register flags that can change the memory lay-out that would likely also need features to set, if going that far with it all.) I'll leave that to the experts.

richardeoin:
[...]

PR #299, opened 2021-12-16
Detailed examples memory.x files for all current STM32H7xx device families

TomDeRybel:
This PR results from the issue report and discussion https://github.com/stm32-rs/stm32h7xx-hal/issues/291
Please inspect it all carefully with an expert eye, as I only somewhat know what I'm doing.

- I am entirely unsure about the itcm and flash2 sections for the double-core parts. This was taken verbatim from @adamgreig examples. Also, the alignment is 4 for itcm in the example, while I think it is usually 8? Except, it seems to copy data from flash, which is 4. I don't quite get this and it would help to add some docs regarding what is actually happening there.
- I've added all the parts to Cargo.toml as well, making a new distinction "q parts" to deal with SMPS variations. This may not be the right thing at all.

In any case, I hope these files are useful and I definitely learned a lot more about these parts by writing them.


TomDeRybel:
I found an issue with the use of REGION_ALIAS to set FLASH and RAM. 
The "flip-link" crate that I use for stack overflow protection does not like it, saying it can't find MEMORY.RAM during the linking stage.

If I disable flip-link in my cargo/config file rustflags, it all builds and works just fine.

Thus: do I forgo this bit of elegance, where MEMORY defines the chip memories and REGION_ALIAS is used to assign things? Or do I/we take it up with the Knurling team to perhaps allow for this in flip-link?

hargoniX:
Maybe we could also have a table in the README with a mapping from chip family to memory.x? Usually people that start their own project go to the HAL and copy the memory.x but now with us having multiple newcomers to the h7 family might be a bit confused at first.

TomDeRybel:
I've added such a table to the README with some explanation on how to use the files. I've also updated the supported devices table as well with the various new part numbers (hope that is allright).

hargoniX:
Alright cool...it would be awesome if we could also get people that have these different devices to actually test the memory.x's before pushing this out so we don't cause any bugs for users. Especially since you remarked you are unsure about some things.

CC @richardeoin @ryan-summers @jordens and @ whoever else might have H7 chips that are more out of the ordinary.

TomDeRybel:
Thank you all for looking into these!

- I'm by now rather convinced that my changes to the Cargo.toml are in error. If so, I'll undo them.
- I've got some tests running using the new memory.x files, running the latest RTIC rc on:
  - NUCLEO-H723ZG board exercising USB (interrupts), GPIO, EXTI, timer 1. 
    Note: this requires some fixes to stm32-rs, PR already submitted (mainly missing interrupts).
  - NUCLEO-H743ZI2 board exercising USB (polled), GPIO.
- However: none of these tests use anything beyond the RAM and FLASH sections, so they are of limited value.

I've taken a look at the flip-link source, as well as the current open issues. The current parser is limited in what it can handle (it explicitly, and exclusively, looks for a RAM entry in the MEMORY section). However, judging by the open issues, there are hints that a new parser is being thought of. Reporting this issue would likely be useful, so it is taken into consideration if/when the parser re-design happens.

jordens:
I am not sure whether these memory.x are generic and complete. At least for our use case they would only be a template. To make use of the ITCM for example, I need the changes below. Also the `INSERT BEFORE .data` absolutely needs to be there.

```patch
--- ../stm32h7xx-hal/memory_743_750_753.x       2021-12-21 13:28:03.338470293 +0100
+++ memory.x    2021-12-23 14:25:06.827958285 +0100
@@ -76,10 +76,13 @@
     . = ALIGN(4);
     } > FLASH2

-  .itcm (NOLOAD) : ALIGN(8) {
+  .itcm : ALIGN(8) {
+    __sitcm = .;
     *(.itcm .itcm.*);
     . = ALIGN(8);
-    } > ITCM
+    __eitcm = .;
+    } > ITCM AT>FLASH
+  __siitcm = LOADADDR(.itcm);

   .axisram (NOLOAD) : ALIGN(8) {
     *(.axisram .axisram.*);
@@ -111,4 +114,4 @@
     . = ALIGN(4);
     } > BSRAM

-};
+} INSERT BEFORE .data;
```

adamgreig:
Presumably you have some extra initialisation routines that use `__sitcm` and `__siitcm` to program it into memory, though, since they're not part of `cortex-m-rt`? In which case why do you need `INSERT BEFORE .data`?

jordens:
Yes. ITCM needs a loader. For that the symbols need to be there. One could argue that since the loader is elsewhere all the details can also be elsewhere. But that makes the definition of the section (not the memory) here somewhat useless as well.
Because of the address pointer handling by the linker, ITCM loadaddr will be at the beginning of the flash without insert before. That collides with for example the vector table. I thought I had mentioned that in https://github.com/rust-embedded/cortex-m-rt/pull/323 but i can't find it now.

adamgreig:
I agree, I don't think there's much benefit to including an ITCM section in these example memory.x without also including a loader and other such details. I think it's fair to say they're often useful as-is and otherwise useful as templates, without an expectation they'd cover every use-case?

Right, I see why you need `INSERT BEFORE .data` for the ITCM section when it's `>ITCM AT>FLASH`, that makes sense. Should only need it for the `sections{}` with ITCM, though, right?

jordens:
Agreed. probably wise to point out in the comments that these memory.x are not complete solutions but templates.
These limitations can apply to all use cases that could involve loading/zeroing mechanisms, e.g. also data, not just the ITCM.

README.md
```diff
@@ -46,8 +46,37 @@ revision (Revision Y) are supported by feature gates without the 'v'
...
+flash memory size as indicated below.
+
+⚠️: If you use [flip-link](https://github.com/knurling-rs/flip-link) for stack
+overflow protection, there is one more change to make. Flip-link does not (yet)
+support the use of region aliases and expects an entry called RAM in the MEMORY
+block. The work-around is to comment-out the "REGION_ALIAS(RAM, DTCM)" line and
+manually substitute the RAM label in the respective MEMORY entry. Eg: replace
+DTCM by RAM.
+
 #### Flash memory size
```
richardeoin, line 80: You can delete this little "Flash memory size" section, I think it's sufficiently covered by the new memory__.x files now

memory_742.x
```diff
@@ -0,0 +1,107 @@
+MEMORY
+{
+  /* This file is intended for parts in the STM32H742/742v families. (RM0433) */
+  /* - FLASH and RAM are mandatory memory sections.                           */
+  /* - The sum of all non-FLASH sections must add to 692K total device RAM.   */
```
richardeoin, line 5: The only family that supports re-assigning memory around the map is STM32H735, so I don't think there's any value in stating this sum for the other families

memory_745_747_755_757.x
```diff
@@ -0,0 +1,131 @@
...
+    *(.axisram .axisram.*);
+    . = ALIGN(8);
+    } > AXISRAM
+    
+  .sram1 (NOLOAD) : ALIGN(4) {
+    *(.sram1 .sram1.*);
+    . = ALIGN(4);
+    } > SRAM1
+    
+  .sram2 (NOLOAD) : ALIGN(4) {
```
richardeoin, line 111: I think it's worth another note here that sram2 is used as the location for the CPU2 stack

memory_743_750_753.x
```diff
@@ -0,0 +1,114 @@
+MEMORY
+{
+  /* This file is intended for parts in the STM32H743/743v/753/753v families (RM0433), */
```
richardeoin, line 3: You can remove the references to the Rev V parts here, the memory layout was the same anyhow

richardeoin (commented):
Even though they won't cover every use case, these are really valuable templates to have for new users. Since they won't be checked by CI, is it possible these could live outside the repo and we link to them instead? In the wiki perhaps, or even something like a blog post? Open to suggestions here. At the same time the `memory.x` we use for builds can be slimmed down to the minimum needed to run the examples.

I think the feature flags for the "q parts" belong in a separate PR. There are a few places in the code where (for example) the `stm32h7b0` flag is used to mask unavailable pins, so those sites would need to be updated also.

memory_745_747_755_757.x
```diff
@@ -0,0 +1,131 @@
...
+    } > FLASH2
+
+  /*.itcm (NOLOAD) : ALIGN(8) { */
+  /*  *(.itcm .itcm.*);         */
+  /*  . = ALIGN(8);             */
+  /*  } > ITCM                  */
+    
+  .itcm : ALIGN(4) {
+    *(.itcm .itcm.*);
+    . = ALIGN(4);
```
olback, line 98: Is this correct? Line 77 says "ITCM, DTCM and AXISRAM connect to a 64-bit wide bus -> align to 8 bytes."

PR #300, opened 2021-12-27
USB example improvements

richardeoin:
- [x] Add notes about internal USB regulator to examples
- [x] Add an interrupt-driven example for USB, maybe using RTIC

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -211,6 +211,10 @@ name = "usb_passthrough"
 required-features = ["usb_hs", "rm0399"]
 
 [[example]]
+name = "usb_phy_serial_interrupt"
+required-features = ["rt", "usb_hs"]
+
+[[example]]
 name = "rtc"
 required-features = ["rt", "rtc"]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -204,7 +204,7 @@ required-features = ["usb_hs"]
 
 [[example]]
 name = "usb_rtic"
-required-features = ["rt", "usb_hs"]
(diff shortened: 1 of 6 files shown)


mattico:
I was going to enable `xcvrdly` with synopsys-usb-otg 0.3.0, background information: 

It should be safe to enable this on all ULPI PHYs: https://github.com/hathach/tinyusb/pull/1195/files. I believe it should be safe to enable when using the internal PHY as well, but I haven't been able to get USB working on my nucleo board to test it.

If we can't just enable xcvrdly when using the internal PHY we should expose it as a feature.

richardeoin:
Sounds good, we should enable this if we can. I have a H747I-DISCO board (with a USB3320C PHY) to test on.

Interesting that you can't get the on board FS PHY working on your nucleo, do you have any idea why?

mattico:
ESD or IO pin current injection maybe. Pretty sure it used to work and that it's not software related.

richardeoin:
@mattico Could we just add [this text](https://github.com/stm32-rs/synopsys-usb-otg#ulpi-transciever-delay) to the [module-level docs](https://docs.rs/stm32h7xx-hal/latest/stm32h7xx_hal/usb_hs/index.html) with an explaination of how to enable `synopsys-usb-otg/xcvrdly` ? Since cargo dependencies are a union through the tree, and this crate doesn't specifically need `xcvrdly`, it seem more cargo-y to let users select it for themselves.

richardeoin:
Rebased onto master and added ULPI Transciever Delay to documentation

PR #303, opened 2021-12-27
Add backup domain voltage regulator control

mattico:
Pretty simple, but it's useful for me.

Checking the RM again the backup regulator is only used to power the backup SRAM, which can also be powered by V_CORE (the default). Junction temperature monitoring and battery voltage monitoring also require the backup voltage regulator to be enabled.


richardeoin:
Look good to me! From reading the RM I see that PC13 is a special backup IO.. TIL

mattico:
> Look good to me! From reading the RM I see that PC13 is a special backup IO.. TIL

Yeah, that can be a real pain if you don't know about the limited drive strength of the backup GPIOs and try to use it for something.

PR #304, opened 2021-12-27
Add optional support for printing some types with defmt

mattico:
> `defmt` ("de format", short for "deferred formatting") is a highly efficient logging framework that targets resource-constrained devices, like microcontrollers.

https://defmt.ferrous-systems.com/

I've been maintaining this support in my fork for some time, it hasn't been burdensome. defmt has become popular enough that I think it is worthwhile to upstream this. Supporting defmt formatting in this library makes it easier to print error messages (don't need to use `defmt::Debug2Format`) and eliminates the code bloat of the `core::fmt` machinery. We could similarly move `Debug` derives behind a feature flag to help people who are trying to eliminate `core::fmt`, but I don't particularly care.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
           - stm32h7b0
           - stm32h735
     env:                        # Peripheral Feature flags
-      FLAGS: rt,xspi,sdmmc,sdmmc-fatfs,fmc,usb_hs,rtc,ethernet,ltdc,crc,rand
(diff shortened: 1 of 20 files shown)


mattico:
Okay, defmt 0.3 crates require edition 2021 - MSRV 1.56.0. I could bump the MSRV, or just change CI to not use defmt 0.3 on 1.52.0.

richardeoin:
Thanks, this is quite nice (although I've not used `defmt` much myself). Could you add a short section in the `lib.rs` docs to help users find this flag? Otherwise LGTM!

I bumped the MSRV in #236 so should be good now

bors delegate=mattico

PR #306, opened 2022-01-04
Add inner and inner_mut functions to peripherals where it's easy to do so

### LINKED ISSUE #232: `inner()` for more peripherals
The `Spi` peripheral has [`Spi::inner()`](https://docs.rs/stm32h7xx-hal/0.9.0/stm32h7xx_hal/spi/struct.Spi.html#method.inner) which is very useful for doing things that the HAL doesn't support yet. It's nicer, safer, and more discoverable than using `unsafe { &*pac::UART8::ptr() }`. We should add it to more peripherals, possibly all of them.

## Open Questions

1. When should we have `&mut`? The `RegisterBlock`'s `Reg`s don't need `&mut`, even for register writes. Traits that `RegisterBlock`s implement like [`TargetAddress`](https://docs.rs/stm32h7xx-hal/0.9.0/stm32h7xx_hal/dma/traits/trait.TargetAddress.html) don't need `&mut` either. However, we may want to use `&mut` to indicate that operations may *semantically* modify the peripheral, even though it is not required for memory safety. The HAL APIs generally use `&mut self` as such to help keep things sane. We could keep both `inner()` and `inner_mut()` and document that `inner_mut()` is useful to get the compiler to ensure exclusive access. We could only have `inner_mut()` to be on the safe side. I'm leaning towards keeping both.
2. Which peripherals `inner()` functions should be `unsafe`. The DMA peripherals are obvious since they can be used to write to arbitrary memory. Are there any others?

mattico:
Where the registers are owned by a single struct and not shared amongst many channels, etc.

Fixes #232

### DIFF
--- a/src/dma/mod.rs
+++ b/src/dma/mod.rs
@@ -814,6 +814,16 @@ where
         }
     }
 
+    /// Returns a reference to the inner peripheral
+    pub fn peripheral(&self) -> &PERIPHERAL {
+        &self.peripheral
+    }
+
+    /// Returns a mutable reference to the inner peripheral
+    pub fn peripheral_mut(&mut self) -> &mut PERIPHERAL {
+        &mut self.peripheral
+    }
+
     /// Clear all interrupts for the DMA stream.
     #[inline(always)]
     pub fn clear_interrupts(&mut self) {


richardeoin (approved):
Agreed these are all reasonable and safe. Some are implemented on types that are typically consumed quickly (like `FMC`) and thus aren't particularly useful, but still ok to include.

I'll leave the PR open for a bit in case anyone else wants to comment

richardeoin:
I'm less keen on those methods to access the peripheral. Whilst I see your argument, I'd like to get more experience with actually using the DMA API before deciding what to do. You're right that `pause()` allows access, but it also mutates the stream state (disables it) so maybe more obvious for users what the outcome will be.

richardeoin:
@mattico Do you mind if I pick only the first commit and merge that for now?

mattico:
@richardeoin fine with me!

PR #307, opened 2022-01-04
DWT::cycle_count requires cortex-m 0.7.4

mattico:


richardeoin:
Good point, thanks!

bors r+

PR #308, opened 2022-01-06
Always enable the embedded-hal/unproven feature

richardeoin:
The crate completely fails to build without this feature, remove it so that trying to build with --no-default-features gives a useful error message


richardeoin:
bors retry

PR #309, opened 2022-01-14
add more traits to SPI error

andrewgazelka:
- Adds `Copy, Clone, PartialEq, Eq`


richardeoin (approved):
LGTM. Eq is reasonable here. Thanks for muting the errant clippy lint

bors r+

PR #310, opened 2022-01-19
impl `Countdown`

andrewgazelka:
Implements a polling version of delay (`CountDown`) in accordance to `embedded_hal`


richardeoin (approved):
Thanks, indeed useful to have `CountDown` available for SysTick as well as the TIMs. You've implemented CountDown for `type Time = u32; //microseconds` rather than `type Time = Hertz;` used for the implementations in timer.rs. That's fine, but I'm interested to know why?

The current wrapper types have some problems (for example this panics `let t: Hertz = 2_000_000.us().into()`), to the extent that I would consider moving over to something like [fugit](https://docs.rs/fugit/latest/fugit/index.html) instead. What are your thoughts?

andrewgazelka:
> Thanks, indeed useful to have `CountDown` available for SysTick as well as the TIMs. You've implemented CountDown for `type Time = u32; //microseconds` rather than `type Time = Hertz;` used for the implementations in timer.rs. That's fine, but I'm interested to know why?

> The current wrapper types have some problems (for example this panics `let t: Hertz = 2_000_000.us().into()`), to the extent that I would consider moving over to something like [fugit](https://docs.rs/fugit/latest/fugit/index.html) instead. What are your thoughts?

Ah, I suppose `Hertz` would work just fine. For me it is more intuitive, however, to specify a duration rather than a frequency.  So I agree—I think `fugit` seems like a good idea.

andrewgazelka:
Not sure if you want another PR, but I needed to make the MSRV 1.57 for the CI to work with `fugit` @richardeoin

richardeoin:
Thanks, this is nice as a first use of fugit. The MSRV bump also good 👍

richardeoin:
I made another issue #312 to track using fugit in more places, which we could try in future

PR #311, opened 2022-01-24
I2S: Actually write both words

DerFetzer:


richardeoin (commented):
Good catch, thanks!

bors r+

richardeoin:
bors retry

Issue #314, opened 2022-02-10
Add divider and mux debug prints to clocking configuration


jordens:
`CoreClocks` provides the resulting frequencies of the clocking configuration. But often that's not sufficient to understand or debug it.
Being able to see the actual divider and clock mux choices would help a lot when trying to match a desired clock configuration with the one the HAL chooses. That can e.g. be addressed with adding a bunch of `debug!()` prints. Currently the PLL and RCC setup routines are silent and partially black boxes.

@SingularitySurfer

richardeoin:
+1 from me also. Printing via the [log crate](https://crates.io/crates/log) expects that users have a working log implementation (for example by copying one from the examples). I think that's reasonable.

PR #318, opened 2022-02-16
RCC debug prints

SingularitySurfer:
I'm thinking about how to implement RCC debug prints https://github.com/stm32-rs/stm32h7xx-hal/issues/314 in the cleanest fashion. 

There is a lot going on in the RCC setup and a complete feedback about all the exact register settings is a substantial amount of information that is not easily made human readable. 
I think what would make sense is to give feedback corresponding to the Clock Configuration in CubeMX:
Screenshot from 2022-02-16 12-36-19
That would be all the MUX settings, prescalers, dividers and the PLL-frac.

I could have one compact codeblock at the end of the freeze function where I read the necessary registers again and give the formatted debug info (currently its just printing raw reg bits). 
https://github.com/stm32-rs/stm32h7xx-hal/blob/d924165915c74f34e8b1610934afc20e133b10e5/src/rcc/mod.rs#L976-L977

The other option would be to sprinkle in the debug statements within the macros that setup PLLs etc. This would not necessitate reading the registers again but the code would be less clean.

Lastly I want to ask if using the log crate like this is OK or if we would rather use [defmt](https://crates.io/crates/defmt)?

### DIFF
--- a/src/rcc/mod.rs
+++ b/src/rcc/mod.rs
@@ -974,9 +974,14 @@ impl Rcc {
         });
         while syscfg.cccsr.read().ready().bit_is_clear() {}
 
+        // This section prints the final register configuration for the main RCC registers:
+        // - System Clock and PLL Source MUX
+        // - PLL configuration
+        // - System Prescalers
+        // Does not include peripheral/MCO/RTC clock MUXes
         #[cfg(feature = "log")]
         {
-            debug!("--- RCC register settings (selection)");
+            debug!("--- RCC register settins");
 
             let cfgr = rcc.cfgr.read();
             debug!(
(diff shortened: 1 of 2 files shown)


mattico:
I would like to be able to disable the logs, at least. Pulling in the `log` crate and all the format machinery can be many kilobytes of code. It would be most convenient to define our own macros that have the `#[cfg]`s inside, so we don't need to have a `#[cfg]` on every log statement. If we're doing that, we may as well support either `log` or `defmt` based on the enabled cargo features, [like embassy does](https://github.com/embassy-rs/embassy/blob/59f909e66533097b56093ec55fdd1341dbff6744/embassy/src/fmt.rs). We should rarely (never?) be formatting third-party types directly, so shouldn't have to deal with them not having defmt support.

mattico:
A stretch goal would be to replace all our `unwrap()`,  `expect()`, and `panic!()`s with a macro that forwards to defmt (when enabled) like embassy does, so that defmt users get nicer panic formatting and to save the space of the formatting machinery and all the panic message strings.

SingularitySurfer:
Another point is that it would be nice to have a debug implementation for registers in general. I fount [this](https://github.com/rust-embedded/svd2rust/issues/48) issue in svd2rust from 2016, where someone is also fighting with printing out PLL configs. That would be a bigger effort yet.

richardeoin:
I'm happy with the approach of using the log crate for this, with defmt support being an added bonus but not required. 

@mattico Interestingly if there's no logging facade implemented by the user (or the examples), then almost nothing from the `log` crate should appear in the final binary. I checked `cargo bloat --example blinky --features=rt,stm32h743v --release` and found it gives an identical result on this branch compared to master

```
rustc 1.59.0-nightly (e012a191d 2022-01-06)
    Finished release [optimized + debuginfo] target(s) in 0.03s
    Analyzing target/thumbv7em-none-eabihf/release/examples/blinky

File  .text   Size         Crate Name
0.3%  70.6%   876B        blinky blinky::__cortex_m_rt_main
0.1%  10.6%   132B stm32h7xx_hal <stm32h7xx_hal::delay::Delay as embedded_hal::blocking::delay::DelayMs<u16>>::delay_ms
0.0%   5.6%    70B   cortex_m_rt Reset
0.0%   5.0%    62B stm32h7xx_hal stm32h7xx_hal::rcc::pll::calc_ck_div
0.0%   1.6%    20B   cortex_m_rt HardFaultTrampoline
0.0%   0.8%    10B           std core::panicking::panic
0.0%   0.8%    10B           std core::panicking::panic_fmt
0.0%   0.8%    10B     [Unknown] main
0.0%   0.5%     6B      cortex_m __primask_r
0.0%   0.3%     4B      cortex_m __cpsie
0.0%   0.3%     4B      cortex_m __cpsid
0.0%   0.2%     2B   cortex_m_rt HardFault_
0.0%   0.2%     2B    panic_halt rust_begin_unwind
0.0%   0.2%     2B   cortex_m_rt DefaultPreInit
0.0%   0.2%     2B   cortex_m_rt DefaultHandler_
0.0%   0.0%     0B               And 0 smaller methods. Use -n N to show more.
0.5% 100.0% 1.2KiB               .text section size, the file size is 253.7KiB
```

mattico:
Sounds good to me!

src/rcc/mod.rs
```diff
@@ -987,67 +992,70 @@ impl Rcc {
...
             let pll1divr = rcc.pll1divr.read();
             debug!(
-            "PLL1DIVR register ->   DIVN1: {:?}  DIVP1: {:?}  DIVQ1: {:?}  DIVR1: {:?}",
-            pll1divr.divn1().bits(),
-            pll1divr.divp1().bits(),
-            pll1divr.divq1().bits(),
-            pll1divr.divr1().bits(),
-        );
+                "PLL1DIVR register ->   DIVN1: {:?}  DIVP1: {:?}  DIVQ1: {:?}  DIVR1: {:?}",
+                pll1divr.divn1().bits() + 1,
```
jordens, line 1054: Adding one is probably not a good idea.

SingularitySurfer: Printing the raw values in hex now.

SingularitySurfer:
I have implemented debug outputs using the log crate for the main RCC clock distribution registers. With this it is relatively easy to check back using CubeMX or to compare to the reference manual. As stated above, I opted to only display a selection that is useful in a general project. Printing all the peripheral clock settings and individual flags would be a lot at this point.
The debug output can be enabled/disabled with the "log" feature. Maybe I should call this "rcc_debug_log" for now since nothing except the examples use the log crate?

richardeoin (approved):
I'm happy with this approach. The "log" feature is good compromise imo. It's annoying to have a feature at all, but some users might want to use the `debug!` log level for something else. Multiple features that add a new layer of flags would be excessive I think.

Perhaps worth a comment in Cargo.toml so maintainers don't miss the dependency below
```toml
log = { version = "0.4.14", optional = true} # see also the dev-dependencies section
```

SingularitySurfer:
Alright, comment added.

jordens (approved):

PR #319, opened 2022-02-17
PD12 is linked to TIM4 via AF2

yorickdewid:
Allow PD12 to be used in TIM4


richardeoin:
Ha quite the coincidence, in any case thanks for the PR @yorickdewid! I just found a couple more missing on TIM15, will follow up in another PR in a sec

bors r+

Issue #281, opened 2021-11-26
[Feature Request] FLASH implementation


ostenning:
A FLASH implementation that allows for unlocking flash and overwriting it would be great for those wanting to write custom bootloaders for their apps

David-OConnor:
https://github.com/David-OConnor/stm32-hal/blob/main/src/flash.rs

Supports H7. Ignore the irrelevant feature flags.

ostenning:
Thanks @David-OConnor, that helped me get started with it.

https://github.com/stm32-rs/stm32h7xx-hal/pull/322

PR #322, opened 2022-02-22
Flash implementation for rm0433

ostenning:
Hello, I've spent some time [implementing flash](https://github.com/stm32-rs/stm32h7xx-hal/issues/281) read/write/erase functions for the stm32h743. 

The current limitations of the implementation are:
 - rm0433 devices supported only
 - "Simple" mode operations only (No secure flash operations)
 - AXI clock speed is not yet configured / optimized

But I think this could be a good start, ideally I'll also get the AXI clock configured but I'm running out of time.

As this is my first contribution I would love any feedback for improvements.

Thanks very much

### DIFF
--- a/examples/flash.rs
+++ b/examples/flash.rs
@@ -13,6 +13,7 @@ use stm32h7xx_hal::{flash::Bank, pac, prelude::*};
 
 #[entry]
 fn main() -> ! {
+    utilities::logger::init();
     info!("stm32h7xx-hal example - erase, write and read FLASH");
 
     let _cp = cortex_m::peripheral::Peripherals::take().unwrap();
--- a/examples/flash.rs
+++ b/examples/flash.rs
@@ -1,7 +1,6 @@
 //! Example of Flash erasing, writing and reading with the stm32h743zi (rm0433)
 //! Assumes this example runs at sector 0 of bank 1 (0x0800_0000)
 
-#![deny(warnings)]
 #![no_main]
 #![no_std]
 
@@ -20,11 +19,11 @@ fn main() -> ! {
     info!("Setup PWR...                  ");
     let pwr = dp.PWR.constrain();
     let pwrcfg = example_power!(pwr).freeze();
-    
+
     info!("");
     info!("stm32h7xx-hal example - FLASH erase, write and read");
     info!("");
-    
+
     // Constrain the flash peripheral,
     // giving access to read/write/erase operations
     let mut flash = dp.FLASH.constrain();
--- a/examples/flash.rs
+++ b/examples/flash.rs
@@ -1,4 +1,5 @@
 //! Example of Flash erasing, writing and reading with the stm32h743zi (rm0433)
+//! Assumes this example runs at sector 0 of bank 1 (0x0800_0000)
 
 #![deny(warnings)]
 #![no_main]
@@ -8,7 +9,7 @@ use cortex_m_rt::entry;
 #[macro_use]
 mod utilities;
 use log::info;
-use stm32h7xx_hal::{pac, prelude::*, flash::Bank};
+use stm32h7xx_hal::{flash::Bank, pac, prelude::*};
 
 #[entry]
 fn main() -> ! {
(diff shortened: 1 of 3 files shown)


src/flash.rs
```diff
@@ -34,3 +49,436 @@ impl ACR {
...
+    fn bank_registers(&self, bank: Bank) -> &BANK {
+        match bank {
+            Bank::UserBank1 => self.regs.bank1(),
+            Bank::UserBank2 => self.regs.bank2(),
+        }
+    }
+
+    fn _configure_clocks(&mut self) {
+        // TODO: Configure clocks for optimal performance for both read and write operations
+
```
richardeoin, line 142: The RCC module already sets the ACR parameters, so this section isn't needed

src/flash.rs
```diff
@@ -34,3 +49,436 @@ impl ACR {
...
+            panic!("Sector out of bounds")
+        }
+        if buff.len() % 4 != 0 || start % 4 != 0 {
+            panic!("buff length and start offset must be a multiple of a 32bit word");
+        }
+
+        let regs = self.bank_registers(bank);
+
+        // Check that no Flash main memory operation is ongoing by checking the BSY bit
+        while regs.sr.read().bsy().bit_is_set() {}
```
richardeoin, line 219: The RM description of the BSY bit refers to "write, erase or option byte" operations, not any operation. Specifically there can be (and likely are) flash read operations ongoing when BSY is clear. Could you update the comment? (also appears in write_sector and erase_sector)

src/flash.rs
```diff
@@ -34,3 +49,436 @@ impl ACR {
...
+
+        unsafe {
+            // Offset it by the start position
+            addr = addr.add(start / 4);
+
+            let chunks = buff.chunks_exact_mut(4);
+
+            for chunk in chunks {
+                if addr as usize > bank.end_address() {
+                    if bank == Bank::UserBank2 {
```
richardeoin, line 242: There's no reason to call read_sector recursively here. Perhaps you can remove that and simplify the code?

richardeoin (commented):
Thank you for the PR! I've made a few small comments, please also see the messages in Github Actions.

It would be great to support flash on more H7 parts in future, but for now you can use a [statement like this ](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/Cargo.toml#L141-L143) in Cargo.toml so that Github Actions only checks the example for RM0433

src/flash.rs
ostenning: Yeah, I wasn't sure if this was the best of ideas. I'll remove it 👍

src/flash.rs
ostenning: Ah thanks @richardeoin I wasn't aware of this. I'll remove this

ostenning:
> Thank you for the PR! I've made a few small comments, please also see the messages in Github Actions.
> 
> It would be great to support flash on more H7 parts in future, but for now you can use a [statement like this ](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/Cargo.toml#L141-L143) in Cargo.toml so that Github Actions only checks the example for RM0433

Thanks for the feedback @richardeoin, I've fixed your suggestions. Let me know what you think, thanks!

examples/flash.rs
```diff
@@ -0,0 +1,63 @@
...
+
+use cortex_m_rt::entry;
+#[macro_use]
+mod utilities;
+use log::info;
+use stm32h7xx_hal::{flash::Bank, pac, prelude::*};
+
+#[entry]
+fn main() -> ! {
+    info!("stm32h7xx-hal example - erase, write and read FLASH");
```
richardeoin, line 16: Add a call to init here
```suggestion
    utilities::logger::init();
    info!("stm32h7xx-hal example - erase, write and read FLASH");
```

examples/flash.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    let _cp = cortex_m::peripheral::Peripherals::take().unwrap();
+    let dp = pac::Peripherals::take().unwrap();
+
```
richardeoin, line 20: For some parts the power peripheral needs to be configured, so the CI requires you have this

```suggestion
    // Constrain and Freeze power
    info!("Setup PWR...                  ");
    let pwr = dp.PWR.constrain();
    let pwrcfg = example_power!(pwr).freeze();
    
```

examples/flash.rs
```diff
@@ -0,0 +1,63 @@
...
+    flash
+        .read_sector(Bank::UserBank2, 0, 256, &mut read)
+        .unwrap();
+    for i in 0..read.len() {
+        assert_eq!(read[i], buff[i]);
+    }
+
+    info!("Successfully erased, written and read back flash data");
+
+    loop {}
```
richardeoin, line 62: You can avoid the clippy lint here by calling `nop()`

```suggestion
    loop {
        cortex_m::asm::nop()
    }
```

richardeoin (commented):
Ok, great! All those changes look good.

I have one extra comment and there are some changes needed so that the example passes CI. If it's not clear have a look at some of the other examples ([this one is good](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/examples/crc.rs))

src/flash.rs
```diff
@@ -2,6 +2,11 @@
```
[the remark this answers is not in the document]
ostenning: Thanks @richardeoin I've fixed those changes you suggested.

I have one question regarding the stm32h750 which is also apart of rm033: Should I also disable this model from this feature? Something like:

```
#[cfg(any(feature = "rm0433"), not(feature = "stm32h750"))]
```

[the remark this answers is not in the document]
richardeoin: Yes that would be sensible, since the implementation doesn't support the stm32h750. you'd also need to narrow the `required-features` in Cargo.toml by using a specific device flag rather than just`rm0433`

richardeoin:
@ostenning The CI is still failing - could you take a look? I'd suggest removing the `#![deny(warnings)]` from the example and running rustfmt again

ostenning:
> @ostenning The CI is still failing - could you take a look? I'd suggest removing the `#![deny(warnings)]` from the example and running rustfmt again

Hey, okay I've fixed the formatting. It seems the `clippy` job is the final failing one because its running on my fork

richardeoin:
Ok, merged. Thanks!

ostenning:
@richardeoin I just fixed the feature flags to exclude `stm32h750` and `stm32h750v`. Shall I create new PR for that?

richardeoin:
@ostenning Yes, sure

PR #324, opened 2022-02-25
serial: support reconfiguring ports, Tx/Rx pin swap, and join() for Tx/Rx

### LINKED ISSUE #323: [Feature Request] Reconfigurable serial ports
I wanted to be able to reconfigure the baud rate on a particular serial port after creation, but I didn't find a way to do it safely. It seems like it may be useful to be able to set a new config after creation, however, I'm not sure how feasible this is.

mattico:
Fixes #323

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -543,10 +543,6 @@ macro_rules! usart {
                     let brr = usartdiv as u16;
                     self.usart.brr.write(|w| { w.brr().bits(brr) });
 
-                    // disable hardware flow control
-                    // TODO enable DMA
-                    // self.usart.cr3.write(|w| w.rtse().clear_bit().ctse().clear_bit());
-
                     // Reset registers to disable advanced USART features
                     self.usart.cr2.reset();
                     self.usart.cr3.reset();
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -134,6 +134,7 @@ pub mod config {
         pub clockphase: ClockPhase,
         pub clockpolarity: ClockPolarity,
         pub lastbitclockpulse: bool,
+        pub swaptxrx: bool,
     }
 
     impl Config {
@@ -150,6 +151,7 @@ pub mod config {
                 clockphase: ClockPhase::First,
                 clockpolarity: ClockPolarity::IdleLow,
                 lastbitclockpulse: false,
+                swaptxrx: false,
             }
         }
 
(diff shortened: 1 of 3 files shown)


mattico:
I've been using this for a while to reconfigure a serial port between a runtime configuration (1M 8N1) to a bootloader configuration (57600 8E1 SwapTxRx) and back.

src/serial.rs
```diff
@@ -505,39 +515,40 @@ macro_rules! usart {
...
+                    serial.configure(&config $(, $synchronous )?);
 
-                    // Enable clock for USART and reset
-                    prec.enable().reset();
+                    Ok(serial)
+                }
 
-                    // Get kernel clock
-	                let usart_ker_ck = Self::kernel_clk_unwrap(clocks).0;
+                fn configure(&mut self, config: &config::Config $(, $synchronous: bool)?) {
```
richardeoin, line 530: This is an internal method, but could you add a comment as a reminder that it can only be called when the peripheral is disabled?

src/serial.rs
```diff
@@ -704,9 +743,19 @@ macro_rules! usart {
...
                             _usart: PhantomData,
+                            ker_ck: self.ker_ck,
                         },
                     )
                 }
+
+                #[allow(unused_variables)]
+                pub fn join(tx: Tx<$USARTX>, rx: Rx<$USARTX>) -> Self {
+                    Self {
+                        usart: unsafe { core::mem::transmute::<(), $USARTX>(()) },
```
richardeoin, line 754: You can also use [mem::zeroed](https://doc.rust-lang.org/stable/std/mem/fn.zeroed.html) here, which gets the job done while being a bit easier to understand

```rust
usart: unsafe { core::mem::zeroed::<$USARTX>() },
```

Even better if you add an assert to validate the assumption that `$USARTX` is a ZST.

```rust
assert_eq!(mem::size_of::<$USARTX>(), 0);
```

richardeoin (commented):
Looking good, just a couple of small comments.

I guess the `merge=union` needs to be on the base branch to work?

mattico:
The `merge=union` seemed to work when I rebased locally, maybe GitHub doesn't use it?

richardeoin (approved):
PR looks good now, thanks!

bors r+

src/serial.rs
mattico: The assert isn't really necessary because of E0512:
```
error[E0512]: cannot transmute between types of different sizes, or dependently-sized types
  --> mainboardfw\src\bootloader\mod.rs:25:17
   |
25 |     let gpiog = transmute::<u32, pac::GPIOG>(0);
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: source type: `u32` (32 bits)
   = note: target type: `GPIOG` (0 bits)

For more information about this error, try `rustc --explain E0512`.
```

Though it'll get optimized out so no big deal.

PR #328, opened 2022-02-27
SPI: Error on DuplexFailed errors 

richardeoin:
Duplex failed errors occour when a word was sent, but no corresponding word was received

* Refactor `byte` -> `word` since this is implemented for 8, 16 and 32 bits
* Make long transactions slightly more efficient by not writing CSTART after the
4th word and only reading SR once per word (rather than twice)
* Fix arithmetic error in spi divider calculation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 * timer: Fix bug in `counter()` TIM1/8/12/13/14 where return value would also
   include UIF bit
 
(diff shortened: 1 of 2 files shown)


mattico (approved):
Tested, my SPI peripherals still worked. Didn't try to test triggering SPI errors deliberately.

richardeoin:
Thanks for testing it out @mattico 👍

Issue #329, opened 2022-03-01
[Feature Request] DMA configuration for SAI peripherals does not include all possible cases


backtail:
I recently came across (in my specific case) a bug. Here, you'll find the details: https://github.com/antoinevg/daisy_bsp/issues/8#issuecomment-1054506275

What it boils down to is that the DMA config for all SAI peripherals are "hardwired" in the hal library.

https://github.com/stm32-rs/stm32h7xx-hal/blob/72adb08daf594d623511d652647c23088b54ef12/src/dma/dma.rs#L938-L941

Channel A is always configured as TX and channel B is always configured as RX even though the individual I²S channels can be configured either way. 

https://github.com/stm32-rs/stm32h7xx-hal/blob/72adb08daf594d623511d652647c23088b54ef12/src/sai/i2s.rs#L73-L76
https://github.com/stm32-rs/stm32h7xx-hal/blob/72adb08daf594d623511d652647c23088b54ef12/src/sai/i2s.rs#L177-L178

This means if one would need to use channel A of the SAI as RX channel and vice versa for channel B, it's currently not possible with the DMA.

richardeoin:
Thanks for reporting! The DMA support in this HAL is still relatively immature, and I think the Rust community is still figuring out how the best approach to DMA. So problems like this one are good to highlight where improvements (or new approaches) are needed.

SAI breaks the assumption that each peripheral has a single address for DMA write and a single address for DMA read. As you've noticed, the current implementation re-imposes that assumption that by fixing one channel as read and one channel as write.

Right now the init call does not specify either Channel A or Channel B. Currently we use type inference to select a single [`TargetAddress` ](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/dma/traits.rs#L393) implementation based on just `SAI1` and `dma::MemoryToPeripheral`, and therefore that implementation is hard-coded as Channel A.
```rust
let dma1_str0: dma::Transfer<_, _, dma::MemoryToPeripheral, _, _> = dma::Transfer::init(
    dma1_streams.0,
    unsafe { pac::Peripherals::steal().SAI1 },
    tx_buffer,
    None,
    dma_config,
);
```

My first idea for fixing this is to implement `TargetAddress` for a compound type that represents both the SAI peripheral and the channel. The implementation would look like this.

```rust
peripheral_target_address!(
[...]

PR #330, opened 2022-03-01
Prototype of implementing SAI DMA on ChannelA/B compound types

richardeoin:
Implement as wrapper type instead, which allows us to pretend this is a smart pointer type and make use of [deref coercion](https://doc.rust-lang.org/std/ops/trait.Deref.html)

This is actually backwards compatible, so the old fixed mapping (write on channel A, read on channel B) still works


richardeoin:
ref #329

backtail (approved):
Looks great to me!

PR #337, opened 2022-03-21
Add inner and inner_mut functions to peripherals where it's easy to d…

richardeoin:
…o so

Where the registers are owned by a single struct and not shared amongst many channels, etc.


richardeoin:
bors retry

PR #339, opened 2022-03-22
Replace crate::Never with core::convert::Infallible

richardeoin:
There's only a few remaining uses of crate::Never


richardeoin:
@mattico - as discussed on #334

richardeoin:
@mattico - as discussed on #334

mattico (approved):

PR #344, opened 2022-03-29
Add slope method to ADC and depreciate max_sample

### LINKED ISSUE #343: ADC example scaling wrong
The ADC examples scale the ADC codes with `VREF/max_sample`.

https://github.com/stm32-rs/stm32h7xx-hal/blob/a46841c4465978aeddef3795b4f571c6c25cfcc9/examples/adc.rs#L76-L80

`max_sample = ((1 << bits) - 1) << lshift` by itself is correct according to its docstring. 

https://github.com/stm32-rs/stm32h7xx-hal/blob/a46841c4465978aeddef3795b4f571c6c25cfcc9/src/adc.rs#L790-L793

But the conversion is not. VREF corresponds to `1 << bits + lshift` here as in all ADCs that I know of.
c.f. https://github.com/stm32-rs/stm32f4xx-hal/issues/362#issuecomment-941014100 and other places.

The best approach is probably to change `max_sample()` and its docstring.

richardeoin:
Closes #343

### DIFF
--- a/examples/adc.rs
+++ b/examples/adc.rs
@@ -76,7 +76,7 @@ fn main() -> ! {
         info!(
             "ADC reading: {}, voltage for nucleo: {}",
             data,
-            data as f32 * (3.3 / adc1.max_sample() as f32)
+            data as f32 * (3.3 / adc1.slope() as f32)
         );
     }
 }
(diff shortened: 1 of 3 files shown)


jordens (approved):
Oh you beat me to it! Thanks LGTM

PR #346, opened 2022-04-05
update GPIO 2

burrbull:


burrbull:
cc @richardeoin

mattico:
I am 👍 on this because I end up switching pins from pull-up to pull-down in certain situations and having the typestate makes this very annoying.

mlamoore:
It looks like you removed pull-up and floating from some examples but never
set the input type elsewhere. I feel like the examples should set the input
type to maintain previous functionality, especially for simple dev board
setups that often rely on the internal pull ups to set signal level of
buttons etc...

burrbull:
> It looks like you removed pull-up and floating from some examples but never set the input type elsewhere.

`into_pull_up_input` never gone. It is just not present in type definition now.

mlamoore:
Oops, sorry about that. I was just looking at the diff which said nothing
about pull-up/pull-down, the initialization didn't change so I didn't
examine closely enough.

Looks great to me!

burrbull:
rebased on #347

burrbull:
rebased

richardeoin:
@burrbull Is this ready to merge?

burrbull:
should be

richardeoin:
Ok thanks, let's merge it

Issue #348, opened 2022-04-06
Question : Configure Rx FIFO trigger for Usart Serial


yann77500:
I begin in Rust,
I can't found how configure Rx fifo for serial on usart.
When I start listen() on RxEvent, it seems interrupt every byte received. 
Is it possible to configure Rx Fifo trigger ?

Thank you for your reply

mattico:
The FIFO level is not currently configurable with the HAL.

It should be easy to add similar to: https://github.com/stm32-rs/stm32h7xx-hal/pull/340/files#diff-fb3a60b3de1be206f6d00d6331d24886f8e145d6091d8d4b4557247b8017bd82

yann77500:
Thank you for your reply, 
I will try to study for add the feature

richardeoin:
Closed by #350

PR #350, opened 2022-04-10
[add] Add serial Rx and Tx fifo capacity and Fifo interrupt event

yann77500:
Related to issue : #348

### DIFF
--- a/src/serial.rs
+++ b/src/serial.rs
@@ -71,10 +71,10 @@ pub mod config {
 
     #[derive(Copy, Clone, PartialEq)]
     pub enum FifoThreshold {
-        OneEighth,
-        OneQuarter,
-        OneHalf,
-        TreeQuarter,
+        Eighth,
+        Quarter,
+        Half,
+        ThreeQuarter,
         SevenEighth,
         Full,
     }
@@ -159,8 +159,8 @@ pub mod config {
                 swaptxrx: false,
                 invertrx: false,
                 inverttx: false,
-                rxfifothreshold: FifoThreshold::OneEighth,
-                txfifothreshold: FifoThreshold::OneEighth,
+                rxfifothreshold: FifoThreshold::Eighth,
+                txfifothreshold: FifoThreshold::Eighth,
             }
         }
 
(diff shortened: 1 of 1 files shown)


src/serial.rs
```diff
@@ -58,11 +58,27 @@ pub enum Event {
...
 }
 
 pub mod config {
     use crate::time::Hertz;
 
+    #[derive(Copy, Clone, PartialEq)]
+    pub enum FifoThreshold {
+        OneEighth,
+        OneQuarter,
+        OneHalf,
```
richardeoin, line 76: For consistency with the DMA API, it's nicer to just say "Quarter", "Half"...

```suggestion
        Eighth,
        Quarter,
        Half,
```

src/serial.rs
```diff
@@ -58,11 +58,27 @@ pub enum Event {
...
+        TreeQuarter,
```
richardeoin, line 77:
```suggestion
        ThreeQuarter,
```

src/serial.rs
```diff
@@ -551,6 +581,37 @@ macro_rules! usart {
...
+                        FifoThreshold::OneQuarter => 1,
+                        FifoThreshold::OneHalf => 2,
+                        FifoThreshold::TreeQuarter => 3,
+                        FifoThreshold::SevenEighth => 4,
+                        FifoThreshold::Full => 5,
+                    };
+
+
+                    unsafe{
+                        self.usart.cr3.write(|w| {w.txftcfg().bits(fifo_threshold_bits) });
```
richardeoin, line 610: You should use the `modify` method instead of `write` here. Otherwise the rxftcfg bit will be overwritten!

To avoid this in general, I would actually suggest using `modify` instead of `write` in both cases.

```suggestion
                        self.usart.cr3.modify(|_, w| {w.txftcfg().bits(fifo_threshold_bits)});
```

richardeoin (commented):
Thank you for the contribution! I've made some comments in the code.

Also, could you run rustfmt just to make the formatting consistent with the rest of the project? You can follow the Quick Start section on the [rustfmt page](https://github.com/rust-lang/rustfmt)

src/serial.rs
yann77500: Oh thanks. Careless mistake
One question more, I begin Rust, could you tell me why on some registers it is mandatory to put code in safe section meanwhile others register can be modify outside unsafe section?

src/serial.rs
yann77500: Ok thanks

src/serial.rs
richardeoin: It's a good question, the answer is in the svd2rust documentation [on this page](https://docs.rs/svd2rust/latest/svd2rust/#write)

> // NOTE ^ unsafe because you could be writing a reserved bit pattern into
// the register. In this case, the SVD doesn't provide enough information to
// check whether that's the case.

So it depends if someone has contributed a definition of the valid (not reserved) bit patterns to the [stm32-rs project](https://github.com/stm32-rs/stm32-rs). If there is a definition then it will be safe, otherwise unsafe is required.

yann77500: Ok thanks you for your complete answer

PR #353, opened 2022-04-16
Re-work DMA internals

richardeoin:
Make less use of macros, and more use of trait bounds. StreamX struct matches f4 API

### DIFF
--- a/src/dma/bdma.rs
+++ b/src/dma/bdma.rs
@@ -3,9 +3,11 @@
 //! For RM0455 parts, only BDMA2 is implemented
 
 use super::{
-    config,
-    traits::sealed::{Bits, Sealed},
-    traits::*,
+    config, traits,
+    traits::{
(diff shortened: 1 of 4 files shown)


richardeoin:
@burrbull This is similar to your PR on the F4 HAL https://github.com/stm32-rs/stm32f4xx-hal/pull/341

Could you review this PR if possible?

burrbull:
Looks good to me.

P.S. I have a counter request https://github.com/stm32-rs/stm32-rs/pull/684 :)

richardeoin:
Found an error after trying this out on hardware, fix pushed now

I responded to your review request @burrbull :)

burrbull:
> Found an error after trying this out on hardware, fix pushed now
> 
> I responded to your review request @burrbull :)


As I don't have any H7 hardware, I can run it only in my mind.

PR #357, opened 2022-04-20
fix: flash bank impl visibility

ostenning:
Adjusts the visibility of various Flash `Bank` methods which will help users determine memory locations of sectors etc. 

For example if the user is writing a bootloader beginning at Bank 1, Sector 6, and they need to know the address to jump to, this will allow them easily.

### DIFF
--- a/src/flash.rs
+++ b/src/flash.rs
@@ -137,21 +137,21 @@ pub enum Bank {
 ))]
 impl Bank {
     /// The start address of the bank
-    fn start_address(&self) -> usize {
+    pub fn start_address(&self) -> usize {
         match *self {
             Bank::UserBank1 => BANK1_ADDR_START,
             Bank::UserBank2 => BANK2_ADDR_START,
         }
     }
     /// The end address of the bank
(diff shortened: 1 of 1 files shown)


src/flash.rs
```diff
@@ -137,21 +137,21 @@ pub enum Bank {
...
         match *self {
             Bank::UserBank1 => BANK1_ADDR_END,
             Bank::UserBank2 => BANK2_ADDR_END,
         }
     }
     /// The address of a given sector of a bank
-    fn sector_address(&self, sector: usize) -> usize {
+    pub fn sector_address(&self, sector: usize) -> usize {
         if sector > self.num_sectors() {
             panic!("Sector outside of bank range");
```
richardeoin, line 156: Could you add a Panics section to the docstring please? There's some [background/explaination in the Rust API guidelines](https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure)

richardeoin: Otherwise looks good 👍

PR #359, opened 2022-04-21
Convert ResetMode to own witness pin and track past pin mode

### LINKED ISSUE #358: Moder register not being reset after with-input method called
When using the `with-input` method on an output pin, the next outputs will be "eaten" and no longer physically happen. I tested with the stm32h723zg nucleo board.  It happened for every pin I tried as well.

I created a minimal-ish test case for it [here](https://github.com/jg2562/stm32hal-with-input) for anyone that wants to test it out.

jg2562:
Previously when resetting it would assume the pin was in the ORIG pin mode and try to convert it to the ORIG pin mode, which the if statements in mode blocked. Now by tracking the witness pin and the original MODE the reset can properly set the mode register bits.

By fixing ResetMode it allows for the `with_<mode>` methods to work and set the MODER bits. Fixes #358.

Since I've fairly new to heavy generics and microcontroller level optimization, it would be really great if someone could double-check that this doesn't break anything the previous implementation upheld! Thank you!

### DIFF
--- a/examples/gpio_with_input.rs
+++ b/examples/gpio_with_input.rs
@@ -1,0 +1,58 @@
+//! Demonstrates the use of the GPIO method `with_input` (and similar methods)
+//!
+//! https://docs.rs/stm32h7xx-hal/latest/stm32h7xx_hal/gpio/struct.Pin.html#method.with_input
+
+#![deny(warnings)]
+#![no_main]
+#![no_std]
+
+use cortex_m_rt::entry;
+
+use stm32h7xx_hal::{pac, prelude::*};
+
+use log::info;
+
+#[macro_use]
+mod utilities;
+
+#[entry]
+fn main() -> ! {
+    utilities::logger::init();
+    let cp = cortex_m::Peripherals::take().unwrap();
+    let dp = pac::Peripherals::take().unwrap();
+
+    // Constrain and Freeze power
+    info!("Setup PWR...                  ");
+    let pwr = dp.PWR.constrain();
+    let pwrcfg = example_power!(pwr).freeze();
+
+    // Constrain and Freeze clock
+    info!("Setup RCC...                  ");
+    let rcc = dp.RCC.constrain();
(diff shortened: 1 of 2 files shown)


richardeoin:
Thanks for tracking down this bug, and the fix. For the record, the underlying bug was incorrect use of the `mode()` method.

```rust
impl<..., MODE: PinMode> Pin<..., MODE> {
    /// Puts `self` into mode `M`.
    ///
    /// This violates the type state constraints from `MODE`, so callers must
    /// ensure they use this properly.
    pub(super) fn mode<M: PinMode>(&mut self) {
        if (MODE != M) { ... } // Edited for clarity
    }
}
```

but in the `Drop` implementation for `ResetMode`, it was called like this

```rust
impl <...> Drop for ResetMode<..., ORIG> {
    fn drop(&mut self) {
        self.pin.mode::<ORIG>();
    }
}
```

This is always a no-operation because `ORIG` == `ORIG`, hence the bug reported in #358.

I think this solution is a reasonable one, especially since `Pin` is a [ZST](https://doc.rust-lang.org/nomicon/exotic-sizes.html#zero-sized-types-zsts). I will tidy it up a little however, since it is correct but not as easy to read as it could be.

richardeoin:
This works on hardware for me, let's merge

bors r+

richardeoin:
@burrbull Does this bug need to be fixed in any other HALs? It seems to be present in at least [F4 as well](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/gpio/convert.rs#L350).

burrbull:
> @burrbull Does this bug need to be fixed in any other HALs? It seems to be present in at least [F4 as well](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/gpio/convert.rs#L350).

Yes, please.

jg2562:
Awesome! Thank you for reviewing and refactoring that! 

> For the record, the underlying bug was incorrect use of the mode() method.
I also want to add on top of that, ResetMode needed to keep track of both of the states hence adding the new pin to ResetMode along with the original state as part of the generics. This kept track of all the states allowing for proper use of mode()

PR #364, opened 2022-05-09
adc: simplify cr reads and sample time code

jordens:
* remove redundant CR register re-reads when checking multiple status bits
* rafactor common code in set_chan_smp()


richardeoin:
Thanks!

bors r+

PR #365, opened 2022-05-24
Fix intermittent flash programming error.

lachlansneff-parallel:
Writing flash sectors fails after a couple of writes without waiting for the EOP flag in the FLASH_SR# register to go low after each write. The reference doesn't mention that this is necessary, but it fails without this on my STM32H743IIT6 devboard.


richardeoin:
Thanks for the PR @lachlansneff-parallel ! This is a very useful improvement.

I modified the PR to remove the if statement below, since it was writing to the `SR` register (which is read-only) it's not needed.

PR #371, opened 2022-07-12
Write flash in 256-bit chunks correctly

lachlansneff-parallel:
Currently, the flash driver doesn't correctly split data into 256 bit chunks when writing. This PR fixes that as well as fixes error handing around flash operations and adds more expressive flash errors.

### DIFF
--- a/src/flash.rs
+++ b/src/flash.rs
@@ -447,9 +447,7 @@ impl Flash {
         self.unlock(bank)?;
 
         // 3. Enable double-word parallelism.
-        unsafe {
-            regs.cr.modify(|_, w| w.psize().bits(0b11))
-        }
+        unsafe { regs.cr.modify(|_, w| w.psize().bits(0b11)) }
 
         // 4. Enable write operations by setting the PG1/2 bit in the FLASH_CR1/2 register.
         regs.cr.modify(|_, w| w.pg().set_bit());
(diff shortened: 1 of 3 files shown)


src/flash.rs
```diff
@@ -416,26 +440,29 @@ impl Flash {
...
-        // 4. Write one Flash-word corresponding to 32-byte data starting at a 32-byte aligned address.
         let mut addr = bank.sector_address(sector) as *mut u32;
 
         // Offset it by the start position
         addr = unsafe { addr.add(start / 4) };
 
-        // Iterate on chunks of 32bit
-        for chunk in data.chunks_exact(4) {
+        // 6. Write a double-word corresponding to 32-byte data starting at a 32-byte aligned address.
+        // TODO: Use `.array_chunks` when it's stable.
```
ryan-summers, line 464: What does `array_chunks` give us over just `chunks_exact`? Seems like the existing impl is fine and we can remove the TODO?

ryan-summers (changes requested):
Can you please add a CHANGELOG entry as well?

src/flash.rs
lachlansneff-parallel: It would be a little cleaner (wouldn't need a `try_into` lower down), but I'm pretty sure it'll get optimized out, so doesn't really matter!~

lachlansneff-parallel: Removed!

richardeoin (approved):
Looks good, thanks for the PR @lachlansneff-parallel and @ryan-summers for reviewing

bors r+

PR #375, opened 2022-07-29
Correct flash word check

diondokter:
The check was accidentally done with the bit size instead of the byte size.
Later in the function the correct value of 32 is used in the chunks_exact function.


richardeoin (approved):
That is a small but painful bug. Thanks for the fix!

bors r+

PR #376, opened 2022-07-29
Added IWDT implementation

diondokter:
I had this still lying around.
Updated it a bit


diondokter:
Lol, so the CI decided to have an error...

I *think* everything is correct now

richardeoin:
I re-ran the CI :)

The IWDG peripheral is present on all the H7 sub-families. Can this HAL implementation be used with all the sub-families? In any case I would remove the [cfg guards from the example.](https://github.com/stm32-rs/stm32h7xx-hal/pull/376/files#diff-eab7cd5b9c35060b3e75d2d83ca618a9d3c58e30ecefdab1f35fdabae65f8024R29)

richardeoin:
Just tried building it, looks like the PAC is broken for the rm0468 sub-family but ok for the other three (rm0433, rm0399, rm0455)

richardeoin (approved):
I tested on a STM32H7A3 Nucleo board, and it works for me. Thanks @diondokter !

bors r+

diondokter:
Oh sweet! You finished it 😁
Cool, thank you for the work too! This is one of the most complete HALs out there which is really nice for such a complex chip.

PR #378, opened 2022-08-06
Update sdmmc support

richardeoin:
Update sdio-host version and improve eMMC support

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ cast = { version = "0.3.0", default-features = false }
 nb = "1.0.0"
 paste = "1.0.1"
 bare-metal = "1.0.0"
-sdio-host = { version = "0.7", optional = true }
+sdio-host = { version = "0.9", optional = true }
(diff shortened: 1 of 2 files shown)


richardeoin:
@ostenning There's a few changes to the sdmmc support here, it would be helpful if you could try this out!

ostenning:
Hi @richardeoin , I've given this branch a fairly quick test with the `embedded-sdmmc` crate, reading / writing data to the FAT32 formatted SDCard. I've had no problems as of yet.

richardeoin:
Thanks @ostenning 👍  I was looking for a sanity check really, so that test with `embedded-sdmmc-rs` is great

burrbull:
@richardeoin Do you plan to update f4xx-hal also?

richardeoin:
@burrbull No I don't plan to update f4xx-hal. The F4 reference manuals I found don't mention eMMC at all (although it might still be possible..) and I don't have any f4 + eMMC hardware to test on

Issue #374, opened 2022-07-28
Changing the ADC clock prescaler needs a workaround.


nkrackow:
Hi, I've been trying to set the clock prescaler for the internal ADCs and found it to be rather cumbersome.

The [adc initialization function](https://github.com/stm32-rs/stm32h7xx-hal/blob/71e2b75ceaa33e654271144af444c5a4c7dc85b7/src/adc.rs#L461) does not use or do anything with the ADC_COMMON registers in which the clock configuration takes place. Manually changing anything in these registers only works after the ADC is initialized (since it is reset in init) and before it is enabled ([condition form datasheet](https://www.st.com/resource/en/reference_manual/dm00314099-stm32h742-stm32h743-753-and-stm32h750-value-line-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf#page=1042)). 
But then the ADC has to be re-calibrated again since the clock impacts the offset values.
I've tested this [here](https://github.com/quartiq/stabilizer/blob/c040135dde7fcd7e3b6b108ed8a4193ea0016602/src/hardware/driver/adc_internal.rs#L57-L60).

All in all it might be nicer to incorporate the clock prescaler and clock mode from ADC_COMMON into the init function.
But I'm not sure how to do this in the best way. I think it would be a breaking change in any way, right?

nkrackow:
Additionally we now found that the 100 MHz maximum clock input that the [HAL checks for](https://github.com/stm32-rs/stm32h7xx-hal/blob/71e2b75ceaa33e654271144af444c5a4c7dc85b7/src/adc.rs#L36) is never a valid clock frequency.
https://github.com/quartiq/stabilizer/issues/571

jordens:
"never" as in "in the recent RMs the ADC max clock is 80 MHz".

richardeoin:
Thanks for reporting! Indeed the clock prescaler (and potentially also clock mode) should be integrated into the init function. It is a breaking change, but necessary so let's do it.

richardeoin:
Closed in #379

PR #379, opened 2022-08-14
Add initialization argument for ADC frequency

richardeoin:
Use the prescaler to set the fastest ADC clock below or equal to this limit

Addresses #374 

TODO:
- [x] ~Don't use Boost mode below 20MHz~ Set `BOOST` based on f_adc
- [x] Helpful panic message if calculated prescaler > 256
- [x] Check the ~f_ADC~ kernel clock frequency based on the current voltage scale, not just VOS0
- [x] ~Check minimum f_ADC~ There is no minimum with `BOOST = 0`
- ~Support CKMODE[1:0]~
- [x] Changelog
- [ ] ~Also check that the relevant pll output divider is even. Though that may require more fields and code in `ccdr.clocks`.~

### DIFF
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -580,14 +580,6 @@ macro_rules! adc_hal {
                     f_adc
                 }
 
-                /// The current ADC frequency. Defined as f_ADC in device datasheets
-                ///
-                /// The value returned by this method will always be equal or
-                /// lower than the `f_adc` passed to [`init`](#method.init)
-                pub fn frequency(&self) -> Hertz {
-                    self.frequency
-                }
-
                 /// Disables Deeppowerdown-mode and enables voltage regulator
                 ///
                 /// Note: After power-up, a [`calibration`](#method.calibrate) shall be run
@@ -866,6 +858,14 @@ macro_rules! adc_hal {
                     self.set_resolution(Resolution::SixteenBit);
                     self.set_lshift(AdcLshift::default());
                     cfg
+                }
+
+                /// The current ADC frequency. Defined as f_ADC in device datasheets
+                ///
+                /// The value returned by this method will always be equal or
+                /// lower than the `f_adc` passed to [`init`](#method.init)
+                pub fn frequency(&self) -> Hertz {
+                    self.frequency
                 }
 
                 /// Get ADC samping time
--- a/src/adc.rs
+++ b/src/adc.rs
@@ -673,7 +673,17 @@ macro_rules! adc_hal {
                     #[cfg(not(feature = "revision_v"))]
                     self.rb.cr.modify(|_, w| w.boost().set_bit());
                     #[cfg(feature = "revision_v")]
-                    self.rb.cr.modify(|_, w| w.boost().lt50());
+                    self.rb.cr.modify(|_, w| {
+                        if self.frequency.raw() <= 6_250_000 {
+                            w.boost().lt6_25()
+                        } else if self.frequency.raw() <= 12_500_000 {
+                            w.boost().lt12_5()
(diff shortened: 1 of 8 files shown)


src/adc.rs
```diff
@@ -495,11 +493,57 @@ macro_rules! adc_hal {
...
+                    #[cfg(feature = "revision_v")]
+                    let f_adc = Hertz::from_raw(ker_ck.raw() / (divider * 2));
+
+                    // Calculate actual value Revison Y. See RM0433 Rev 7 - Figure 137.
+                    #[cfg(not(feature = "revision_v"))]
+                    let f_adc = Hertz::from_raw(ker_ck.raw() / divider);
+
+                    // Maximum value, for VOS0 only. Other voltage scale values
+                    // result in a lower maximum f_adc
+                    assert!(f_adc.raw() <= 50_000_000);
```
jordens, line 539: I think we want 40 MHz here.

```suggestion
                    assert!(f_adc.raw() <= 40_000_000);
```

c.f. https://github.com/quartiq/stabilizer/pull/584#discussion_r945547973

jordens (commented):
* [ ] Also check that the relevant pll output divider is even. Though that may require more fields and code in `ccdr.clocks`.

richardeoin:
This PR now [checks the kernel clock](https://github.com/stm32-rs/stm32h7xx-hal/pull/379/files#diff-09d1f58a7f7c271d63f1a9a00c0c48c886e094bc98c207a8d7688c9f874259bbR557) is within the limits set in the Reference Manual.

The duty cycle check still needs to be implemented. @jordens do you have any thoughts about what the HAL should do if the duty cycle is not 50%? Clearly some deviation from 50% is acceptable at frequencies well below the maximum given in the RM.

src/pwr.rs
```diff
@@ -220,6 +220,48 @@ macro_rules! d3cr {
...
+    #[cfg(all(
+        feature = "revision_v",
+        any(feature = "rm0433", feature = "rm0399")
+    ))]
+    if unsafe { (*SYSCFG::ptr()).pwrcr.read().oden().bit_is_set() } {
+        return VoltageScale::Scale0;
+    }
+
+    // NOTE(unsafe): Read-only access
+    let d3cr = unsafe { d3cr!(*PWR::ptr()).read() };
```
jordens, line 235: Doesn't clippy complain about dead code if the cfg above is active?

jordens (approved):
> The duty cycle check still needs to be implemented. @jordens do you have any thoughts about what the HAL should do if the duty cycle is not 50%? Clearly some deviation from 50% is acceptable at frequencies well below the maximum given in the RM.

I am not sure that the duty cycle constraint is there to ensure a minimum low/high half clock cycle duration for timing closure reasons. I could imagine some ADC architectures that also depend on very balanced clock cycles for their accuracy. But yes, in any case a large odd divider (correlating with a low output frequency) should be less bad than a small odd divider.

I think the correct behavior would be to panic on odd divider feeding the ADC. I don't know in which cases it is fine in practice.

That said, I don't think this constraint is so critical. And it would require quite some code and data to CoreClocks to track duty cycle as well. Maybe leave this for later.

jordens:
Great work. Thanks a lot.
I checked the RM0433 aspects of this on paper. Might try running it on hardware later.

src/pwr.rs
richardeoin: This code is still reachable at runtime even in the cfg is active, since the return is conditional

richardeoin:
Ok, thanks for the review. Agreed, let's merge this as it is now!

PR #381, opened 2022-08-17
Pwm channels cleanups

burrbull:
I can't find any sense `Polarity` to be part of channel type because it doesn't affect on what traits and methods can be implemented on type. So I've removed `ActionLow` and `ActionHigh` from type and added `set_polarity` and `set_comp_polarity` runtime methods.

### DIFF
--- a/src/pwm.rs
+++ b/src/pwm.rs
@@ -1019,8 +1019,12 @@ macro_rules! pwm_ext_hal {
 
 // Implement PWM configuration for timer
 macro_rules! tim_hal {
-    ($($TIMX:ident: ($timX:ident, $Rec:ident,
-                     $typ:ty, $bits:expr $(, DIR: $cms:ident)* $(, BDTR: $bdtr:ident, $moe_set:ident, $af1:ident, $bkinp_setting:ident $(, $bk2inp_setting:ident)*)*),)+) => {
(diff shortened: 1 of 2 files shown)


richardeoin:
Thanks for the PR! I can't immediately think of a good reason to have Polarity as part of channel type, and agreed it's not needed for the implementation

The examples are still building fine, which is great. However this PR does changes the crate's API, so I will review it a bit more.

src/pwm.rs
```diff
@@ -1233,16 +1276,16 @@ macro_rules! tim_hal {
                 $(
                     /// Set the deadtime for complementary PWM channels of this timer
                     #[must_use]
-                    pub fn with_deadtime<T: Into<NanoSeconds>>(mut self, deadtime: T) -> Self {
+                    pub fn with_deadtime(mut self, deadtime: NanoSeconds) -> Self {
```
richardeoin, line 1279: Accepting `NanoSeconds` instead of `Into<Nanoseconds>` is less flexible and a breaking change. Could it be left as before?

burrbull: reverted

richardeoin (approved):
I tested most of the [pwm_advanced](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/examples/pwm_advanced.rs) example again using this PR. Looks good to me! Thank you very much for this refactoring.

bors r+

PR #386, opened 2022-09-01
Poll LDORDY bit before continuing with ADC calibration

richardeoin:


richardeoin:
This bit does not exist on RM0468 ADC3

PR #390, opened 2022-09-19
SPI : fix UB by avoiding the need of two aliasing (mutable) references

pwnorbitals:


src/spi.rs
```diff
@@ -1168,16 +1168,12 @@ macro_rules! spi {
 
                     /// Internal implementation for blocking::spi::Transfer and
                     /// blocking::spi::Write
```
jannic, line 1170: This comment is no longer true: Transfer and Write now have separate internal functions.
Which makes me wonder: Is it still useful to have those internal functions, or could they be inlined into Transfer::transfer and Write::write?
(Keeping them separate could be useful if they can be reused to implement embedded-hal 1.0.0 traits.)

pwnorbitals: It looks like they could be useful to implement `embedded-hal 1.0.0` traits `SpiBus`, `SpiBusRead` and `SpiBusWrite` (https://docs.rs/embedded-hal/1.0.0-alpha.8/embedded_hal/spi/blocking/index.html). In the meantime, I agree inlining into the traits would be cleaner

richardeoin:
Thanks for the PR! Agreed on the principle here, this change makes the code more verbose but avoids UB. Out of interest, did you find this from reviewing the code or from debugging a behavioural issue?

For anyone interested here's the [link to the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2923dc234e972299fa59b20485c807b3) where MIRI flags an issue (originally posted by @jannic I think).

pwnorbitals:
I found it while actively looking for UB in my dependencies around the SPI code. I'm debugging a behavioral issue where the re-initilaization of SPI in the (non-inlined) panic handler would always make all other (non-panic-related) SPI accesses fail. It's most probably a miscompilation caused by UB somewhere but fixing this one didn't help, unfortunately

src/spi.rs
```diff
@@ -1273,7 +1273,7 @@ macro_rules! spi {
                                 words[i - FIFO_WORDS] = read_value;
 
                             // Finish emptying the read FIFO
-                            } else if i < len + FIFO_WORDS {
+                            } else {
```
pwnorbitals, line 1274: Ah yes, it's redundant but was done on purpose to help the reader understand the code. Fair to remove it though

richardeoin:
Interesting, please do make a new issue if it is eventually something connected to the HAL. Thanks for this PR in any case!

bors r+

PR #397, opened 2022-10-25
Documentation improvements

richardeoin:
* Add missing links to examples from docs
* Add a description to more examples
* Update some comments


jordens (changes requested):
Otherwise LGTM.

examples/adc.rs
```diff
@@ -1,4 +1,4 @@
-//! Example of reading a voltage with ADC1
+//! Example of reading a voltage with ADC1 through DMA
```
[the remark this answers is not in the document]
richardeoin: Thanks for the review, indeed a typo

examples/adc_dma.rs
```diff
@@ -1,4 +1,4 @@
-//! Example of reading ADC1 data using DMA1/2
+//! Example of reading ADC1 data using DMA1/2 through DMA
```
jordens, line 1: I think this was fine before

```suggestion
//! Example of reading ADC1 data using DMA1
```

jordens (changes requested):

examples/adc_dma.rs
richardeoin: Thanks, corrected now

PR #408, opened 2023-01-01
Add halfduplex to serial::config::Config and example

dsmcfarl:
This allows the USART's to be used with serial devices that operate in a half-duplex mode (transmit and receive on one wire). An example was also added (serial-halfduplex) and has been tested with a NUCLEO-H723ZG board. It has also been tested with a Frysky RX4R RC receiver Smart Port interface.

### DIFF
--- a/examples/serial-halfduplex.rs
+++ b/examples/serial-halfduplex.rs
@@ -71,14 +71,14 @@ fn main() -> ! {
     loop {
         // Tx from usart3
         for c in "Hello, world!".chars() {
-            let _res = block!(usart3.write(c as u8)).unwrap();
+            block!(usart3.write(c as u8)).unwrap();
             let received = block!(usart6.read()).unwrap();
             info!("usart6 rx {}", received as char);
         }
 
         // Tx from usart6
         for c in "Hello, world!".chars() {
-            let _res = block!(usart6.write(c as u8)).unwrap();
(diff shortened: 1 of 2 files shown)


richardeoin:
Thanks for the PR! This looks really nice and I appreciate the explaination of the setup in the example.

In the existing HAL it was already possible to configure synchronous mode (3 wires!) by passing 3 pins during the initialisation. That then sets the `CLKEN` bit high [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/serial.rs#L691).

```rust
let mut usart6 = dp
        .USART6
        .serial((usart6txpin, usart6rxpin, usart6clk), ...);
```

According to the reference manual the `CLKEN` and `HDSEL` bits are mutually exclusive - so you can't set both at the same time. I think the HAL should return an error or panic in that configuration rather than just going ahead. Could you add a check and return `Err()` for this?

richardeoin:
The CI should now pass

bors try

richardeoin:
I would like to have this PR merged for the next release, so at some point I will make the change I described in my earlier comment. @dsmcfarl please let me know if that's a problem or you would prefer something else!

richardeoin:
Updated and rebased onto master

dsmcfarl:
Just saw this and see you merged it. Looks great thanks!

PR #409, opened 2023-01-04
make `Alternate` variants readable

andrewgazelka:
- allows us to read any alternate pin
- no longer have to read `Alternate` registers manually
- I am not super familiar with the ref manual, so if anyone has any other thoughts/criticism I would love to hear.

### DIFF
--- a/src/gpio.rs
+++ b/src/gpio.rs
@@ -190,6 +190,7 @@ mod marker {
 impl<MODE> marker::Interruptable for Output<MODE> {}
 impl marker::Interruptable for Input {}
 impl marker::Readable for Input {}
+impl<const A: u8, MODE> marker::Readable for Alternate<A, MODE> {}
 impl marker::Readable for Output<OpenDrain> {}
 impl marker::Active for Input {}
 impl<Otype> marker::OutputSpeed for Output<Otype> {}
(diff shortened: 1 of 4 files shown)


src/gpio.rs
```diff
@@ -190,6 +190,7 @@ mod marker {
 impl<MODE> marker::Interruptable for Output<MODE> {}
 impl marker::Interruptable for Input {}
 impl marker::Readable for Input {}
+impl<const A: u8, MODE> marker::Readable for Alternate<A, MODE> {}
```
ryan-summers, line 193: It's a bit... Odd to read an AF pin IMO, but I referenced the datasheet and it's certainly allowed on this architecture. Ref section 11.3.12 of the RM

ryan-summers (approved):
Mind also hitting the Github Actions warning lints as well?

andrewgazelka:
> Mind also hitting the Github Actions warning lints as well?
> 
> Edit:
> 
> * [ ]  Please also add a CHANGELOG entry for the update to make Alternate pins readable

Looks like the warnings are an open issue

- https://github.com/actions-rs/clippy-check/issues/2

burrbull:
What about `Interruptable` marker. Should it also be implemented for `Alternate` pins?

PR #410, opened 2023-01-04
pin clippy

andrewgazelka:
- avoids [errors popping up in PRs all of a sudden](https://github.com/stm32-rs/stm32h7xx-hal/pull/409)


richardeoin:
Agreed, let's do this to avoid unrelated errors appearing on new PRs

bors r+

PR #411, opened 2023-01-04
add VSC settings

andrewgazelka:
- add VSC settings
- helps contributers be able to work out of the box
- currently `clippy` and `rust-analyzer` do not work out of the box


richardeoin:
Thanks, this is useful for new contributors

bors r+

PR #418, opened 2023-02-06
Add some missing Defmt impls

mattico:
Plus a drive-by deprecation warning fix.


richardeoin (approved):
LGTM, thanks!

bors r+

PR #419, opened 2023-02-17
Add LPTIM traits

andrewgazelka:


src/prelude.rs
```diff
@@ -35,7 +35,9 @@ pub use crate::spi::HalEnabledSpi as _stm32h7xx_hal_spi_HalEnabledSpi;
 pub use crate::spi::HalSpi as _stm32h7xx_hal_spi_HalSpi;
 pub use crate::spi::SpiExt as _stm32h7xx_hal_spi_SpiExt;
 pub use crate::time::U32Ext as _stm32h7xx_hal_time_U32Ext;
-pub use crate::timer::TimerExt as _stm32h7xx_hal_timer_TimerExt;
+pub use crate::timer::HalDisabledLpTimer as _stm32h7xx_hal_timer_HalDisabledLpTimer;
```
richardeoin, line 38: The `TimerExt` for ordinary timers needs to remain in the prelude

richardeoin (approved):
Seems reasonable. Thanks for the PR!

src/prelude.rs
andrewgazelka: oops did not mean to delete will fix

andrewgazelka: fixed :)

https://github.com/stm32-rs/stm32h7xx-hal/blob/31d696732992cbdec28100cbe3b12d525aa351ed/src/prelude.rs#L38-L41

richardeoin (approved):
bors r+

PR #423, opened 2023-02-24
pwr: remove unused parameter

JoseGuilhermeCR:
`Pwr::vos0` is the only function in Pwr that takes an unused reference to `SYSCFG`. I suppose it remained undetected because of the `_` as it's name.

This removes the parameter and makes it the same as `Pwr::vos{1,2,3}`.

### DIFF
--- a/src/pwr.rs
+++ b/src/pwr.rs
@@ -396,7 +396,7 @@ impl Pwr {
         any(feature = "rm0433", feature = "rm0399", feature = "rm0468")
     ))]
     #[must_use]
-    pub fn vos0(mut self, _: &SYSCFG) -> Self {
+    pub fn vos0(mut self) -> Self {
         self.target_vos = VoltageScale::Scale0;
         self
     }


richardeoin:
Thank you for taking the time to make this PR! Unfortunately the unused reference to `SYSCFG` is deliberate in this case and shouldn't be removed.

The unused reference to `SYSCFG` is part of the external API as it forces the user to ensure that there is no other *mutable* reference to `SYSCFG` in existence at the point that the builder method is called. This is done so that there is no concurrent access to SYSCFG (for example in an interrupt context) when [SYSCFG is modified in the freeze method](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/pwr.rs#L507). This API pattern isn't watertight - and perhaps should be better documented - but was intentional.

JoseGuilhermeCR:
Oh I had a bad feeling about this, tried to check if it was being used anywhere else but supposed it wasn't. Either way, it's a nice thing that the language allows us to do so in order to protect usage in other places, but I'd definitely leave it documented so something like this doesn't happen anymore :sweat_smile:.

PR #428, opened 2023-03-31
Make support for all the H7(2,3)(3,5) variants explicit

jlogan03:
For now, the settings are all the same as the H735 because the security features that exist in some variants and not others are not supported for any variant, and the only other differences are in memory layout.

I have not exhaustively tested support for the added chips, but have been able to run blink and UART serial programs on the NUCLEO-H723ZG board: https://github.com/jlogan03/nucleo-h723zg .


richardeoin:
Thanks for the PR! The change to the README looks good! As you say these parts can be supported with the same configuration as the stm32h735.

Although we didn't always do this in the past, I would prefer to only add new feature flags if they can be tested in Continuous Integration [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/.github/workflows/ci.yml#L24) and [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/.github/bors.toml#L12), and if these test a meaningfully different configuration. The idea would be to keep the delta between the list of possible features and those tested in CI as small as possible. Because of that I don't see a big benefit in adding the new flags to Cargo.toml.

There are some tables in the documentation that currently refer to stm32h725/735 [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/pwr.rs#L65) [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/system_watchdog.rs#L15) [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/independent_watchdog.rs#L17) and [here](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/src/xspi/mod.rs#L8). I guess that all of 723,725,733,735 (and 730) apply in these cases?

jlogan03:
Makes sense - I removed the new feature flags and just updated the documentation instead. How's that look?

richardeoin:
Looks good, thanks! I think you can remote the patch version bump now?

jlogan03:
Done. I had added the patch version bump because there is a change to one of the feature flag gates to remove the stm32h725 flag, but since that flag did not exist to begin with, I don't think there are any cases where it could have a real effect.

richardeoin:
Got it. I'll merge this now

bors r+

PR #431, opened 2023-04-16
alternate pin function enums

burrbull:


burrbull:
@richardeoin Is there other reason for `new_unchecked` except of possibility to `release`?

New API allows you to keep pins inside structure without advanced `<PINS>` generic (at least for simple cases). See:
https://github.com/stm32-rs/stm32f4xx-hal/blob/557b7f2de9963ef2cbe11b554d1eab7894b9a007/src/i2c.rs#L71-L74

richardeoin:
The `new_unchecked` methods are also useful for using a peripheral without encoding its GPIOs in the type system. In some situations the code and effort required to state the type of all GPIOs is excessive, so the `new_unchecked` methods are a useful escape hatch. For example to setup the GPIOs in `main()`, and setup a peripheral in `my_func()` without passing GPIOs as arguments to `my_func()`. With this PR there is less need for the `new_unchecked` methods, but I would rather keep them available for now.

Overall this PR is quite a nice usability improvement. I would like to play with it some more, but in general looks good! :)

burrbull:
This PR requires #346 to be merged first especially for h735.

burrbull:
rebased

PR #440, opened 2023-07-22
Bump dependencies

richardeoin:
* Bump embedded-sdmmc from 0.4 to 0.5
* Bump fdcan from 0.1 to 0.2
* Bump embedded-graphics from 0.7 to 0.8
* Update [[example]] targets

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ stm32-fmc = { version = "0.3", optional = true }
 synopsys-usb-otg = { version = "^0.3.0", features = ["cortex-m"], optional = true }
 embedded-display-controller = { version = "^0.1.0", optional = true }
 log = { version = "0.4.14", optional = true} # see also the dev-dependencies section
-fdcan = { version = "0.1", optional = true }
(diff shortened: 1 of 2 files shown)


phansel:
For some reason the crates.io index has not been updated: https://github.com/rust-lang/crates.io-index/blob/master/st/m3/stm32h7xx-hal#L19

None of the changes made here are actually reflected at the crate registry, i.e. crates.io thinks stm32h7xx_hal still wants embedded-sdmmc=0.4, not 0.5, and so on.

richardeoin:
Hey @phansel. This is expected, because stm32h7xx_hal hasn't been released since March. Releasing the crate to crates.io is something that I (or Andrew) have to do manually, and currently I try to do it every 6 months or so! :)

In the meantime you have to either use the crate as a [git dependency](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories), or yes just be patient. When I release the crate I will also increase the version number.

phansel:
Thanks for the pointer!

PR #442, opened 2023-07-28
feat(adc): spike for ADC reads using DMA w/ circular buffer

dobrite:
This implements continuous ADC reads of multiple channels using DMA and a circular buffer. Values should be able to be read directly from the buffer from any context, not just interrupts.

This is unsafe, but it's working and so gives a jumping off point for improvements. I'm not actually sure if the use-case is even possible in safe Rust.

Seeking ideas to improve:

* Needing to [pass in an array](https://github.com/dobrite/stm32h7xx-hal/pull/1/files#diff-43871c78ae970b705edf0c2c78115ac3ce5c692832a4e021a55095474622ec64R99-R100) of `u8`s instead of an array of pins.
* [Having to `read_volatile`](https://github.com/dobrite/stm32h7xx-hal/pull/1/files#diff-43871c78ae970b705edf0c2c78115ac3ce5c692832a4e021a55095474622ec64R105) because the transfer has `&mut` ownership over the buffer.  There would also need to be an API to access this buffer from outside an interrupt context, where we might not have access to the transfer.
* `DmaConfig` needs to be setup correctly (e.g. `circular_buffer(true)`), otherwise it won't work.
* Anything else! I'm relatively new to embedded, and I'm sure there are _tons_ of spots where this approach and code can be improved.

### DIFF
--- a/examples/adc_dma_circ.rs
+++ b/examples/adc_dma_circ.rs
@@ -1,0 +1,111 @@
+#![no_main]
+#![no_std]
+
+use core::{mem, mem::MaybeUninit};
+use log::info;
+
+use cortex_m_rt::entry;
+
+use embedded_hal::adc::Channel;
+use stm32h7xx_hal::{
+    adc,
+    delay::Delay,
+    dma::{
+        dma::{DmaConfig, StreamsTuple},
+        Transfer,
+    },
+    gpio, pac,
+    prelude::*,
+    stm32::ADC1,
+};
+
+#[macro_use]
+mod utilities;
+
+#[entry]
+fn main() -> ! {
+    utilities::logger::init();
+    let cp = cortex_m::Peripherals::take().unwrap();
+    let dp = pac::Peripherals::take().unwrap();
+
+    // 1 slot because we are only using 1 channel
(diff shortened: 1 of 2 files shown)


examples/adc_dma_circ.rs
```diff
@@ -0,0 +1,111 @@
...
+        .memory_increment(true);
+
+    // Setup the DMA transfer on stream 0
+    let streams = StreamsTuple::new(dp.DMA1, ccdr.peripheral.DMA1);
+    let mut transfer: Transfer<_, _, _, _, _> =
+        Transfer::init(streams.0, adc1, &mut adc_buffer[..], None, config);
+
+    // This closure runs right after enabling the stream
+    transfer.start(|adc| {
+        let channel = <gpio::Pin<'C', 4> as Channel<ADC1>>::channel();
```
burrbull, line 99:
```suggestion
        let channel = <gpio::PC4 as Channel<ADC1>>::channel();
```

dobrite:
Closing as no interest.

everdrone:
This would have been extremely useful to me.
I see that it's currently supported in the stm32f4xx-hal crate but not here

PR #444, opened 2023-07-31
Ptp

ZachParallel:


adamgreig:
It looks like this PR incorporates a lot of code from https://github.com/stm32-rs/stm32-eth/pull/77 which is under a different licence (Apache 2 vs 0BSD here), so probably shouldn't be merged as-is.

I think either have stm32h7xx-hal depend on stm32-eth (if/once that PR is merged), which I think is the best choice, or acknowledge the source of the code with a copyright header in those files and an update to the main stm32h7xx-hal README and Cargo.toml about the multiple licences and other authors.

If it's possible for @datdenkikniet to agree to relicense the code 0BSD that could also work, but they should have been asked first, and they would still need acknowledging since they're not in the git commit history this way.

richardeoin:
In the medium-term it would be great for stm32h7xx-hal to depend on stm32-eth, although that requires the resources for https://github.com/stm32-rs/stm32-eth/pull/77 to be finished, tested and merged.

Incorporating parts of that code without acknowledging the license is of course not a solution. I would agree that this PR can't be merged as-is and ideally there would be a reference to the original code in a header or otherwise asap.

HarishgunaS:
I've asked permission from @datdenkikniet and they have agreed to relicense the code to stm32h7xx-hal under 0BSD. I have added the relevant acknowledgement headers in the appropriate files as requested by them as well. i believe there are more changes that need to be made before merging, but hopefully this means that this PR can get merged after that @adamgreig @richardeoin

datdenkikniet:
Confirmed! The bits of code reused in this PR are relicensed under 0BSD (new bits, too, so long as I've authored them. I can't speak for authors other than myself, but AFAIK we're not touching any of that here). We discussed and this seemed like a reasonable approach :)

HarishgunaS:
PR is ready for review

PR #451, opened 2023-09-11
Replace casts of &T to &mut T with UnsafeCell

richardeoin:
Specifically we cast `*const T` to `*const UnsafeCell<T>` then use the [`UnsafeCell::raw_get`](https://doc.rust-lang.org/stable/std/cell/struct.UnsafeCell.html#method.raw_get) method.


richardeoin:
Casts of `&T` to `&mut T` are UB. Caught by [this lint](https://doc.rust-lang.org/nightly/rustc/lints/listing/deny-by-default.html#invalid-reference-casting).

richardeoin:
Changes to the SPI implementation validated on hardware

PR #452, opened 2023-09-18
Fix invalid read test logic in sdmmc example

phoracek:
The read check is processing a buffer 10 blocks long. The check then reads data from the SD card 10 times to then report the speed of "Read 10 blocks at X bytes/second". The problem is that it actually reads 10*10 blocks.

With this patch, `read_blocks` is used correctly to populate a continuous buffer.


phoracek:
Better to tackle this as a part of https://github.com/stm32-rs/stm32h7xx-hal/pull/453 which reworks the example anyway.

PR #453, opened 2023-09-18
Introduce SDMMC `write_blocks`

phoracek:
`write_block` needs to be called for every 512 bytes long block that is to be written to the SD card. Each of these calls requires the overhead of negotiating the write command and if I understand it correctly it also requires to erase the whole SD segment and copying it back. 

### Speed

With this patch, `write_blocks` is introduced, mirroring the behavior of `read_block` and `read_blocks`. This new method by itself helped me to improve write speed from the original example of 9.2 kB/s to 65 kB/s.

Write performance with 480 MHz system clock, 50 MHz bus speed, and enabled caches:

| Buffer size | Total size written | Speed |
| -- | -- | -- |
| 64 kB | 64 kB | 513 kB/s |
| 64 kB | 32 MB | 4.9 MB/s |
|128 kB | 32 MB | 5.2 MB/s |

### Erase before writing

Using ACMD23 to erase block data before writing in my case did not improve the result. Either my card does not benefit from it or I used it wrong.

### Possible improvements

Although the results look good Class 10 SD cards should be much faster. Even with half the bus speed, it should achieve at least 10 MB/s. If anybody has an idea what may be the bottleneck, please share.

### Follow ups

- Utilize block write/read in FAT bindings: #456

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Update `smoltcp` dependency to `0.9.0`
 * MSRV increased to 1.65.0
 * add `IntoAf` trait to restrict `into_alternate` [#346]
+* sdmmc: Fix read speed test.
 
 ## [v0.14.0] 2023-03-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * Update `smoltcp` dependency to `0.9.0`
 * MSRV increased to 1.65.0
 * add `IntoAf` trait to restrict `into_alternate` [#346]
-* sdmmc: Fix read speed test.
+* sdmmc: Introduce `write_blocks` [#453]
 
 ## [v0.14.0] 2023-03-22
 
--- a/examples/sdmmc.rs
+++ b/examples/sdmmc.rs
@@ -145,15 +145,13 @@ fn main() -> ! {
     info!("");
 
     // Read test
-    let mut buffer = [0u8; 5120];
+    let mut buffer = [0u8; 512 * 10];
 
     cp.DWT.enable_cycle_counter();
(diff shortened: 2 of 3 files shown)


ostenning:
Strange, we seem to have come up with a PR to solve this at the same time. 

I use CMD23 to set the SD block length for the transfer, which may help, see https://github.com/stm32-rs/stm32h7xx-hal/pull/455

phoracek:
@ostenning ha! :grin:

Interesting, I did not know CMD23 can be used instead of CMD12.

I tried using CMD23 from your patch in my benchmark instead of `stop_transmission`. CMD23 was a little bit slower (4.9 MB/s instead of 5.3 MB/s), but that may have been just my specific card or the benchmark config.

ostenning:
> @ostenning ha! 😁
> 
> Interesting, I did not know CMD23 can be used instead of CMD12 for SD.
> 
> I tried using CMD23 from your patch in my benchmark instead of `stop_transmission`. CMD23 was a little bit slower (4.9 MB/s instead of 5.3 MB/s), but that may have been just my specific card or the benchmark config.

I'm sure @richardeoin will have greater insights

phoracek:
Thanks @ostenning.

I have posted a draft of this new method being used in FAT writes. It still needs some care, so I keep it in a separate PR: https://github.com/stm32-rs/stm32h7xx-hal/pull/456

richardeoin:
Great! I don't have any specific insights, other than that this looks like a nice PR.

It's possible that some cards use CMD23 to optimise writes, but here I like the simplicity of having `write_blocks` as similar to `write_block` as possible.

It looks like this PR is ready, so I'll go ahead and merge it

richardeoin (approved):

PR #456, opened 2023-09-19
Utilize write_blocks in FAT binding 

phoracek:
With this patch, the FAT write binding would use sdmmc CMD25 to write blocks of data.

It also extends the FAT example so it exercises both `write_block` and `write_blocks`.


phoracek:
Hey @richardeoin, once you'll have some time, this one is ready for a review. Thanks o/

richardeoin:
Great. I'll take a look in the next week or so

richardeoin (approved):
Looks good to me!

PR #462, opened 2023-10-10
Second try at adding PTP support

HarishgunaS:
Adding PTP support for H7 that is completely feature gated. When the PTP gate is disabled, there should be no difference between the current EthernetDMA functionality and interface


HarishgunaS:
Hey @richardeoin, this PR is ready. Please review if and when you have time. Thanks!

jmgrosen:
@richardeoin I'd love if you could review and merge this as well!

PR #469, opened 2023-11-20
Updating usb crates

ryan-summers:
This PR updates the various usb-* crates, which just got updated releases last week.

### DIFF
--- a/src/usb_hs.rs
+++ b/src/usb_hs.rs
@@ -18,7 +18,7 @@
 use crate::rcc;
 use crate::stm32;
 
-use crate::gpio::{self, Alternate, Analog, Speed};
+use crate::gpio::{self, Alternate, Speed};
 
 use crate::time::Hertz;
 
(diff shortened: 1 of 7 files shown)


src/usb_hs.rs
```diff
@@ -50,8 +50,8 @@ impl USB1 {
         usb_global: stm32::OTG1_HS_GLOBAL,
         usb_device: stm32::OTG1_HS_DEVICE,
         usb_pwrclk: stm32::OTG1_HS_PWRCLK,
-        _pin_dm: gpio::PA11<Analog>,
-        _pin_dp: gpio::PA12<Analog>,
+        _pin_dm: gpio::PA11<crate::gpio::Analog>,
```
ryan-summers, line 53: This change is because clippy was throwing annoying warnings when the features weren't enabled about an unused `Analog` import

richardeoin (approved):
Really awesome work on `usb-device` upstream, thanks @ryan-summers !

I re-tested the `usb_serial` and `usb_rtic` examples on a NUCLEO-H7A3ZI-Q and they worked correctly.

PR #479, opened 2024-01-12
add new example to control LED1, 2, 3 and 4 on stm32h747i-disco

youndong:
Example : Control LED1, 2, 3 and 4 on stm32h747i-disco
- GPIO port : PI12 for LED1 
- GPIO port : PI13 for LED2 
- GPIO port : PI14 for LED3 
- GPIO port : PI15 for LED4

### DIFF
--- a/examples/blinky-stm32h747-disco.rs
+++ b/examples/blinky-stm32h747-disco.rs
@@ -1,0 +1,67 @@
+//! Basic example that produces a 1Hz square-wave on Pin PE1
+ 
+//#![deny(warnings)] // This code runs on stm32h747i-disco and does not use example! macro.
+#![no_main]
+#![no_std]
+ 
+use cortex_m_rt::entry;
+use stm32h7xx_hal::{pac, prelude::*};
+ 
+use log::info;
+ 
+#[macro_use]
+mod utilities;
+ 
+#[entry]
+fn main() -> ! {
+    utilities::logger::init();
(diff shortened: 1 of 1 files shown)


examples/blinky-stm32h747-disco.rs
```diff
@@ -0,0 +1,67 @@
+//! Basic example that produces a 1Hz square-wave on Pin PE1
+ 
+//#![deny(warnings)] // This code runs on stm32h747i-disco and does not use example! macro.
```
richardeoin, line 3: Please keep the deny(warnings), but you can use an allow(unused_macros)

```suggestion
#![deny(warnings)] 
#![allow(unused_macros)] // This code runs on stm32h747i-disco and does not use example! macro.
```

richardeoin:
Thanks for the PR! This is a nice board-specific example.

Because this example will not compile for some STM32H7 parts (for example those without a SMPS), you need to add an [entry for it in Cargo.toml](https://github.com/stm32-rs/stm32h7xx-hal/blob/master/Cargo.toml#L154) . Right now they are in alphabetical order, so this one would go at the top

```
[[example]]
name = "blinky-stm32h747-disco"
required-features = ["rm0433"]
```

Also you will need to run `rustfmt` on your rust file, which we use for consistent formatting

youndong:
Thank you for your very helpful comment. I am going to request a new PR soon.

PR #483, opened 2024-01-20
Remove references to `static mut`

richardeoin:
[rust/#114447](https://github.com/rust-lang/rust/issues/114447) made `&mut MY_STATIC_MUT` a warning, with the intention of disallowing it entirely in the 2024 edition. We never used this in the crate itself, but it is used in many examples.

This commit removes these references from the examples. In general the new approach is to change `static mut` of type `X` to type `MaybeUninit<X>`. These can then cleanly be accessed with [`MY_STATIC_MUT.write()`](https://doc.rust-lang.org/core/mem/union.MaybeUninit.html#method.write) followed by [`MY_STATIC_MUT.assume_init_mut()`](https://doc.rust-lang.org/core/mem/union.MaybeUninit.html#method.assume_init_mut).

Where we need to cast a (reference to a)`MaybeUninit<[X; N]>` to a `[MaybeUninit<X>; N]` in order to initialise it element-by-element, the `&mut *ptr::addr_of_mut!(MY_STATIC_MUT)` construction is still used. After initialisation, we can then access it with `MY_STATIC_MUT.assume_init_mut()`.

In a few DMA Examples we were taking a reference `&mut TARGET_BUFFER` without initialising TARGET_BUFFER. This commit changes this to `TARGET_BUFFER.assume_init_mut()` *without* initialising `TARGET_BUFFER`. This was and remains UB.

* USB Examples: `static mut EP_MEMORY` now has type `MaybeUninit<[u32; 1024]>`
* Ethernet Examples: `static mut DES_RING` now has type `MaybeUninit<ethernet::DesRing<4, 4>>`

### DIFF
--- a/examples/adc_dma.rs
+++ b/examples/adc_dma.rs
@@ -4,6 +4,7 @@
 //! For an example of using ADC1 and ADC2 together, see examples/adc12.rs
 //! For an example of using ADC1 and ADC2 in parallel, see examples/adc12_parallel.rs
 
+#![deny(warnings)]
 #![no_main]
 #![no_std]
 
@@ -33,8 +34,11 @@ fn main() -> ! {
 
     let adc_buffer: &'static mut [u16; 32_768] = {
         // Convert an uninitialised array into an array of uninitialised
-        let buf: &mut [MaybeUninit<u16>; 32_768] =
-            unsafe { mem::transmute(&mut BUFFER) };
+        let buf: &mut [MaybeUninit<u16>; 32_768] = unsafe {
+            &mut *(core::ptr::addr_of_mut!(BUFFER)
+                as *mut [MaybeUninit<_>; 32_768])
+        };
+
         // Initialise memory to valid values
         for slot in buf.iter_mut() {
             // Never create even a _temporary_ reference to uninitialised memory
--- a/examples/dma.rs
+++ b/examples/dma.rs
@@ -1,11 +1,10 @@
 //! Example of Memory to Memory Transfer with the DMA
 
-#![allow(clippy::transmute_ptr_to_ptr)]
 #![deny(warnings)]
 #![no_main]
 #![no_std]
 
-use core::{mem, mem::MaybeUninit};
+use core::mem::MaybeUninit;
 
 use cortex_m_rt::entry;
 #[macro_use]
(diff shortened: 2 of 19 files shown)


richardeoin:
I re-tested the following examples and they are still working:
* `examples/ethernet-rtic-stm32h747i-disco.rs`
* `examples/mdma.rs`
* `examples/usb_serial.rs`

I'm going to tag a few of the original authors by their examples below. If you still have access to some suitable hardware it would be really awesome if you can re-test these examples in the next week or two(!)

`examples/ethernet-rtic-nucleo-h723zg.rs` @jlogan03 
`examples/sai_dma_passthru.rs` @antoinevg 
`examples/serial-dma.rs` @mattico 
`examples/spi-dma-rtic.rs` @ryan-summers

jlogan03:
Hi! Thanks for doing this - I'm traveling this week, but I'll test it as soon as I'm back to my hardware

mattico:
@richardeoin thanks again for all the work you do maintaining this!

I tested examples/serial-dma.rs on an ST NUCLEO-H743ZI2 board and it worked, I checked the transmitted data.

I also tested examples/spi-dma-rtic.rs on the same board and checked the transmitted data. Well the first few bytes at least, I had to decode manually because my scope's SPI decoder wasn't working. It's been a while.

---

Rather than doing `TARGET_BUFFER.assume_init_mut()` on an uninitialized buffer, what would the "correct" solution be? Creating a wrapper struct containing a pointer and implementing `embedded_dma::WriteBuffer` on it? I won't ask you to implement that in this PR but maybe you could add a comment saying it's UB and pointing in the right direction, wherever that may be.

antoinevg:
Can confirm `examples/sai_dma_passthru.rs` is still working fine.

richardeoin:
> Rather than doing `TARGET_BUFFER.assume_init_mut()` on an uninitialized buffer, what would the "correct" solution be? Creating a wrapper struct containing a pointer and implementing `embedded_dma::WriteBuffer` on it? I won't ask you to implement that in this PR but maybe you could add a comment saying it's UB and pointing in the right direction, wherever that may be.

Creating a mutable reference like `&mut [u8]` to uninitialised memory is [UB](https://doc.rust-lang.org/reference/behavior-considered-undefined.html). The solutions would be:
* Initialise the memory first
* Or yes have a wrapper type around a raw pointer that implements `embedded_dma::WriteBuffer` without creating that `&mut` reference

richardeoin:
Rebased onto master

jlogan03:
Confirmed `ethernet-rtic-nucleo-h723zg` still runs

PR #488, opened 2024-03-13
Add support for using different modes for each phase of an xspi operation

dne:
This adds `Config.modes` and `Xspi.configure_modes` methods but keeps the existing API (which sets the same mode for all phases).

### DIFF
--- a/src/xspi/mod.rs
+++ b/src/xspi/mod.rs
@@ -101,10 +101,9 @@
 //! interface. Automatic polling or memory-mapped modes are not supported,
 //! except for the OCTOSPI Hyperbus mode.
 //!
-//! Using different operational modes (1-bit/2-bit/4-bit etc.) for different
-//! phases of a single transaction is not supported. It is possible to change
-//! operational mode between transactions by calling
-//! [`configure_mode`](#method.configure_mode).
+//! It is possible to change operational mode between transactions by
+//! calling [`configure_mode`](#method.configure_mode) or
+//! [`configure_modes`](#method.configure_modes)
 
 // Parts of the Quad and Octo SPI support are shared (this file), but they are
 // different enough to require different initialisation routines and pin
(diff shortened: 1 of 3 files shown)


src/xspi/mod.rs
```diff
@@ -176,6 +175,28 @@ mod common {
...
+        pub instruction: XspiMode,
+        /// IO lines used for the address phase.
+        pub address: XspiMode,
+        /// IO lines used for the alternate byte phase.
+        pub alt_byte: XspiMode,
+        /// IO lines used for the data phase.
+        pub data: XspiMode,
+    }
+    impl XspiModes {
+        pub fn new(mode: XspiMode) -> Self {
```
Abestanis, line 191:
```suggestion
        pub fn all(mode: XspiMode) -> Self {
```
Suggestion (I just saw this while browsing around, feel free to ignore): If this is called `all` then it is very clear that all modes are set to the given mode.

richardeoin:
Nice that this can be added without breaking backwards compatibility! I think `configure_modes` should have some documentation ("use this if you want different modes for each phase..") but otherwise looks good.

If you're able to share, it would be great to know if you have tested this on hardware and what the results were.

dne:
> Nice that this can be added without breaking backwards compatibility! I think `configure_modes` should have some documentation ("use this if you want different modes for each phase..") but otherwise looks good.

Thanks for reviewing! I've amended the PR to improve the docs, and also attempting to fix the octospi variant that failed in CI.

> If you're able to share, it would be great to know if you have tested this on hardware and what the results were.

I've used this successfully with the 'H745 and Infineon S25FL127S QSPI Flash (in quad mode), using the 4QOR and 4QPP commands ("quad output read" and "quad page program" with 4-byte addresses).

src/xspi/mod.rs
dne: > Suggestion (I just saw this while browsing around, feel free to ignore): If this is called `all` then it is very clear that all modes are set to the given mode.

@Abestanis Thanks for the suggestion; I'll let the maintainers decide what they prefer.

richardeoin: I see the reasoning behind calling it `all`, but `new` is definitely closer to the [naming convention](https://rust-lang.github.io/api-guidelines/naming.html#c-case) and is less surprising when quickly discovering the API. Thanks for the suggestion though @Abestanis !

richardeoin (approved):
Thanks for the updates

PR #489, opened 2024-03-14
Add access_banks flash accessor

Abestanis:
This adds a `access_banks` helper to the `FlashExt`, which is a middle ground between `access_bank1` and `split`.

## Problem
Right now the only way to access bank 2 is using the `split` method, which consumes the `FLASH` instance. In my usecase I want to access both flash banks but I also need to flip the banks at some point, so I need to have access to the option control register via [`FLASH::optcr(&self)`](https://docs.rs/stm32h7/0.15.1/stm32h7/stm32h743v/flash/struct.RegisterBlock.html#method.optcr), which needs the `FLASH` instance.

## Solution
Using the new `access_banks` helper I can keep the `FLASH` instance around and get access to both banks as well as the option control register when I need it.

### DIFF
--- a/src/flash/mod.rs
+++ b/src/flash/mod.rs
@@ -123,8 +123,13 @@ pub trait FlashExt {
     fn address(&self) -> usize;
     /// Size in bytes
     fn len(&self) -> usize;
-    /// Access to memory banks
+    /// Split the flash into its banks.
     fn split(self) -> (LockedFlashBank, Option<LockedFlashBank>);
+    /// Access to memory banks.
+    fn access_banks<T>(
+        &mut self,
+        f: impl FnOnce(&mut LockedFlashBank, Option<&mut LockedFlashBank>) -> T,
+    ) -> T;
     ///
     fn access_bank1<F, T>(&mut self, f: F) -> T
     where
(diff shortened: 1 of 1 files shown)


richardeoin:
This looks useful, and I don't think it breaks any existing assumptions. After all it doesn't modify registers.

It would be interesting to know if similar crates ([stm32-rs/stm32f4xx-hal](https://github.com/stm32-rs/stm32f4xx-hal) or [stm32-rs/stm32f7xx-hal](https://github.com/stm32-rs/stm32f7xx-hal) perhaps) already have methods like this

Abestanis:
* [`stm32-rs/stm32f4xx-hal`](https://github.com/stm32-rs/stm32f4xx-hal)'s flash extension doesn't really distinguish between the two  banks, they have a `FlashExt::dual_bank(&self) -> bool`, but their way of accessing the flash just takes a mutable self and returns an accessor that operates on sectors/offset from the flash base  regardless of the bank: [`FlashExt::unlocked(&mut self) -> UnlockedFlash`](https://github.com/stm32-rs/stm32f4xx-hal/blob/master/src/flash.rs#L52). So they don't have a `split` method and the `unlocked` method is like the `access_banks` method proposed here but it gives access to the entire flash range in one struct rather than two.
* [`stm32-rs/stm32f7xx-hal`](https://github.com/stm32-rs/stm32f7xx-hal) does not have a concept of banks, they just provide a thin wrapper over the `FLASH` peripheral with the [`fn lock(&mut self)`](https://github.com/stm32-rs/stm32f7xx-hal/blob/main/src/flash.rs#L47), [`fn program<'a, 'b>(&'a mut self, start_offset: usize, data: &'b [u8]) -> Result<ProgrammingSequence<'a, 'b>, Error>`](https://github.com/stm32-rs/stm32f7xx-hal/blob/main/src/flash.rs#L97C12-L101C52) and [`fn unlock(&mut self)`](https://github.com/stm32-rs/stm32f7xx-hal/blob/main/src/flash.rs#L35C12-L35C29) methods.

richardeoin:
Nice, I appreciate for the summary. Given this crate's abstraction over the banks these methods definitely make sense. Thanks again for the contribution!

PR #491, opened 2024-04-24
Fix off-by-one-tick in timer period calculation

dne:
STM32 timers, in upcounting mode, count from 0 to `ARR` (inclusive) before restarting at 0/generating an overflow event, so we need to subtract 1 from the number of desired ticks to get the correct period.

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -424,7 +424,7 @@ macro_rules! hal {
                 }
 
                 /// Sets the timer's prescaler and auto reload register so that the timer will reach
-                /// the ARR after `ticks` amount of timer clock ticks.
+                /// the ARR after `ticks - 1` amount of timer clock ticks.
                 ///
                 /// ```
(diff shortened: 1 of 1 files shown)


richardeoin:
Thanks for the PR and catching this off-by-one error! This change invalidates a doc comment on `timer.rs:427`, with this change the comment needs to state that the counter reaches the ARR after N-1 ticks and restarts at zero after N ticks. Could you update it? 
This change will be marked `Breaking` in the CHANGELOG, feel free to do this now or I can add it later.

dne:
> This change invalidates a doc comment on `timer.rs:427`, with this change the comment needs to state that the counter reaches the ARR after N-1 ticks and restarts at zero after N ticks. Could you update it? 

Thanks for the review. I've updated the doc comment I had missed.

richardeoin (approved):
Great, thanks!

PR #502, opened 2024-07-26
Adding blinky for stm32h7g-dk

anandsuresh9988:
Changes:
1) Added a blinky program for stm32h7g-dk DISCO board as it has different LED pin mappings as compared to other boards
2) Added openocd documentation for the stm32h7g-dk board as it is a very convenient way of testing and debugging the programs


richardeoin:
Thanks for the PR! I appreciate the effort to share something that will hopefully be useful for others. Since it's mostly information about the development board and openocd, I won't merge the PR into the main repository. There are other places to find information about debuggers and programming the H7, and while it is convenient to have the information here we sadly don't have the time and resources to keep it up-to-date.

PR #504, opened 2024-08-06
Moving embedded-hal 0.2 traits to separate package

ryan-summers:
This PR migrates the existing embedded-hal traits into the `hal_02` specifier so that we can continue with freely adding in new e-h 1.0 implementations as they are desired.

PR #505, opened 2024-09-21
Add embedded-hal 1.0 traits to stm32h7xx-hal

adamhott:
I've added both the embedded-hal 0.2.7 and embedded-hal 1.0 declarations in Cargo.toml.

I've also added the embedded-hal 1.0 trait definitions, but we need to correct the naming of the embedded-hal 1.0 traits because there are now duplicate definitions between 0.2.7 trait and 1.0 traits.

Please help me understand how to properly name the 1.0 traits.

Thank you!


adamhott:
This work has already been completed in PR #504

PR #508, opened 2024-11-28
Upgrade smoltcp to 0.12.0

tlevora:


src/ethernet/eth.rs
```diff
@@ -816,7 +816,7 @@ pub struct RxToken<'a, const RD: usize>(&'a mut RDesRing<RD>);
 impl<'a, const RD: usize> phy::RxToken for RxToken<'a, RD> {
     fn consume<R, F>(self, f: F) -> R
     where
-        F: FnOnce(&mut [u8]) -> R,
+        F: FnOnce(&[u8]) -> R,
     {
         let result = f(unsafe { self.0.buf_as_slice_mut() });
```
richardeoin, line 821: With this change it also becomes possible to return an immutable reference from the `buf_as_slice_mut` function on `RDesRing` and thus avoid triggering the `clippy::mut_from_ref` lint! That would be a nice simplification, and since `RDesRing` is private it wouldn't impact the external API

richardeoin:
Thanks for the PR! Because smoltcp 0.12.0 is only compatible with rustc back to 1.80.0, you need to raise the MSRV of this crate to pass CI. There's a [commit here](https://github.com/stm32-rs/stm32h7xx-hal/pull/415/commits/eedafcb7ad3a427c8812b62b0de9cbb8c1a9fb4f) that shows how to change the MSRV in both the README and the CI.

richardeoin (approved):

PR #509, opened 2024-11-29
In examples, disable the static-mut-refs lint

richardeoin:
[static-mut-refs](https://doc.rust-lang.org/nightly/rustc/lints/listing/warn-by-default.html#static-mut-refs)

For the future, need to investigate:
* How to implement DMA without requiring the lifetime of the buffer to be `'static`
* How to specific the link_section of non-static buffers

Suggest merging to fix CI on stable and nightly. (clippy and MSRV versions will still need a bump as well)

### DIFF
--- a/.github/workflows/clippy.yml
+++ b/.github/workflows/clippy.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@master
         with:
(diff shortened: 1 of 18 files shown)


richardeoin:
I think static-cell would be a much better alternative actually. Feel free to make an alternative PR!

richardeoin:
Closing in favour of #512, or something based on the ideas discussed in #513

Issue #507, opened 2024-11-15
Wrong Ethernet DMA Descriptors Memory Alignment


tlevora:
It could happen that `ethernet::DesRing` instance get wrong memory alignment. It is an issue since ethernet dma peripheral requires the ring descriptor to be aligned to 32-bit in memory. If the alignment requirement is not satisfied, dma does not use descriptors properly.

```rust
// The DES_RING has an alignment of 1 byte, so that it could be placed to e.g. an odd
// memory address. The DMA does not 'see' the descriptor because it ignores two least
// significant bits in descriptor's address.
static mut DES_RING: MaybeUninit<ethernet::DesRing<4, 4>> = MaybeUninit::uninit();
```

**Proposal**

In my opinion the `DesRing` alignment should be explicitly specified within descriptor ring structs declaration.

When looking to `TDesRing` and `RDesRing` respectively I'm not pretty sure why they are `packed`, guessing it should be just fine to change `packed` to `align(4)`.

```rust
#[repr(C, packed)] // TODO: replace packed with align(4) ?
struct TDesRing<const TD: usize> {
    ...
}
```

If I am overlooking something and `packed` is required here, the alignment could be alternatively forced in `DesRing` structure declaration.

```rust
// TODO: add #[repr(align(4))]
pub struct DesRing<const TD: usize, const RD: usize> {
    ...
}
```

PR #510, opened 2024-12-02
Fix eth::DesRing alignment

tlevora:
Fixing issue https://github.com/stm32-rs/stm32h7xx-hal/issues/507.

### DIFF
--- a/src/ethernet/eth.rs
+++ b/src/ethernet/eth.rs
@@ -69,8 +69,12 @@ use self::emac_consts::*;
 /// Note that Copy and Clone are derived to support initialising an
 /// array of TDes, but you may not move a TDes after its address has
 /// been given to the ETH_DMA engine.
+///
+/// Both order and alignment is required by the ETH peripheral. The repr(C)
+/// ensures that alignment of the members is 4 with no padding on the target
+/// platform.
 #[derive(Copy, Clone)]
-#[repr(C, packed)]
+#[repr(C)]
 struct TDes {
     tdes0: u32,
     tdes1: u32,
@@ -94,7 +98,6 @@ impl TDes {
 }
 
 /// Store a ring of TDes and associated buffers
-#[repr(C, packed)]
 struct TDesRing<const TD: usize> {
     td: [TDes; TD],
     tbuf: [[u32; ETH_BUF_SIZE / 4]; TD],
@@ -187,7 +190,7 @@ impl<const TD: usize> TDesRing<TD> {
         self.td[x].tdes2 = (length as u32) & EMAC_TDES2_B1L;
 
         // Create a raw pointer in place without an intermediate reference. Use
-        // this to return a slice from the packed buffer
+        // this to return a slice from the buffer
         let addr = ptr::addr_of_mut!(self.tbuf[x]) as *mut _;
         core::slice::from_raw_parts_mut(addr, len)
     }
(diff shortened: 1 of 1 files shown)


richardeoin:
I suspect the `packed` annotation was originally an attempt to prevent padding between the members of the struct. However it forces only byte alignment of the struct itself, which is not intended. And the `repr(C)` is sufficient to prevent padding.
* Would it make sense to remove all the uses of the `packed` annotation?
* Why not add the `align(4)` annotation?
* There's a comment on line 190 that refers to "packed", maybe good to remove this too

tlevora-dtml:
> I suspect the `packed` annotation was originally an attempt to prevent padding between the members of the struct. However it forces only byte alignment of the struct itself, which is not intended. And the `repr(C)` is sufficient to prevent padding.
> 
> * Would it make sense to remove all the uses of the `packed` annotation?
> * Why not add the `align(4)` annotation?
> * There's a comment on line 190 that refers to "packed", maybe good to remove this too

Originally I wanted to make just a minimal code update to fix the issue, eventually I decided to review all the related `repr` statements and here is my opinion:
- `[TR]Des` structs
  - They must be `repr(C)` because of ordering.
  - Alignment of all their members is platform specific, which means it should be 4 for `u32` on our target platform.
  - If we prefer the code to be **portable**, we can explicitly define the alignment to be **`align(4)`**.
- `[TR]DesRing` structs could be `repr(Rust)` because
  - `[tr]d` has already defined proper order and alignment
  - `[tr]buf` shall be aligned to 4 (platform specific) and there should be no padding between values in the array, because `u32` size is 4.
  - `[tr]didx` shall be aligned to 4 (platform specific), so that no hard-fault can occur because of unaligned access.
  - The member order is not important at all.
  - Padding between members does not matter at all. _(But there is no padding.)_

What do you think?

tlevora:
> I suspect the `packed` annotation was originally an attempt to prevent padding between the members of the struct. However it forces only byte alignment of the struct itself, which is not intended. And the `repr(C)` is sufficient to prevent padding.
> 
> * Would it make sense to remove all the uses of the `packed` annotation?
> * Why not add the `align(4)` annotation?
> * There's a comment on line 190 that refers to "packed", maybe good to remove this too

Originally I wanted to make just a minimal code update to fix the issue, eventually I decided to review all the related `repr` statements and here is my opinion:
- `[TR]Des` structs
  - They must be `repr(C)` because of ordering.
  - Alignment of all their members is platform specific, which means it should be 4 for `u32` on our target platform.
  - If we prefer the code to be **portable**, we can explicitly define the alignment to be **`align(4)`**.
- `[TR]DesRing` structs could be `repr(Rust)` because
  - `[tr]d` has already defined proper order and alignment
  - `[tr]buf` shall be aligned to 4 (platform specific) and there should be no padding between values in the array, because `u32` size is 4.
  - `[tr]didx` shall be aligned to 4 (platform specific), so that no hard-fault can occur because of unaligned access.
  - The member order is not important at all.
  - Padding between members does not matter at all. _(But there is no padding.)_

What do you think?

richardeoin:
Thanks for the clear explanation! I'm agreed.

> [TR]Des structs

The code doesn't need to be portable since this crate specifically targets a 32-bit architecture. So agreed with `repr(C)` and that the comment explains it sufficiently

PR #513, opened 2024-12-07
WIP Remove static muts from examples

tlevora:


richardeoin:
As discussed in #512, dereferencing a `static` of type `StaticCell` with `#[link_section = ".."]` is UB if the runtime doesn't initialise the memory area specified by `link_section`.

I think it should be something like
```rust
#[link_section = ".axisram.eth"]
static DES_RING: MaybeUninit<SomeKindOfUnsafeCell<ethernet::DesRing<4, 4>>> = MaybeUninit::uninit();
```
but need to work out what `SomeKindOfUnsafeCell` is exactly

usbalbin:
Also related https://github.com/rust-embedded/cortex-m/issues/538

Joker2770:
It's OK to use `core::mem::transmute`?
https://github.com/stm32-rs/stm32h7xx-hal/blob/0735f793e6d432653f2a7f1d8a2d05f073ca4f17/examples/adc_dma.rs#L49

tlevora:
I am sorry I haven't replied here for a long time.

I like #518 and do not have a better solution, so that closing this PR. In specific situations I would use something like https://crates.io/crates/linker-sections, but probably not here.

PR #515, opened 2024-12-14
Fix more clippy warnings

usbalbin:
Turned out to be more warnings that were drowned by the static mut errors


usbalbin:
Let me know if you want me to revert 2b21494

richardeoin (approved):
I only like https://github.com/stm32-rs/stm32h7xx-hal/commit/2b2149406ba4e695ebe265dc65aea222b6dbdf52 as a short-term solution, but worth it to get the CI passing again. Thanks!

Issue #521, opened 2025-03-03
Wrong pins in ethernet-rtic-nucleo-h723zg example


TimH1502:
These are the pins for the lan8742A-CZ-TR, which are used from the example (https://github.com/stm32-rs/stm32h7xx-hal/blob/master/examples/ethernet-rtic-nucleo-h723zg.rs) in version 0.16.0:

```
let rmii_ref_clk = gpioa.pa1.into_alternate();
let rmii_mdio = gpioa.pa2.into_alternate();
let rmii_mdc = gpioc.pc1.into_alternate();
let rmii_crs_dv = gpioa.pa7.into_alternate();
let rmii_rxd0 = gpioc.pc4.into_alternate();
let rmii_rxd1 = gpioc.pc5.into_alternate();
let rmii_tx_en = gpiob.pb11.into_alternate();
let rmii_txd0 = gpiob.pb12.into_alternate();
let rmii_txd1 = gpiob.pb13.into_alternate();
```

The real pins are visible in the st schematics (https://www.st.com/resource/en/schematic_pack/mb1364-h723zg-e01_schematic.pdf) and should be changed to the following:

```
let rmii_tx_en = gpiog.pg11.into_alternate();
let rmii_txd0 = gpiog.pg13.into_alternate();
```

Due to using GPIOG the IO should be also included, before the ethernet init:

`let gpiog = ctx.device.GPIOG.split(ccdr.peripheral.GPIOG);`

PR #522, opened 2025-03-03
bugfix, wrong pins used in nucleo stm32h723zg ethernet example

TimH1502:
Fixed Issue in #521 
Different pins used from st schematic v1.0 (https://www.st.com/resource/en/schematic_pack/mb1364-h723zg-e01_schematic.pdf). Maybe there were different pins used for a version below 1.0, when the example was created.
Board is now pingable.


richardeoin:
Thanks for the PR!

Issue #448, opened 2023-08-25
regression in ethernet DMA initialization for H743 chips


ritikmishra:
One of the changes merged in by #445 moves the write to the ETH_DMAMR register to before the ethernet peripheral clocks are enabled

https://github.com/stm32-rs/stm32h7xx-hal/pull/445/files#diff-bc6d950ee1ed8e0df070d5ac2e9190dea3730fc251d9ca1b3c9bdad8c7f0bd3aR458-L478

This breaks the ethernet DMA initialization on my STM32H743IIT6

ryan-summers:
I can confirm that Stabilizer's hardware-in-the-loop tests indicate that networking no longer functions on the H743 chips [with the latest hal rev](https://github.com/quartiq/stabilizer/pull/774#event-10206101055)

richardeoin:
I can reproduce this on two development boards:
* STM32H747I-DISCO (STM32H747XIH)
* STM32H735G-DK (STM32H735IGK)

For me the symptom is that the PHY never reports the link is up (over the MDIO interface) and the board does not respond to pings. Reverting the change to `eth.rs` from #445 resolves the issue.

Why the change in #445 is required for the NUCLEO-H723ZG but fails on the STM32H735G-DK is really not clear. Those STM32 part share a reference manual, so I would expect them to both work with the same initialisation sequence. Perhaps either the board power configuration (LDO vs SMPS) or PHY configuration is to blame.

In the meantime I would propose reverting the change to `eth.rs` from #445 until we understand why it is needed for the NUCLEO-H723ZG.

ryan-summers:
I'm on board with your assessment @richardeoin - I didn't see a thorough explanation in #445 why it was required outside of "Seems to work now". If that truly is the case, we can device gate it, but feels like more investigation is required.

PR #524, opened 2025-04-30
Add back early ethernet DMA reset as an optional feature flag

guineawheek:
I hit the regression in https://github.com/stm32-rs/stm32h7xx-hal/issues/448 on an h743 board, and while the PR that fixed that was reverted, I still think it would be useful to a lot of people if the early ethernet DMA reset was available as a feature flagged quirk; as it's seemingly not even consistent across the same stm32h7 chip. 

My h743zi nucleo board works but the prod board simply does not, despite seemingly identical PHY setups. Both seem to work with the initialization being earlier, and I don't see this PR impacting anyone's existing code unless they specifically opt-into the reordering.

Thank you for your time.


guineawheek:
yeah turns out the freeze happens if your PHY or whatever isn't outputting refclk

richardeoin:
Seems this is quite a tricky topic especially with new boards. Just to confirm, the fix in this PR didn't actually help and the correct fix was to configure the PHY differently?

Glad you fixed the problem in the end and thanks for taking the effort to report here.

